Issue #15, opened 2018-11-14
Docker: Allow mysql databases to persist


bryanpizzillo:
Docker-compose down, the common way of shutting down a stack, will delete containers. MySQL stores its data in the containers. So restarting docker wipes out the database. The data folder for mysql should allow the data to persist across restarts.

blairlearn:
We've decided we *want* the database to go away between container runs.

PR #37, opened 2018-11-16
Persist database via a named volume. (Possible fix for #15)

blairlearn:
The Docker Library project documents a problem with some versions of MariaDB
not starting correctly when a host system directory is mounted to /var/lib/mysql.
https://github.com/docker-library/mariadb/issues/38

It appears to work correctly when a named volume is used; at least,
it works on Windows.  This should be tested on OS/X as well
before being merged.


blairlearn:
After discussion with @bryanpizzillo, we have decided not to use this PR after all as database persistence does not appear to be desirable at this time.

PR #88, opened 2018-12-03
Resolves #87

### LINKED ISSUE #87: Documentation: The setup instructions are out of date
It does not cover the cgov-init nor the new composer install on the client.

bryanpizzillo:


bkline:
I guess since I don't have write access, my approval won't do you any good. 🤔

PR #104, opened 2018-12-06
Feature/nciocpl#72 cannot run more than one docker stack at a time

### LINKED ISSUE #72: Cannot Run More than One Docker Stack at a Time
Currently with the last change to the docker-compose.yml file, you cannot run more than one container at a time based on the name.  Technically, this was not possible beforehand, but the `-p <project name>` flag with `docker-compose` would use `<project name>` to prepend the containers. So through the command line you could set them.

Partially this was done in order to have the hostname appear on the bash command line of the web container and not say `docker_web_1`. 

So this ticket is to fix the compose file so that we can run more than one stack at a time (with a different -p option). This should also set the hostname in the container TO the full container name. Additionally it should take in account the fact that we expose ports and they would need to either NOT be exposed, or random WHEN THERE IS MORE THAN 1 CONTAINER. (Please make it 80 and mysql when running a single stack)

### LINKED ISSUE #96: Allow custom ~/.bashrc to be injected into docker containers (per local user)
Allow devs to add their own custom code to be injected into the docker containers' ~/.bashrc file to store aliases, function and other customizations.   Customizations should be local to the computer and not be committed to repo.

Use a mounted file that `source`'s the customizations (if they exist) from `/root/.bashrc`

### LINKED ISSUE #97: Add custom commands to be added to local (host) ~/.bashrc or ~/.bash_profile from repo.
Add option to install custom code into host os's startup file, similar to how `composer init-cgov` concats to the startup file.

bgnih:
Closes #72 Closes #97 - Use environment variables in docker-compose.yml, update 'composer cgov-init' to add ldock() function and aliases to local ~/.profile, remove blank dc-override.yml.sample file.  Update docker/README.md.   (I rolled 72 and 97 into a single PR since only fixing 72 would have left the docker-compose.yml file in an unusable state without the ldock() function and macros.)

This update makes the `compose cgov-init` file idempotent (thanks for the new vocabulary word, @bryanpizzillo ), and updates the error messages to warnings appropriately.   

The `$project_root` setting was changed to remove the `../..` to make the more descriptive errors more user-friendly, and `$path` for file arrays was broken into `$srcpath` and `$destpath` for more flexibility and to support the needed `.docker.bashrc.example` file placement.

The `compose cgov-init` will now `source ~/.docker.bashrc` when the shell opens (using either .profile, .bash_profile, or .bashrc) and include an updated ldock() function and some example aliases to bring up/down/ssh-into 3 example environments.   Feel free to tweak the image names if desired.  It also sets default environment variables for `COMPOSE_PROJECT_NAME`, `WEB_PORT`, and `MYSQL_PORT` so `docker-compose` will work by itself if desired with a single container without failing due to the environment variables now needed within the `docker-compose.yml` file.

(Don't forget to SQUASH me!  - PR#104)


bgnih:
I agree that the aliases are a personal thing, and I included them so that users would have at least a handy command that could spin things up 'out of the box' from the `composer cgov-init` run.   They are easily editable/removable by editing the `~/.docker.bashrc.sh` file after the initialization (which would not be affected by future `cgov-init` runs).

blairlearn:
I'm of the same mind as @bryanpizzillo on this. The aliases shouldn't be included and very much don't belong as a side effect.

bryanpizzillo:
Closed this as it is more of a personal preference how people want to manage their docker containers.

PR #122, opened 2018-12-07
Resolves #119

### LINKED ISSUE #119: Combine field storage tests in cgov_core
The current tests must spin up an empty Drupal for each field test. If they are combined into one method tests would finish faster.

bryanpnih:
- Combined all tests into one method


bkline:
Agreed! This will give me a leg up on my own kernel tests.

Issue #106, opened 2018-12-06
Tests: create automated tests for workflow configuration


bkline:
See issue #64 Implement Editorial Workflow.

PR #186, opened 2018-12-12
Bug/64 fix workflow

bkline:


bkline:
@bryanpizzillo: The PR failed because RolesTest.php couldn't find the `content_author` role (and I assume would have had the same problem with other roles had it not given up with the first failure). Do we (a) comment out the role tests? (b) hoist the tests out of the `modules` section higher in the tree? (c) adopt some other solution?

I'm a little puzzled, to be honest, that the tests inside the `modules` section can't find things which are configured in the enclosing profile's `config`. So nesting things inside a hierarchy doesn't accomplish what it normally would, I assume. (In a normal scoping world, inner contexts can see things created by enclosing contexts, but not the other way 'round.)

bkline:
I tried setting `protected $profile = 'cgov_site';` in the `RolesTest` class and adding `$this->setInstallProfile($this->profile);` in the `setUp()` method, but that didn't have any effect.

bryanpizzillo:
@bkline Did you add any tests, or are you working on that as a separate item?

bkline:
That's issue #106.

bkline:
Thanks! And thanks for your help.

PR #231, opened 2018-12-17
Ticket/71 final lang branch

### LINKED ISSUE #71: Enable Spanish Content
In order to kick the tires on Layout Builder and "On this page" Body fields, we need to have the Spanish language enabled. This will probably cause MANY other issues, but we need to pull off the band-aid sooner rather than later.

bryanpnih:
Closes #71
- Added Spanish
- Enabled Locale in Site Profile
- Created Profile install to fix up language.types
- Created Site Tools class for handling various installation tasks

### DIFF
--- a/docroot/profiles/custom/cgov_site/cgov_site.info.yml
+++ b/docroot/profiles/custom/cgov_site/cgov_site.info.yml
@@ -41,6 +41,7 @@ install:
   - workflows
   - content_moderation
   - language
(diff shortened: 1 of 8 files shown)


bryanpizzillo:
@bkline For the record, the tests are actually testing that the installation profile changes the configuration as expected. There should be a separate unit test for the CgovSiteTools class which tests that code.

PR #295, opened 2018-12-19
Add helper function to attach content types to workflow (closes #127)

### LINKED ISSUE #127: Add a helper function to attach content types to workflow

bkline:
I have included the test for this new method by modifying `WorkflowTest.php` to invoke it instead of loading the workflow through the entity type manager directly.

It seems to me that as time goes on and the `CgovCoreTools` class is expanded to support more and more functionality that the argument list for the constructor is going to get pretty unwieldy. Not only is this ungainly and a bit hard to read in the code, but it's somewhat inefficient as well. As I understand it, one of the selling points for dependency injection is that the services needed for a component are only loaded when they are needed. However, that advantage starts to unravel when you build an omnibus service which in turn might need one or two of a bunch of other services, depending on which method of the new service is being invoked. If I were King of the Forest, I think I might have designed an architecture which allowed the dependency container to be passed to the constructor and saved as a member of the service class, and have the `$container->get()` method invoked for the specific service(s) needed by each of the methods. But I don't see a way to do that with the way Drupal has things set up (if there is a way, please let me know). And I'm not KOTF, so ....

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.services.yml
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.services.yml
@@ -4,4 +4,7 @@ services:
     arguments: ['@current_route_match']
   cgov_core.tools:
     class: Drupal\cgov_core\CgovCoreTools
-    arguments: ['@config.factory', '@language_negotiator']
+    arguments:
+      - '@config.factory'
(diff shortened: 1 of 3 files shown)


bkline:
Not sure what the difference is between "Reviewer" and "Assignee"; I think I have inadvertently done some overkill by setting both. 🤔

Issue #129, opened 2018-12-07
 Display the date on Article


VictoriaSunNIH:
This story is to displaythe [common date field](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Common-Fields#content-dates)s to [Article ](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Content-Type-Article)content type. The date field creation and storage was addressed in  #110.

**User Story:**
* As a visitor, I would like to see the date a content author has set on an article on cancer.gov

**Acceptance Criteria:**
   * Date display rules - when more than one date is selected in Date Display Mode - use the most recent of the dates selected. Display the label of that date, and that date. If only one is selected, show the label and the date. If nothing is selected, no date displays.

BrendanBeltzNIH:
Approach: There are multiple ways to handle the logic requirements here. A preprocess function, a custom twig extension, a twig macro, and finally an inline twig conditional. For a first pass, we will use a reusable twig macro.

bgnih:
Twig functions `{{ newestDate(date1, date2, date3) }}` and Computed field `$field_display_date` may also be options (just to be thorough)

BrendanBeltzNIH:
Twig functions are custom twig extensions FYI. The computed field would be nice.

BrendanBeltzNIH:
Current approach:

1. Create a preprocessor to pass in a variable to the twig templates for the effective Date Display Mode
   This is where article can remove all but one display
   Press release would keep them all
   etc.
1. Make a twig template to render dates that could render all of the dates.
   The order of the dates to be displayed is Posted, Updated, Reviewed
   The output will be along the lines of:
```<div class="document-dates horizontal">
            <ul class="clearfix">
                <li><strong>Posted:</strong> September
 27, 2013</li>            
                            <li><strong>Updated:</strong> August
 12, 2014</li>            
            </ul>
 </div>```
1. Then call that twig from within the article template
We can reuse the date display template across the site

bgnih:
If there will be any requirement to be able to sort by this field in Views, then a computed field (database storage) will likely be best option.   Otherwise, anything works.  ;-)

bryanpizzillo:
@bgnih You make a valid point, but I think that should be implemented as a "field_sort_date" computed field. I.E. Use these rules to calculate what the sort date is...

(One type of content even used to have a totally separate Sort Date field. It may be going away though, but it illustrates the point)

Issue #366, opened 2018-12-24
Resolve issues with Translation files and translated configurations


bryanpizzillo:
@bryanpizzillo - Type what you're doing here.

bryanpizzillo:
Attempting to add translations to profile following:
https://api.drupal.org/api/drupal/core%21modules%21locale%21locale.api.php/group/interface_translation_properties/8.2.x

It is not 100% clear if this needs to be added to all themes and modules we make.

bryanpizzillo:
Added some more workarounds. I forced the translations path in settings.php to be at the docroot root. Then I disabled downloads. This prevents an errant drupal-8.x.es.po file from being downloaded.

bryanpizzillo:
@daquinohd Created a blt command to load up translation files. This is a more permanent temporary workaround for now. (Core still has the issue, but process wise this gets our translations in.)

bryanpizzillo:
ACSF still will have an issue.

VictoriaSunNIH:
@bryanpizzillo  said that he created a post update hook and post install hook that will call the drush locale update and so our translations get imported correctly into the ACSF sites. Language still sucks but we can close out this issue.

PR #372, opened 2018-12-27
Date display mode using preprocess approach

BrendanBeltzNIH:
Handles #129.

This is definitely an approach worth reconsidering sooner or later, when we all get a bit more fluent in computed fields and plugins. It may well be a good way to go about it all told, but it's more than likely swimming upstream and not an ideal approach for replication.

The gist is as follows:

Setup:
A date display mode field exists that tracks what kinds of dates (posted, updated, reviewed) a content creator wants to appear on a piece of content. It is effectively a list of strings that match the formatted display text of the options in the editor ui. (This seems to be the crux of it, instead of a substantial link to the other field entities themselves, and is what requires a brittler approach).

Goal:
Be able to constrain the visible dates on certain content types (article in this case) by certain logical parameters.

Approach:
Do this using the field_date_display_mode preprocess function

Implementation:
First of all, the preprocess function is only called when the field is called in a twig template, so we need to at least include it in the node level template. 
Second of all, computed values in the preprocess function get passed to the render array, so they aren't available in the node template in a straightforward way. This means what we really want to do is nest the other fields' templates inside of the display mode one so they are appropriately only rendered conditionally based on the computed values. That's not so straightforward. The implementation I took here is to grab the date fields off the node, filter them if necessary, and then pass them through on the render array of the display mode. However, that wasn't sufficient. I found I needed to first call their 'view' method in order to have them render correctly (using the default templates at least - thereby hopefully not creating new translation problems). (The alternative was reaching onto the node from the twig template itself and wrapping the whole thing in conditionals based on the preprocess output, but this seemed dirtier since it split the logic between the preprocess and template. My approach keeps everything in the preprocess.)

Thirdly, I had to use a hard coded copy of the field_date_display_mode's formatted text in the preprocess function to be able to programmatically call and associate the field entities with the user selections in the field_date_display_mode. 
(Side note, using a stable sort method on the dates, the order of the elements in that array determines their priority in rendering when only one will be rendered but two selected dates are both equally recent. (PHP associative arrays are ordered, unlike JS)).

Lastly, don't merge this PR until I can strip out the in progress type comments!

### DIFF
--- a/docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/cgov_common.theme
+++ b/docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/cgov_common.theme
@@ -5,3 +5,117 @@
  * Functions to support theming in the cgov_common theme.
  */
 
+ /**
+  * Date Display Mode.
+  *
+  * This is some classic Drupal contortionist work possibly better handled with
+  * a computed field value or a custom block in the future.
+  * In order to have three unassociated fields' display mode controlled by a
+  * fourth unassociated field's (this one) preprocess function, we need a way
+  * to both examine and render those fields outside the normal flow.
+  */
+function cgov_common_preprocess_field__field_date_display_mode(&$variables) {
+  $node = $variables['element']['#object'];
+  if (!$node) {
+    return;
+  }
+  // If there are no display modes this preprocess never runs. But
+  // leaving the guard clause anyway for the nonce.
+  if (!isset($variables['items']) || empty($variables['items'])) {
+    return;
+  }
+
+  // This is a mapping of the field machine name and content display text.
+  // Unfortunately the content display text needs to match and it's a string.
+  // Ugh.
+  // NB: The current sort method is stable, so in the case of two dates, being
+  // the same, the order of priority is determined by the order of this map.
+  // I.e., if all three date types are the same time stamp, field_date_reviewed
+  // will be the only one shown on an article page.
+  $dates = [
+    'Posted Date' => 'field_date_posted',
+    'Updated Date' => 'field_date_updated',
+    'Reviewed Date' => 'field_date_reviewed',
+  ];
+
+  // Retrieve date types selected for display.
+  // Also, if there are no display modes this preprocess never runs. but
+  // leaving the array type check anyway for the nonce.
+  $items = $variables['items'];
+  $availableModes = [];
+  foreach ($items as $item) {
+    $type = $item['content']['#markup'];
+    // Weird preferred php push method, not array declaration.
+    $availableModes[] = $type;
+  }
+
+  // Now we need to filter out the date types we don't want.
+  // (Based on the user selected date display modes).
+  $filteredDates = [];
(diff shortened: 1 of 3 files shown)


bryanpizzillo:
Some of these are note from off line discussions, and some are from investigation.

* The preprocessor approach, which you heroically got working, is cumbersome.
* It is best to separate this story into two parts:
  * How do we determine which dates to display?
  * How do we display the dates?

**Date Display**
* The dates & labels are not translated.
   * The date formats differ between English and Spanish
      * Spanish: 29 de abril de 2015
      * English: April 29, 2015
   * Formats can be "translated" if we enable Configuration Translation. *I will look into what new pains that module will cause.*
   * The labels should be able to be translated easily once #366 is figured out. (Although this should not hinder the template creation -- just testing) 
* Additionally the labels displayed on the front-end are different from the back-end field labels. (e.g. UI - `Posted:` and `Publicación` vs Back-end Label `Posted Date`.
* I did find you can output a date in twig with `{{ node.field_posted_date|date('U')|format_date('date_formatter_machine_name')}}`
  * NOTE: `format_date` is actually [DateFormatter::format](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Datetime%21DateFormatter.php/function/DateFormatter%3A%3Aformat/8.6.x) and not the `format_date` function. Googling `format_date` gets you the function...
* The code in the PR does this, just reiterating... We should always assume all the dates *could* display. It is the job of "something else" to figure out what dates *should* display.
* I definitely feel icky about the date display mode field template getting the dates pushed into it. 
  * Normally a field template renders a single field. This could lead to confusion for new devs figuring out how there are additional "fields" on the field.
  * I also don't think we should even include the field in the Article Full View Display. 
    * We can just use `node.field_date_display_mode`
       * I am trying to figure out how we can get an array of values, then you can use the twig `in` operator.
       * The way we have done it in the past, even before the current system, was to make Date Display Mode a bit map. So, Posted = 1, Updated = 2, Reviewed = 4. "Something" adds up the values into a var. Then you use a bitwise and operator to check if the value contains posted. It looks like twig uses `b-and` for the bit.
       * So maybe we can use a calculated field for this bitmap if checking the array does not work out. (More on how we are going to implement the various display choices in the next section)   
  * This method does make it reusable, but I think we can use a [twig macro](https://www.drupal.org/docs/8/theming/twig/macros-in-twig-templates) for the display logic. That way the HTML code is reused.

** How we figure out the date **
* Probably the best way to implement the various business rules for the content type's date displays is to use a virtual field.
   * I had actually though the preprocessor would just make a variable available in the context of the node to determine which one to show -- not shove dates into the date display mode field. 
     * Although even with this approach things get messy because it is the theme owning the logic, not the individual modules. (e.g. if a common does not have a press releases, why does the common theme have logic for that...)
* So each type should get a virtual "calculated_date_display_mode" field. That should either be an array of items, OR, just an int and we can use bitwise ands to check if we should show.

bryanpizzillo:
Closing this as @BrendanBeltzNIH found a better way.

PR #373, opened 2018-12-27
Fix available moderation states for new translation (fixes #371)

### LINKED ISSUE #371: Available states determined incorrectly for new content translations
When an English content item is published, its new translation will not be able to be saved as a draft. We will investigate to see if we can come up with a patch. Temporary workaround would be to adopt a modified (simpler) workflow which does not distinguish between editing/review of content before and after it has been published. see https://www.drupal.org/project/drupal/issues/3021222.

It may be possible to work around this bug by implementing a translation_create hook. I'm working out the details.

bkline:
This is my workaround for https://www.drupal.org/project/drupal/issues/3021222. I gave them the fix (though not as a patch with a new test). We'll see whether they get it into 8.7.0. If they do, the real fix should co-exist peacefully with our workaround until we are satisfied that we can drop it (this workaround, that is).

I noticed when I was testing the workaround that (after pulling down the latest mods from upstream) some, but not all of the moderation states are translated into Spanish. So "Published" becomes "Publicado" but we still have "Draft" and "Review." Seems like they have a ways to go sorting out the fine points of interactions between content translation and content moderation. 🤔

I also noticed that the order of the transitions determines the order of the available states in the dropdown, so I did a tiny bit of tweaking there.

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.module
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.module
@@ -4,3 +4,22 @@
  * @file
  * Contains cgov_core.module.
  */
+
+use Drupal\Core\Entity\EntityInterface;
+
+/**
+ * Implements hook_entity_translation_create().
+ */
+function cgov_core_entity_translation_create(EntityInterface $translation) {
+  $serviceName = 'content_moderation.moderation_information';
+  $moderationInformation = \Drupal::service($serviceName);
+  if ($moderationInformation->isModeratedEntity($translation)) {
+    $workflow = $moderationInformation->getWorkflowForEntity($translation);
+    $configuration = $workflow->getTypePlugin()->getConfiguration();
+    $state = 'draft';
+    if (!empty($configuration['default_moderation_state'])) {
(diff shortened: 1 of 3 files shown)


bryanpizzillo:
About translations -- you have to do the translations yourself for items that you create. (Draft should have a translation)  We are NOT going to guarantee that the Admin UI is completely translated for launch. So you should not worry about that.

bkline:
Do you want comments in the code or the ticket (my understanding is that hook implementations are only supposed to have the "Implements hook...." comments, though that wouldn't preclude having comments sprinkled through the code). Wherever they go, they will say

1. the hook gets passed the entity object for the translation (which is why I named it `$translation`)
2. the hook is called when a new translation is created, giving us the opportunity to modify the object for that translation
3. we're setting the initial moderation state for the translation, and that in turn is used by the logic which determines which states are available when the new translation is saved (the bug just copies the initial state from the current state of the original language)

bryanpizzillo:
@bkline I was looking for code comments. Thanks!

bryanpizzillo:
Hi Bob, the comment looks great. I also was perusing the ticket itself, which should be required reading. Someday the status checks will finish and I will accept the PR. Great work!

PR #383, opened 2018-12-31
Install chrome and fix behat's invocation of it (fixes #48)

### LINKED ISSUE #48: Chrome is missing from Docker image
This is required by beHat.

bkline:
I believe I have finally wrestled this one to the ground. I won't go into _all_ of the problems I ran into, but the highlights included very sketchy documentation (no information on how to generate the patch file or determine what the `--level` should be), next to no information provided when a patch fails, and the fact that `--dry-run` is useless when testing plugins (because there's no mechanism for `composer` to have the plugins say what they would do without doing it).

You have to run `composer cgov-clean` followed by `composer install` and then rebuild the containers. I did that, ran `blt setup` and then `blt tests:behat:run` and the `behat` tests ran and completed successfully.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -153,6 +153,9 @@
         "patches": {
             "drupal/core": {
                 "2599228 : Programmatically created translatable content type returns SQL error on content creation": "https://www.drupal.org/files/issues/2018-11-29/2599228-104.patch"
+            },
+            "acquia/blt": {
+                "NCIOCPL#48": "patches/blt/issue-48-fix-behat-chrome-invocation.patch"
             }
         },
         "drupal-scaffold": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "0dd1209d3a16e9319338664caabab6ad",
+    "content-hash": "a83716b7fd75a8af9c8ad50ed58a187b",
     "packages": [
         {
             "name": "acquia/blt",
@@ -68,6 +68,9 @@
                 "composer-exit-on-patch-failure": true,
                 "branch-alias": {
                     "dev-master": "9.x-dev"
+                },
+                "patches_applied": {
+                    "NCIOCPL#48": "patches/blt/issue-48-fix-behat-chrome-invocation.patch"
                 }
             },
             "autoload": {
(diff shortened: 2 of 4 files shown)


bkline:
The build failed with the error message

> The lock file is not up to date with the latest changes in composer.json, it is recommended that you run `composer update`.

So I did that, and got the following output:

```
Could not scan for classes ... [a bunch of these errors for a familiar `blt` bug]
Gathering patches for root package.
> DrupalProject\composer\ScriptHandler::checkComposerVersion
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 3 installs, 40 updates, 0 removals
Gathering patches for root package.
Gathering patches for dependencies. This might take a minute.
  - Updating symfony/yaml (v3.4.18 => v3.4.20): Downloading (100%)
  - Updating twig/twig (v1.35.4 => v1.36.0): Downloading (100%)
[... lots more lines like this ...]
  - Updating symfony/cache (v4.1.7 => v4.2.1): Downloading (100%)
  - Updating symfony/expression-language (v3.4.18 => v3.4.20): Downloading (100%)
  - Updating drupal/core 8.7.x-dev (0d0356b => 2c88ba7):     The package has modified files:
    M modules/content_translation/content_translation.module
    M modules/content_translation/content_translation.services.yml
    M modules/content_translation/migrations/d7_entity_translation_settings.yml
    M modules/content_translation/migrations/d7_node_translation.yml
    M modules/content_translation/src/ContentTranslationManager.php
    M modules/content_translation/src/ContentTranslationUpdatesManager.php
    M modules/language/migrations/d6_language_content_settings.yml
    M modules/language/migrations/d7_language_content_settings.yml
    M modules/system/tests/modules/entity_test/src/Entity/EntityTestWithBundle.php
    Discard changes [y,n,v,d,s,?]? ?
    y - discard changes and apply the update
    n - abort the update and let you manually clean things up
    v - view modified files
    d - view local modifications (diff)
    s - stash changes and try to reapply them after the update
    ? - print help
    Discard changes [y,n,v,d,s,?]?
```

Any thoughts on what this is all about, and which answer is the right one?

bkline:
Got past that hurdle, with the help of BP.

Another pro tip: `composer update some/target` isn't always as helpful as you might think when trying to run a patch, because `composer` caches the record of already having applied the patch in _lots_ of hidden-away places. Running `composer install` (after cleaning up from the previous installation) is the only approach I'm _really_ confident will do what I expect.

PR #386, opened 2019-01-02
Match default yaml to fields on article node

BrendanBeltzNIH:
Hotfix because article.content.yml was still referencing "body" on article node instead of new "field_article_body"


bryanpizzillo:
Closing this -- it was taken care of in another PR

PR #454, opened 2019-01-09
Feature/image media

bryanpnih:
The long awaited image feature is here.


bryanpnih:
Note, because I don't have to do as I say, there are no real tests. The idea is that there should be behat tests for the images/articles. Kernel tests with file uploads are not something I want to tackle right now.
Additionally, the workflow requirements & permissions are being ironed out, but that can be a backlog/small update.

bryanpnih:
@adriancofie yep about the other displays. Also we need crops, etc. However, much like the workflow, req are pending.

PR #525, opened 2019-01-17
PDQ summary documentation

blairlearn:
Adds basic documentation for the `pdq_cancer_information_summary` module.


blairlearn:
**DO NOT USE.**  My branch seems to be missing a commit from `upstream` and may have picked up some other rubbish.

PR #545, opened 2019-01-23
Ticket/486 landing paragraphs

bryanpnih:
This is just a PR to get my Landing Page work onto an ODE.


bryanpizzillo:
Also, I am keeping all the commits for now on this PR, not squashing. So if this branch does get merged in, then it should be squashed.

PR #580, opened 2019-01-29
Implement PDQ Cancer Information Summaries

blairlearn:


docroot/profiles/custom/cgov_site/modules/custom/pdq_cancer_information_summary/src/Plugin/rest/resource/PDQResource.php
```diff
@@ -0,0 +1,271 @@
...
+   * @return \Drupal\rest\ModifiedResourceResponse
+   *   The response containing node ID(s) or a Drupal node
+   *
+   * @throws \Symfony\Component\HttpKernel\Exception\HttpException
+   *   Throws exception expected.
+   */
+  public function post(array $summary) {
+
+    // Extract the bits we'll use frequently.
+    $nid = $summary['nid'];
```
bryanpizzillo: Could the service look up the NID based on the CDRID? Was this decision because a new CDRID could replace an existing summary?

docroot/profiles/custom/cgov_site/modules/custom/pdq_core/pdq_core.install
```diff
@@ -0,0 +1,37 @@
...
+    'restful get pdq_api',
+    'restful post pdq_api',
+  ];
+  $siteHelper->addContentTypePermissions('pdq_cancer_information_summary', ['pdq_importer'], CgovCoreTools::DEFAULT_PERMISSIONS);
+  $siteHelper->addRolePermissions(['pdq_importer' => $restfulPermissions]);
+  $user = User::create();
+  $user->setUsername('PDQ');
+  $user->setEmail('pdq@cancer.gov');
+  $user->enforceIsNew();
+  $user->setPassword(getenv('PDQ_IMPORTER_PASS'));
```
bryanpizzillo: This is OK for now, but we should add a backlog item for setting up API keys. That way we can rotate them via a env var and not a password that is set in the DB.

docroot/profiles/custom/cgov_site/modules/custom/pdq_core/src/Plugin/rest/resource/PDQResource.php
```diff
@@ -0,0 +1,229 @@
+<?php
```
bryanpizzillo: Any idea why this PR has two of these?

docroot/profiles/custom/cgov_site/modules/custom/pdq_core/src/Plugin/rest/resource/PDQResource.php
```diff
@@ -0,0 +1,229 @@
...
+   * @return array
+   *   Sequence of value pairs, each of which has a node ID and a
+   *   language code.
+   */
+  private function lookupCdrId($id) {
+    if (substr($id, 0, 3) === 'CDR') {
+      $id = substr($id, 3);
+    }
+    $matches = [];
+    $query = \Drupal::database()->select('node__field_pdq_cdr_id', 'c');
```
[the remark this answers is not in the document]
bkline: Possibly. I wanted a query which didn't depend on knowing anything about the specific bundles for the PDQ content types.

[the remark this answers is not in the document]
bkline: Digging further (gently prodded by Brendan) I realize I was mistaken about how `EntityQuery` works. Here's the Slack exchange:

**Brendan**: curious about your pr. especially about why you directly queried the database instead of usinng a service like entityquery. Figure you must have some reason, thought it would be useful to understand for the future.

**Bob**: Excellent (implied) question. I had a bunch of reasons, not all of them very good.

1. More efficient
2. I have found over the years that it's easier to reverse-engineer the Drupal tables than to find comprehensive, useful Drupal documentation
3. I find Drupal's deprecation churn cycle even more annoying than Microsoft's ("learn and use this API -- no, wait, that's no good, use these instead..."), which in combination with the lack of a comprehensive, authoritative, easily identifiable set of documentation is perhaps Drupal's biggest flaw
4. I mistakenly thought that the query APIs depended on knowing the content type ("bundle") and not just the entity type ("node" in this case).

Now that I realize that #4 (which I had thought of as the only "real" reason) is wrong, I'll rewrite to use the service.


Thanks to @bryanpizzillo and @BrendanBeltzNIH for catching this and nudging me to revisit it. 👍 Always learning! 😃

docroot/profiles/custom/cgov_site/modules/custom/pdq_core/pdq_core.install
blairlearn: I've entered a draft of how I understand this as #594.

docroot/profiles/custom/cgov_site/modules/custom/pdq_cancer_information_summary/src/Plugin/rest/resource/PDQResource.php
bryanpizzillo: I mean you could have passed the English summary ID in with the spanish and then CDR would not have to know anything about Drupal... However, I can live with this for now.

docroot/profiles/custom/cgov_site/modules/custom/pdq_core/src/Plugin/rest/resource/PDQResource.php
bryanpizzillo: Right, I should have RTFM. Sorry about that - so used to phantom commits... 😃

docroot/profiles/custom/cgov_site/modules/custom/pdq_core/src/Plugin/rest/resource/PDQResource.php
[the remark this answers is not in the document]
bkline: DB query replaced with entity query.

PR #583, opened 2019-01-29
Ticket/486 Homelanding templates and styles

alifr:
Adding templates, styles and library for homelanding content type.


bryanpizzillo:
Can you please merge this into NCIOCPL:ticket/486-landing-paragraphs please?

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/content/home_landing/paragraph--cgov-card-internal.html.twig
```diff
@@ -20,6 +26,9 @@
     {% set description = content.field_override_card_description[0] %}
   {% elseif content.field_featured_item[0]["#node"].field_feature_card_description[0] %}
```
bryanpizzillo: Now that the templates are combined, the description checks need to also check the card type. This current approach makes it too easy for a developer to accidentally include extra fields in their Display views and then introduce a hard to find bug.

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/content/home_landing/paragraph--cgov-card-internal.html.twig
```diff
@@ -6,6 +6,12 @@
  # we will duplicate it.
  #}
 {% if content.field_featured_item[0] %}
+
+  {% set cardType = 'featureCard' %}
+  {% if 'thumbnail' in _context.theme_hook_suggestions[2] %}
```
bryanpizzillo: How solid it this approach? Random array indexes could be problematic. The theme hooks can change. Is there anyway to get at the display mode?

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/content/home_landing/paragraph--cgov-card-internal.html.twig
```diff
@@ -32,5 +41,5 @@
   {% set image = content.field_featured_item %}
 
   {## Render Card ##}
-  {{cardMacros.featureCard(title, description, image, url)}}
+  {{ attribute(cardMacros, cardType, [title, description, image, url])}}
```
bryanpizzillo: Yeah, this DEFINITELY needs a comment. That is not a normal way to use attribute in any Drupal examples and caused some head scratching. (Neat usage though...)

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/content/home_landing/field--paragraph--cgov-column-one.html.twig
```diff
@@ -0,0 +1,6 @@
+{% if label %}
```
bryanpizzillo: Do you have an example of a use for label? otherwise I would say ignore it.

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/content/home_landing/field--paragraph--cgov-column-two.html.twig
```diff
@@ -0,0 +1,6 @@
+{% if label %}
```
bryanpizzillo: See above

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/content/home_landing/field.html.twig
```diff
@@ -0,0 +1,3 @@
+{% for item in items %}
```
bryanpizzillo: Ah HA! so this is how you cleaned up the other field templates. This should most certainly definitely NOT be hidden in the home_landing if we are changing all field templates.

Are we changing all field templates? This should be a conversation for all and not a little side effect -- basically I don't want any regressions. If we do this it should be at the root of the templates. (I am not saying no, I am just saying not here... I certainly have overridden enough of these to warrant it...)

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/content/home_landing/paragraph--cgov-card-external.html.twig
```diff
@@ -4,6 +4,12 @@
  # get the title, description, image and URL from the paragraph fields.
  #}
 
+{% set cardType = 'featureCard' %}
+
+{% if 'thumbnail' in _context.theme_hook_suggestions[2] %}
```
bryanpizzillo: How solid it this approach? Random array indexes could be problematic. The theme hooks can change. Is there anyway to get at the display mode?

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/content/home_landing/paragraph--cgov-card-external.html.twig
```diff
@@ -17,4 +23,4 @@
 {% set image = content.field_image_promotional %}
 
 {## Render Card ##}
-{{cardMacros.featureCard(title, description, image, url)}}
+{{ attribute(cardMacros, cardType, [title, description, image, url])}}
```
bryanpizzillo: Yeah, this DEFINITELY needs a comment. That is not a normal way to use attribute in any Drupal examples and caused some head scratching. (Neat usage though...)

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/src/styles/_homelanding.scss
```diff
@@ -0,0 +1,578 @@
+@import "~Core/styles/environments/percussion";
```
BrendanBeltzNIH: Bryan covered the rest of the bases, so I'll just point out that this new file is a duplicate from the Core partials directory. Since it's really a content type (Entry) specific group of rules, it should probably be in the Homelanding.scss rather than a partial at all. We're not doing Home/Landing split page specific types anymore it would seem, so no partial really needed. Any code that is more generic will probably be best to hoist into the core partials eventually when/if it becomes necessary  (There are some outstanding questions about approach on microsites having access to base partials for overrides, so depending, some of this may need to be abstracted innto a partial again to allow for that. But that's just a future consideration, not a comment for this PR.)

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/content/home_landing/field.html.twig
alifr: Would this template propagate to other content types? I thought, since it's in the homelanding folder, that it would be contained to homelanding content types, but that may not be correct.

This worked as a good field template reset, and then I added field specific alterations, rather than writing templates for every field. After I hit 11 `field--paragraph` templates I thought to give it a try and it worked well.

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/src/styles/_homelanding.scss
alifr: The only difference in this file is I switched out the `~Core/sprites/svg-sprite` for the local `~Styles/sprites/svg-sprite`.

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/content/home_landing/field.html.twig
bryanpizzillo: Yes, by making this it sets the template for all fields everywhere for this theme. Our folders are just for our organization. 

I am not saying it is a bad thing, just a bad place...

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/content/home_landing/field--paragraph--cgov-column-one.html.twig
alifr: I filled the column content fields with placeholder text and made them look nice. In actuality, if we're looking at Cancer Type Landing and News & Events Landing, then there isn't a case for labels. However, I don't see how or why we would render everything in those pages in a RAW HTML field. I say we just leave them as is until we're ready for the next chunk of work, setting those pages up.

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/src/styles/_homelanding.scss
BrendanBeltzNIH: My thought is that there isn't a need for a homelanding partial, what is here in the Common partial should be in Homelanding.scss, until a need to hoist some of it to a Core or Common partial arises.

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/content/home_landing/paragraph--cgov-card-internal.html.twig
alifr: Updated logic to use cardType variable rather than just picking fields available to those card types

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/content/home_landing/paragraph--cgov-card-internal.html.twig
alifr: switched to use `_context.view_mode` which seems like a much more stable attribute

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/content/home_landing/paragraph--cgov-card-internal.html.twig
alifr: Switched back to an `if/else` statement using cardType variable to pick card macro.

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/content/home_landing/field.html.twig
alifr: Well, I think the field should be reset to prevent future headaches. I'd like to hear reasons why it shouldn't be reset as well as where it should live. Not that I have an opinion. I'm still figuring out the project structure.

bryanpizzillo: We agreed it should be overwritten, but moved out (and it was with another PR)

PR #618, opened 2019-01-31
Define PDQ Drug Information Summary

blairlearn:
Closes (#159)

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/pdq_drug_information_summary/README.md
+++ b/docroot/profiles/custom/cgov_site/modules/custom/pdq_drug_information_summary/README.md
@@ -1,0 +1,24 @@
+## Purpose
+This module defines the fields, etc. specific to the PDQ Cancer Information Summary content type.
+
+## Documentation
+
+See [PDQ Content Design](../../../../../../../wiki/PDQ-Content-Design) in the project wiki
+
+## Dependencies
+* `pdq_core` for PDQ workflow and PDQ-shared field definitions.
+* `cgov_core` for general CancerGov-shared field definitions.
+
+## Testing
+**NOTE:** General information about running tests and project structure can be found in the root [TESTING.md](../../../../../../../TESTING.md) file. Please refer to that before reading further.
+
+### Unit Tests
+
+### Functional Tests
+Loads the editing user interface through a web browser (using `Drupal\Tests\BrowserTestBase`) and tests whether the expected fields are present in the rendered UI.
+
+#### Adding Fields
+
+New fields should be added to [PDQSummaryTest.php](tests/src/Functional/PDQSummaryTest.php) with their existence tested in `testSummaryFieldsExist()`.
+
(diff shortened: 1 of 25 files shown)


docroot/profiles/custom/cgov_site/modules/custom/pdq_drug_information_summary/README.md
```diff
@@ -0,0 +1,24 @@
...
+**NOTE:** General information about running tests and project structure can be found in the root [TESTING.md](../../../../../../../TESTING.md) file. Please refer to that before reading further.
+
+### Unit Tests
+
+### Functional Tests
+Loads the editing user interface through a web browser (using `Drupal\Tests\BrowserTestBase`) and tests whether the expected fields are present in the rendered UI.
+
+#### Adding Fields
+
+New fields should be added to [PDQSummaryTest.php](tests/src/Functional/PDQSummaryTest.php) with their existence tested in `testSummaryFieldsExist()`.
```
bkline: I assume you're going to use the same naming convention for the drug summary tests, in which case you'll need to change the file names here. I use the namespace to distinguish one set of tests from another, keeping the same class name inside the name space for the different types of summaries; same thing for the functionality classes, but that doesn't mean you have to do it the same way I do. You should be consistent with your own classes, though.

docroot/profiles/custom/cgov_site/modules/custom/pdq_drug_information_summary/pdq_drug_information_summary.install
```diff
@@ -0,0 +1,30 @@
...
+ * @file
+ * Contains pdq_drug_information_summary.install.
+ */
+
+use Drupal\cgov_core\CgovCoreTools;
+
+/**
+ * Implements hook_install().
+ *
+ * Set up the site for Cancer Information Summaries.
```
bkline: ... for _Drug_ Information Summaries.

docroot/profiles/custom/cgov_site/modules/custom/pdq_drug_information_summary/tests/src/Functional/PDQDrugSummaryTest.php
```diff
@@ -0,0 +1,99 @@
+<?php
+
+namespace Drupal\Tests\cgov_site\Functional;
+
+use Drupal\Tests\SchemaCheckTestTrait;
+use Drupal\Tests\BrowserTestBase;
+
+/**
+ * Tests Functionality of the PDQ Cancer Information Summary content type.
```
bkline: ... PDQ _Drug_ Information Summary ....

And since you're editing this file anyway, you should probably change the class name in the commented-out code at the bottom, as you did for the code in the CIS test file - unless it's never going to be discarded eventually, in which case just ditch it now.

blairlearn:
Updated to fix copy and paste errors.

PR #635, opened 2019-01-31
Ticket/630 alias generation

### LINKED ISSUE #630: Alias generation is Incorrect, Modify Show in Navigation Field
It looks like there is no `/` separator between the ancestors and parent section. 

`/about-cancercoping/feelings` should be `/about-cancer/coping/feelings`.

Separately:

Modify field_show_in_navigation field to have reasonable machine names.

adriancofie:
Closes #630


docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/cgov_site_section.module
```diff
@@ -55,10 +55,10 @@ function cgov_site_section_taxonomy_term_presave(EntityInterface $entity) {
         $values[] = $value;
       }
       $computed_parent = implode('/', $values);
-      $entity->set('computed_path', $computed_parent . '/' . $field_pretty_url);
+      $entity->set('computed_path', '/' . $computed_parent . '/' . $field_pretty_url);
```
bryanpizzillo: So I am just going to point out that this should be handled with an implode..., but I will approve

PR #645, opened 2019-02-04
Feature/cancer center

sarinapadilla:
Merging Cancer Center updates


docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/content/cancer_center/node--cgov-cancer-center--thumbnail-card.html.twig
```diff
@@ -0,0 +1,17 @@
+<div class="row card-thumbnail">
+	<div class="small-2 columns card-thumbnail-image">
```
bryanpizzillo: This should be removed as there is nothing thumbnail card specific here.  The `node--thumbnail-card.html.twig` will do the same.

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/content/cancer_center/node--cgov-cancer-center--feature-card.html.twig
```diff
@@ -0,0 +1,18 @@
+<div class="feature-card">
```
bryanpizzillo: This is not needed - the `node--feature-card.html.twig` can handle this case.

PR #678, opened 2019-02-05
Feature/add list item view modes

bgnih:
Add base List-Item view modes, add Press Release displays and update all fields.  Add YAML_Content for Press Releases (and an Article fix), add first two List templates.  Update PressRelease.field_image_article label to Lead Image


docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/listing/node--list-item-title-date-image.html.twig
```diff
@@ -0,0 +1,19 @@
...
+{% endif  %}
+
+<div class="list-item-image image container">
+  <a class="item-image image" href="{{ url }}">
+    {{ image }}
+  </a>
+</div>
+
+<div class="title-and-desc title desc container">
+  <a class="title" href="{{ url }}" onclick="NCIAnalytics.SearchResults(this,1);s.linkTrackVars='prop57,prop58,prop59,prop60'; s.prop57='Recent Press Releases'; s.prop58='{{ label }}'; s.prop59='Recent Press Releases:1'; s.prop60='www.cancer.gov' + location.pathname.toLowerCase(); s.tl(this,'o','RecentPressReleases')">
```
bryanpizzillo: @bgnih Can you remove the onlick handler please? You would not know this, but those are the analytics for the item's position in the listing. I think we need a better way, as well as I don't want to deal with indices in our templates. We can find a better way for the handler to get attached.

@daquinohd - Do you have and idea about other ways of handling this onclick so it does not need to be hardcoded?

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/listing/node--list-item-title-date.html.twig
```diff
@@ -0,0 +1,9 @@
+{% import '@cgov_common/content_page_macros.twig' as contentPageMacros %}
+<div class="title-and-desc title desc container">
+  <a class="title" href="{{ url }}" onclick="NCIAnalytics.SearchResults(this,1);">
```
bryanpizzillo: The onclick should be removed here.

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/listing/node--list-item-title-date-image.html.twig
bgnih: Done

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/listing/node--list-item-title-date.html.twig
bgnih: Done

PR #686, opened 2019-02-05
Fix for validation casting error. Update form display.

adriancofie:
* Modifying the return value to a bool and casting the potentially null ID to 0


docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/config/install/core.entity_form_display.taxonomy_term.cgov_site_sections.default.yml
```diff
@@ -1,3 +1,4 @@
+uuid: eee7ac49-1f53-43de-a168-f6b1a270cb35
```
bryanpizzillo: What is that?????? Please fix this.

PR #733, opened 2019-02-11
Ticket/591 Add cancer centers theme library and styles.

alifr:
Add cancer centers theme library. Remove whitespace from some outputs.


docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/content/cancer_center/node--cgov-cancer-center--full.html.twig
```diff
@@ -1,82 +1,71 @@
...
-              {% endif %}
               {% if content.field_institution_type %}
-              <br />
-              {{ content.field_institution_type }}
-              {% endif %}
+                <br/>{{ content.field_institution_type }}
+              {%- endif -%}
             </div>
-            {% if content.field_org_head_name %}
             <p>
```
sarinapadilla: I think there should still be an "if" around this p tag - the org_head_name and org_head_title fields aren't required, so they might both be empty. I believe it's best practice to only output the HTML tags if we need them, so we only want to output the p tags if org_head_name has a value. If that isn't the case, then the p tag isn't needed.

alifr: Done.

bryanpizzillo:
@alifr Can you update your branch, the merge from blog posts caused a conflict.

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/content/cancer_center/node--cgov-cancer-center--full.html.twig
```diff
@@ -1,82 +1,73 @@
...
             {% endif %}
             {% if content.field_phone_label_1 and content.field_phone_number_1 %}
-            <p>
-              {{ content.field_phone_label_1 }}: <a href="tel:{{ content.field_phone_number_1 }}" rel="nofollow">{{ content.field_phone_number_1 }}</a>
-            </p>
+              <p>
+                {{ content.field_phone_label_1 }}: <a href="tel:{{ content.field_phone_number_1 }}" rel="nofollow">{{ content.field_phone_number_1 }}</a>
+              </p>
             {% endif %}
             {% if content.field_phone_label_2 and content.phone_number_2 %}
```
sarinapadilla: @alifr There is one more change that I just noticed needs to be made in addition to the merge. The second field here needs to be updated to content.field_phone_number_2.

alifr: Good catch. I've updated the file.

alifr:
@bryanpizzillo upstream/develop has been merged into this branch. Conflicts are resolved.

PR #742, opened 2019-02-12
Ticket/608 Blog Page Styles

alifr:
Updating Blog Page content and styles so layout is much closer to PROD. Updated yaml content and module configuration where necessary to render images, intro text, authors and date. Fixed HTML formatting in intro text. Added a new condition in the `displayDates` macro for the Blog Post date format.


bryanpizzillo:
Imma defer to the reviewers on this one, so when it gets a review shout. I will say I have reservations with the page template, but I started that mess so I can't complain at this time. Stupid regions and node templates...

PR #810, opened 2019-02-17
Ticket/693 cis site sections

### LINKED ISSUE #693: Add site sections to PDQ Cancer Information Summaries
When a cancer information summary is imported into the system, we need to add it to the navigation structure of the site.
* After a summary is imported (or replaced) OR an update changes the url:
   1.  Determine the language of the summary.
   1. Tokenize the summary URL based on `/`. The url `/foo/bar/bazz/placeholder-treatment` would become `['foo', 'bar', 'bazz', 'placeholder-treatment']`
   2. Find the Term in the Site Section vocabulary with the langcode of the summary language AND a parent TID == 0. (i.e. Either the English Home section or the Spanish Home section)
       * This is depth 0 in the nav tree
   2. Set the current section to the root.
   3. For each token of the URL:
       1. determine if there is a child of the current section with a `field_pretty_url` matching the current token. 
           * If there is, set the current section to that child node.
           * If there is not create a new section supplying the token as the `name` (the title field for a term) and the `field_pretty_url`. Then set the current section to that new node
       2. if this is the last token:
          * Set the `name` to the value of the Nav Label (or whatever the CDR calls it) element
          * Set the `field_landing_page` entity reference to the Summary that was loaded.
       3. IF NOT move on to the next token.
* If a summary is updated and the Nav Label is changed:
   1. Find the site section with that URL path. (`computed_path` is a field that has the full path, it may be quicker with that than traversing the tree).  
    1. Change the `name` field.

bkline:
Closes #693 
* add hooks to create site sections for PDQ CIS content
* add English/Spanish home nodes if missing
* add functional tests


docroot/profiles/custom/cgov_site/modules/custom/pdq_cancer_information_summary/pdq_cancer_information_summary.module
```diff
@@ -4,3 +4,123 @@
...
+
+/**
+ * Common code to add site_section terms.
+ *
+ * @param Drupal\Core\Entity\EntityInterface $entity
+ *   Reference to the PDQ Cancer Information Summary just saved.
+ */
+function _pdq_cis_add_site_sections(EntityInterface $entity) {
+
+  // Collect the values we need from the summary.
```
bryanpizzillo: Does the transitioning a summary through the workflow call this code? We had discussed the want of checking what has changed in order to determine if all the queries are needed.

bkline: I'll add an optimizing check in the morning.

bkline: Something like: find the `cgov_site_sections` term with the `computed_path` matching the pdq url field, and if its `name` field (since we're not setting the nav label field) matches the short title of the summary, there's nothing to do in hook. How's that?

One last thing to check: I mimicked what I saw in the sample yaml site sections terms (see `000_nav_root-L1.content.yml`), setting the nav label field for the Spanish root term, but not for the English root term. Just want to confirm that this is correct (for some undocumented reason).

bkline: Another approach I could to take to optimizing away one of the passes through the login in these hooks would be to look at the `status` of the newly saved revision (or its workflow state, which would amount to the same thing), and only create the site sections when the revision was published (in other words, delay until the second revision is created). There's logical incentive for doing it this way, because if in a strange edge case some super-admin came along and edited a summary directly (which the system allows), saving the changes as a draft, it would be more appropriate for the term to match the revision of the summary which is published, particularly since the terms aren't versioned. I thought about doing it this way earlier, but decided instead to follow the logic as written in the ticket for the import/replacement of the summary. Do you have a preference for how I do the optimization?

bryanpizzillo: `$entity->original` will give you the previous version of the entity. (For your status or moderation_state check)

We had sketched something on the white board around this approach, and I do not recall what the outcome was. Do you happen to remember?

bkline: I just pushed a version which does both checks (publishing status of the summary revision and confirmation that the term is already present, with the label unchanged). Using `$entity->original` won't work with this approach, because I'm deferring doing anything with the site sections until I've got public changes to the summary, so the original and current won't be different when it's time to do the work of creating the site sections.

bkline: A reminder that there's still a minor question or two about the navigation label field hanging above.

bkline: A more likely scenario than the one described above in which a super-admin edits a summary directly, saving the changes as a draft, would be a publishing job which fails before all of the summaries have been released from draft to published status. I don't think, based on history, that we can afford to dismiss that as an edge case. :-)

bkline: Ah, this is why we run the tests. 😃 I need to modify the tests to defer looking for the site section until after the release to publishing. In fact, I'm thinking I should probably expand the test to do another save with changes to the summary's short title, and verify that the site section doesn't get updated until after that change has been published.

PR #951, opened 2019-03-01
Set Drupal timezone

### LINKED ISSUE #950: Bug: Timezone should be set to New York (US Eastern)
## Description 

In the Acquia environment, the site Timezone is set to UTC.  That's because for local development, we're setting it in the Docker environment and letting Drupal pick up the system default.

Do the thing.

* Remove timezone setting from the two Dockerfiles.
* In the php.ini files, set timezone to "Etc/UTC" (matches the OS default and should prevent warning messages about the timezone not being set).
* Set timezone in `system.date` `timezone` property.

blairlearn:
- Set system.date in cgov_site.
- Revert containers to OS-default timzeone (Etc/UTC).
- Specify Etc/UTC for Php INI files in order to prevent timzeone warnings.

Closes #950


blairlearn:
Reworking some bits

PR #939, opened 2019-02-28
Ticket/425 citations common field

sarinapadilla:

PR #966, opened 2019-03-05
Ticket/796 bio press info

sarinapadilla:


docroot/profiles/custom/cgov_site/modules/custom/cgov_biography/config/install/core.entity_view_display.node.cgov_biography.default.yml
```diff
@@ -59,6 +60,16 @@ content:
     region: content
     settings: {  }
     third_party_settings: {  }
+  field_bio_press_info:
```
bryanpizzillo: Please change this to be a dropdown select (Do Not Display vs Display). It should be an integer list with Display being 1 and Do not being 0. @VictoriaSunNIH Can you please update the requirements.  (Checkboxes get hidden in the UI.)

Also can you make the field name something like field_display_bio_press_info? So it is more of a question.

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/content/biography/node--cgov-biography--full.html.twig
```diff
@@ -68,15 +68,9 @@
...
-
-          {# TODO: replace placeholder below with customizable content #}
-          <div class="press-contact">
-            <h2>Press Contacts</h2>
-            <p>To request an interview with NCI researchers, contact the NCI Office of Media Relations.<br />
-              <a href="mailto:ncipressofficers@mail.nih.gov" class="press-contact-email">ncipressofficers@mail.nih.gov</a><br />
-              <span class="press-contact-phone">240-760-6600</span>
-            </p>
-          </div>
+          {% if content.field_bio_press_info[0]['#markup'] == "On" %}
```
bryanpizzillo: this could just be the following?
`content_field_bio_press_info | field_value`

0 should be false and anything not zero is true.

BrendanBeltzNIH: +1

VictoriaSunNIH:
@sarinapadilla  - the dropdown is fine, but there is a "None" option that shouldn't be there.  Please update so that it's only Display or Do Not Display.  Default to Do Not Display.

VictoriaSunNIH:
@lburack  - here's the ODE for review: https://github.com/NCIOCPL/cgov-digital-platform/pull/939

lburack:
This passes my IA review. Thanks @sarinapadilla !

PR #991, opened 2019-03-07
Ticket/928 event page look and feel

arcepaul:
Look and feel for Event Page
Includes change from promotional image to lead image 

TODO: Figure out next steps for Outlook Calendar link creation


docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/src/entrypoints/event/Event.scss
```diff
@@ -0,0 +1,9 @@
+@charset "utf-8";
+@import "~Core/styles/variables";
+@import "~Core/styles/breakpoints";
+
+.event-location {
+    &> span {
```
BrendanBeltzNIH: I don't believe you need to have the & operator here as the > child operator doesn't need to be appended directly to the parent to have the effect of limiting to one generation in the hierarchy. Have you tried it without?

bryanpizzillo:
This should be on http://ncigovcdode45.prod.acquia-sites.com/

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/src/entrypoints/event/Event.scss
arcepaul: Yeah, the '&' is implied.  I'll remove it.

arcepaul:
@VictoriaSunNIH @lburack 
Example Event page available at:
http://ncigovcdode45.prod.acquia-sites.com/about-cancer/coping/feelings/relaxation/2019-investigators-site

VictoriaSunNIH:
@arcepaul , @sarinapadilla - If i select a different end date for an all day event, I am not seeing it display on the page (I'm only seeing the start date). (tagging @lburack for awareness)

Referring to https://lm1a12.axshare.com/#g=1&p=event_listing_no_past_events - if we have selected both a start and end date, and selected all day event, we should see Month DD, YYYY - Month, DD YYYY

arcepaul:
@VictoriaSunNIH @lburack 
Yes, both 1014 and 985 will be addressed separately from this PR.

lburack:
It looks like the date is working for me. I wonder if @VictoriaSunNIH and @sarinapadilla are experiencing cache issues..? Either way, I selected two different dates and times for my start and end times and am seeing only the dates (and not the times) on the page, as expected. The only issue is that the display should say "March 21 - March 22, 2019" (not March 21-22, 2019)

I put in a change request for the Room display though and tooltip helper text #1027

http://ncigovcdode45.prod.acquia-sites.com/research/lbb-event

VictoriaSunNIH:
@arcepaul , @sarinapadilla  - agreed with lindsay that dates are suddenly working.  I added the comment about the display to the separate date ticket (#1014 ).  Otherwise this passes review!

sarinapadilla:
@VictoriaSunNIH @lburack the changes for #1014 (which will actually be included in this PR and I'll create a new ticket for the back-end updates to the dates that Bryan wants) and #1027 are now included in this PR!

VictoriaSunNIH:
@lburack - this looks good to me!  Passing to you for final review

lburack:
Looking good to me too! I did put in a change request #1059 for an edge-case I found with the date display. But I'm marking this as passed IA Review.

PR #1013, opened 2019-03-11
Feature/cards promo

daquinohd:


daquinohd:
Will reopen when the field is on a Blog Series for testing.

PR #1095, opened 2019-03-19
Ticket/298 pr feed

### LINKED ISSUE #298: Create Press Release RSS Feed
This story is to create the RSS Feed for [Press Release](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Content-Type-Press-Release) #155.

* RSS Feed should allow "special characters" (ampersand, less-than, or greater-than) in the RSS feeds.
* All automated RSS feeds are updated every hour (at 20 minutes after the hour), and republished every hour between the hours of 08:00 to 20:00
* On an RSS Feed, the following information from each content item in the RSS feed should appear:
   * The Page Title of each content item
   * The list description of each content item 
      * If no list description exists, then use the meta-description
   * Thumbnail image associated with each content item

bgnih:
Fixes #298.   

TESTING: On http://ncigovcdode15.prod.acquia-sites.com/

EN: http://ncigovcdode15.prod.acquia-sites.com/PublishedContent/RSS/syndication/rss/ncinewsreleases.rss
ES: http://ncigovcdode15.prod.acquia-sites.com/espanol/PublishedContent/RSS/syndication/rss/nci-noticias.rss 

`<description>` uses List Description (if present) or Meta Description.
`<enclosure>` uses last image of:
1. Lead Image
1. Lead Image thumbnail override
1. Promo Image
1. Promo Image thumbnail override

and renders using the `cgov_thumbnail` style. 

[Known issues](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Drupal-Issues-Watchlist#rss-feeds):   
1. The date format is different in this version than on current PROD.   This version is correct, PROD is not using the proper date format, per the [RSS spec](http://www.rssboard.org/rss-specification).
1. This version uses "escaped encoding" instead of the "CDATA encoding" that current PROD version uses.   Either is fine per the RSS spec.
1. This version uses `en` and `es` as the `channel.language` value.  The current PROD uses `en-us` and `es-us`.  Both are permitted by the RSS spec.
1. The Spanish page will needs a redirect from `/PublishedContent/RSS/syndication/rss/nci-noticias.rss` to `/espanol/PublishedContent/RSS/syndication/rss/nci-noticias.rss` to support existing feeds.  (node renders in EN without the /espanol/ prefix).  This can be done with an Akamai reroute, or other solution.


docroot/profiles/custom/cgov_site/modules/custom/cgov_core/src/CgovCoreTwigExtensions.php
```diff
@@ -0,0 +1,170 @@
...
+        $media_image_thumbnail_file = $promo_media_field->get('field_override_img_thumbnail')->entity;
+        if ($media_image_thumbnail_file) {
+          $displaying = 'promo thumbnail override';
+          $media_image_file = $media_image_thumbnail_file;
+        }
+      }
+    }
+
+    // Get URI of media image file (eg: public://2019-03/image.jpg).
+    $image_uri = $media_image_file->getFileUri();
```
daquinohd: Can you add a null check for `$media_image_file` here?

docroot/profiles/custom/cgov_site/modules/custom/cgov_core/src/CgovCoreTwigExtensions.php
```diff
@@ -0,0 +1,170 @@
...
+    // Image style to use when rendering images.
+    $thumbnail_image_style = 'cgov_thumbnail';
+
+    // Load the node.
+    $node = $this->entityTypeManager->getStorage('node')->load($nid);
+    if (!$node) {
+      return FALSE;
+    }
+
+    // Check for available images to display.
```
daquinohd: Does RSS follow the same rules as promo content where the promotional image is prioritized over the article image? And if so, I'd say update the condition to an if-else so that we don't need to create two images.

docroot/profiles/custom/cgov_site/modules/custom/cgov_core/src/CgovCoreTwigExtensions.php
bgnih: It should never be null, as the file field is required and it has already checked for the presence of the Image media entity above.

docroot/profiles/custom/cgov_site/modules/custom/cgov_core/src/CgovCoreTwigExtensions.php
bgnih: Yes, it does.  I generally prefer the 'set a default and replace it conditionally' approach is it reduces the chances of developers missing an edge-case with the if-then-else, is a little easier to follow (at least I think it is), and is effectively just as efficient.  But if you want me to change it,  I will.

VictoriaSunNIH:
@bgnih - I can't seem to access http://ncigovcdode15.prod.acquia-sites.com/PublishedContent/RSS/syndication/rss/ncinewsreleases.rss

I created a new press release, and wanted to see if it shows up on the feed.

VictoriaSunNIH:
Looks good to me (I added a PR, saw it appear in the RSS feed for both English and Spanish).  Also compared the feed to Prod and it looks right.  Thanks!

VictoriaSunNIH:
Moving this back - need to test it on a translated piece of content before passing.

PR #1110, opened 2019-03-20
Fixes #1054 Duplicated records with same pretty-url

### LINKED ISSUE #1054: Bug: When editing a press release, I can't save my edit (error message: pretty URL is already existing)
When logged into Admin role, I am finding this bug on dev, where I can't save changes to an existing press release. This does not happen when logged in as editor or author.

How to reproduce:

1. Edit a press release.  
2. Add some content to one of the fields
3. Go to the bottom, and save the edit

You will get this screen: 
image

bgnih:
Fixes #1054 Caused by duplicate node being added in both `013_press_release_chemotherapy.content.yml` and `press_release.content.yml`.  Removed from `press_release.content.yml`.

Note: This was allowed by a lack of validation on yaml_content imports.   We should add an 
```
$violations = $entity->validate();
// process violations here
```
to the ENTITY_PRE_SAVE event in cgov_yaml_content module (or patch the yaml_content module to do it).

TESTING: On http://ncigovcdode58.prod.acquia-sites.com


VictoriaSunNIH:
I'm still seeing this issue on press releases...

VictoriaSunNIH:
Working as expected!

PR #1112, opened 2019-03-20
Implement Glossification in WYSIWYG

bryanpizzillo:


docroot/profiles/custom/cgov_site/modules/custom/cgov_core/config/optional/editor.editor.streamlined.yml
```diff
@@ -29,12 +29,15 @@ settings:
         -
           name: Formatting
           items:
-            - Format
```
BrendanBeltzNIH: Should Format be taken out? Might need to revert that.

docroot/profiles/custom/cgov_site/modules/custom/cgov_ckeditor/css/old.css
```diff
@@ -0,0 +1,93 @@
+* {
```
BrendanBeltzNIH: This needs to get removed

docroot/profiles/custom/cgov_site/modules/custom/pdq_glossifier/js/plugins/cgov_glossifier/dialogs/reference/old.js
```diff
@@ -0,0 +1,690 @@
+/**
```
BrendanBeltzNIH: This needs to be removed.

VictoriaSunNIH:
@BrendanBeltzNIH , @lburack - Glossification is looking good.  I know there is the outstanding issue of the modals not working (once I've glossified the term and go to look at the page on the front end and click the link, it goes to a 404 since its linking to http://ncigovcdode59.prod.acquia-sites.com/Common/PopUps/popDefinition.aspx?id=CDR0000650566&version=Patient&language=en, but I know this is being investigated.

Another thing to note: The glossification window itself is rather large.  It forces me to have to scroll my page window, to the be able to continue to scroll in my glossification window, which is rather annoying.  This also means that when i close out of the glossification window, my underlying page is no longer scrolled to the body section, but somewhere else.  Any way we can make this window smaller?  I've put in a change request (#1167).

lburack:
@BrendanBeltzNIH  — Glossification is looking good on English. But Spanish terms don't seem to be in place. I translated a page and created a new Spanish page, and tried to glossify terms as is on the live Cgov page. Pretty much all of the terms do not show up as options to glossification. I double checked and the spelling is the same for the word in the content and the term in the Spanish dictionary.

I'm not having the same issue @VictoriaSunNIH reported in her comment above. I am able to scroll only within the glossification window. The form behind the glossification window does not move unless I move my mouse and scroll outside of the glossification window. @VictoriaSunNIH  - maybe you can show me this tomorrow on your computer.

BrendanBeltzNIH:
Yes, you are correct about the links not working. I need to fix that. Did I put the IA review tag on this yet? I didn't think I had. ;)

VictoriaSunNIH:
@lburack  - this ODE: http://ncigovcdode59.prod.acquia-sites.com

VictoriaSunNIH:
Unfortunately I must fail again...due to the following example (which @bkline  and @BrendanBeltzNIH ) are aware of:

On https://www.cancer.gov/types/childhood-cancers, Under Types of Cancer in Children,
central nervous system (CNS) tumors comes as one glossified term in percussion, but in drupal it comes as 3 terms, and there is no choice to select it as central nervous systems tumors

In Drupal:
image


In Percussion:
image

lburack:
@BrendanBeltzNIH @bkline @VictoriaSunNIH  – the central nervous system (CNS) tumors example above isn't a bug. Content owners will often ask us to glossify a term that doesn't exactly match what's in the dictionary. This is one of those cases where the link text was edited after glossification.

This passes IA review for me in both English and Spanish. Thanks!

Issue #1109, opened 2019-03-20
BUG: 'List Description' field should be unlimited size.


blairlearn:
The [List Description](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Common-Fields#list-description) common field is supposed to have an unlimited size.  It is currently set to a maximum of 320 characters.

Do the thing.

*  In `cgov_core`, edit `field.storage.node.field_list_description.yml1
   * Set type to `string_long`.
   * Remove `settings.max_length`
*  In `cgov_media`, edit `field.storage.node.field_list_description.yml1
   * Set type to `string_long`.
   * Remove `settings.max_length`

VictoriaSunNIH:
Closing this - I've referenced this issue in a number of places for each content type that has the wrong size.

PR #1127, opened 2019-03-21
Implement Infographic

blairlearn:
Closes #219 and #1109

Issue #592, opened 2019-01-30
Implement front-end display of Infographic


VictoriaSunNIH:
This story is to implement front-end display of the Infographic page-level content type.  

Annotations for infographic added https://hzgfzw.axshare.com/#g=1&p=infographic. This is also linked on the [wiki](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Media-Type-Infographic)

PR #1129, opened 2019-03-21
Implement Infographic

### LINKED ISSUE #219: Create Infographic Multimedia Content Type
This story is to create the [Infographic](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Media-Type-Infographic) Multimedia content type. 

Example URLs [here](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Media-Type-Infographic#example-urls)

#### I am required to enter the following common fields:

- [Page Title](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Common-Fields#title-fields)
- [Short Title](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Common-Fields#Short-Title)
- [Meta Description](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Common-Fields#Meta-Description)
- [Posted Date](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Common-Fields#Content-Dates)
- [Updated Date](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Common-Fields#Content-Dates)
- [Reviewed Date](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Common-Fields#Content-Dates)
- [Search Engine Restrictions](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Common-Fields#Search-Engine-Restrictions)

#### I have the option to enter the following common fields:

- [Card Title](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Common-Fields#Card-Title)
- [Browser Title](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Common-Fields#Browser-Title)
- [List Description](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Common-Fields#List-Description)
- [Feature Card Description](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Common-Fields#Feature-Card-Description)
- [Pretty URL](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Common-Fields#Pretty-URL)
- [Date Display Mode](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Common-Fields#Content-Dates)
- [Body](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Common-Fields#Body)
- [Promotional Image](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Common-Fields#Promotional-Image)
- [Public Use Text](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Common-Fields#Public-Use)
- [Syndicate](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Common-Fields#Syndicate)
- [Syndication Keywords](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Common-Fields#syndication-keywords)

#### I am required to enter the following local fields:

- [The Actual Infographic](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Media-Type-Infographic#the-actual-infographic)

#### I have the option to enter the following local fields:
- [Accessible Text Version](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Media-Type-Images#accessible-text-version)

* The Infographic will automatically have an enlarge button which will allow the user to select and see the full sized infographic.

**Tasks/Subtasks:**
- [x] Add workflow transitions for Infographic Multimedia page type. 
- [x] Add content author and content reviewer permissions for Infographic Multimedia page type. 
- [x] Create Local fields
- [x] Attach all fields

blairlearn:
Closes #219, #1109, and #1132.

@VictoriaSunNIH and @lburack: The code is deployed to http://ncigovcdode61.prod.acquia-sites.com

**NOTE:** This has _not_ been styled beyond copying markup from production.


VictoriaSunNIH:
@blairlearn, @lburack - some things i'm noticing:

- Feature Card Description should be a single row, and only allow a max of 255 characters (I wasn't sure if this is supposed to be a separate ticket, but put it in.  Let me know if this isn't just on Infographic https://github.com/NCIOCPL/cgov-digital-platform/issues/1132)
- Site section should be using the site section browser
- Promotional Image should be using the image browser
- Date display mode should be checkboxes that allow multi select

blairlearn:
Feature card impacts infographic and video.  I'll take care of both.

blairlearn:
Updated to fix:

* Feature Card Description (#1132) for Infographic and Video.
* Site section browser (Infographic only)
* Promotional Image
* Date display mode  for Infographic and Video.

VictoriaSunNIH:
@blairlearn , @lburack based on #1127 this passes IA review.

VictoriaSunNIH:
Removing the label for now - we need to implement and test #592 in order to call this passed IA review. For now - the backend has passed review!

arcepaul:
@lburack @VictoriaSunNIH 
#592 is now merged onto this PR and available for testing on http://ncigovcdode61.prod.acquia-sites.com.

Note: the sidebar navigation will not work/appear until Adrian has updated it to work with 'media'

lburack:
Hi @blairlearn, two issues:
- the body field should be the full wysiwyg by default. Right now it has a dropdown where the user can select which wysiwyg type to use.
- I created an English infographic and published it. I then tried to translate that infographic, but when I go to save the Spanish translation, I get an invalid transition status error. Apparently @bkline worked on this translation workflow issue before, so hopefully he can offer some insight on how to fix!

blairlearn:
@bkline - I've updated the `feature/infographic` branch with the latest from `develop` and the problem persists. Nothing shows up in the watchdog log for it. Can you take look?

bkline:
Hi, @blairlearn, I've been wrestling with this bug most of the day (what you said about the branch only being three commits behind `develop` was right, which meant that the patch to the core bug which was clobbering the transition for `node` content had already been applied to this ODE, and yet saving new media content translations still failed, whereas saving new `node` content translations succeeds with my patch to core Drupal). I believe I have unraveled most, but not all of the mysteries behind this latest bug.

There are two problems in play here. One is yet another bug in Drupal core. When an entity which is subject to content moderation (or its translation) is about to be saved, the `content_moderation` module determines what moderation state we're starting from and which state the user wants the entity/translation to be in after the save, and then asks the question "do the moderation rules for this content type allow this state transition?". When the entity (or translation) is being saved for the first time, the state we're starting from is determined for this exercise by asking the moderation to tell us what the starting state is for a new entity (or translation). Here is the method which answers that question.

```
  public function getInitialState($entity = NULL) {
    // Workflows are not tied to entities, but Content Moderation adds the
    // relationship between Workflows and entities. Content Moderation needs the
    // entity object to be able to determine the initial state based on
    // publishing status.
    if (!($entity instanceof ContentEntityInterface)) {
      throw new \InvalidArgumentException('A content entity object must be supplied.');
    }
    if ($entity instanceof EntityPublishedInterface && !$entity->isNew()) {
      return $this->getState($entity->isPublished() ? 'published' : 'draft');
    }

    return $this->getState(!empty($this->configuration['default_moderation_state']) ? $this->configuration['default_moderation_state'] : 'draft');
  }
```

Notice that if the entity is of a type which implements the `EntityPublishedInterface` and the entity is "new" (we'll discuss what "new" means in this context below), then the method arbitrarily says the starting moderation state is `published` or `draft`, depending on whether the entity (or translation) is published.

So first of all, that's wrong because it assumes it can know not only which moderation states we've said are valid states for this content type, but also that it can know which state we've set as the starting state for this content moderation rule set. I might possibly be willing concede that this approach could be a reasonable fallback if the moderation rule set is unable to figure out what we have set as the initial state for new entities, but only reluctantly, because if it can't figure that out, then the core `content_moderation` module is even more badly broken than it already is, and I'd be more inclined to throw an exception, in order to prevent such a broken release from getting into production.

This bug is made worse by the second problem I ran into. For `node` content, that code which is making assumptions it shouldn't be making gets it right by sheer luck, because we have indeed set `draft` as our starting moderation state for this rules set, and the answer to the `isPublished()` test comes back correctly as `FALSE`. (This is why I didn't notice this new bug until today.) However, for some reason which I have not yet been able to track down, `media` content entities have that published flag set to `TRUE`, even for the new translation (even if I set it to `FALSE` in my `entity_translation_create` hook). Perhaps @bryanpizzillo can help us track down where/why that is happening for `media` content entities, but not for `node`s.

Even without understanding what's causing that second problem, I have been able to make the unwanted behavior go away with yet another patch to Drupal core, which moves down to the bottom of the method that pushy code which is assuming it knows what our initial state is, and I'm instead checking to see if the moderation configuration can tell us which initial moderation state we have chosen for our default, and if it can (which is always true), we use that state, and our validation of the proposed transition passes. With my patch, the fixed method looks like this.

```
  public function getInitialState($entity = NULL) {
    // Workflows are not tied to entities, but Content Moderation adds the
    // relationship between Workflows and entities. Content Moderation needs the
    // entity object to be able to determine the initial state based on
    // publishing status.
    if (!($entity instanceof ContentEntityInterface)) {
      throw new \InvalidArgumentException('A content entity object must be supplied.');
    }
    if (!empty($this->configuration['default_moderation_state'])) {
      return $this->getState($this->configuration['default_moderation_state']);
    }
    $fallback = 'draft';
    if ($entity instanceof EntityPublishedInterface && !$entity->isNew() && $entity->isPublished()) {
      $fallback = 'published';
    }
    return $this->getState($fallback);
  }
```

If Bryan gives me the green light to do so, I'll file another bug report against Drupal core and include my new patch, and I'll modify `composer.json` to apply the patch to our builds.

Here's my postscript where I talk about what "new" means in the context of translations, as I promised to do above. The `isNew()` method is a little funky, because even though the object on which it is invoked represents a translation in this case, it's answering the question about whether the whole entity (in any language) is new, which it does by seeing whether the object has an entity ID. The reason I say that's funky is that in this context what the logic should really be looking at is whether this particular translation is new. There's a way to make `isNew()` return `TRUE`, which is by calling the poorly documented `enforceIsNew()` method on the translation's object, but that backfires, because in addition to bypassing the buggy code we're trying to avoid, it causes Drupal to try to save the translation in a way which violates database uniqueness constraints. This is another example of the Drupal core team going wild with the idea of thinking of everything as an `entity`, and getting deeply confused by all of these `$entity` variables floating around, and forgetting what each of them represents.

bkline:
I have crawled further into the Drupal cave, in an effort to track down why the `media` content translation reports `TRUE` for `isPublished()`. When the form prepares to validate the submission, it builds up the object representing the entity being saved. Part of that process involves sweeping through the fields on the form and plugging in the results of that sweep into the object. There's a field on the form for an infographic (and it seems likely that this will turn out to be true for any `media` entity) which isn't on the create translation form for `node` content.

```
<!-- BEGIN OUTPUT from 'core/themes/classy/templates/form/input.html.twig' -->
  | <input data-drupal-selector="edit-content-translation-status" aria-describedby="edit-content-translation-status--description" disabled="disabled" type="checkbox" id="edit-content-translation-status" name="content_translation[status]" value="1" checked="checked" class="form-checkbox" />
  |  
  | <!-- END OUTPUT from 'core/themes/classy/templates/form/input.html.twig' -->
```

Note that not only is the `status` field set to `TRUE` (represented as the string "1" here), but the field is disabled, so the user can't even fix the problem. I haven't (yet) figured out why that's showing up for non-node content, when node content doesn't have this monkey wrench. One of the reasons I'm posting updates here is provide opportunities for the PMs to tell me to stop, and to crawl back out of the cave. Although it's annoying and confusing to have this silly uneditable field on the `media` content forms, if we use my patch to fix the first of the two bugs described above, we can ignore that field and have the proposed moderation state of `draft` accepted. In the meantime, I'll continue with this unpleasant muddy spelunking until I'm told to stop.

lburack:
As noted on #1310, the translation fix for infographics is working! The only issue is that the infographic isn't being placed in the assigned site section, with its given pretty URL. We need to apply the site section fix that was just pushed to dev for other media items to the infographic.

bryanpizzillo:
@lburack Can you check the URLs please? Blair added them in.

docroot/profiles/custom/cgov_site/modules/custom/cgov_infographic/cgov_infographic.routing.yml
```diff
@@ -0,0 +1,9 @@
+# Set up route for retrieving an infographic's long description text.
+cgov_infographic.long_description:
+  path: '/cgov_site/cgov_infographic/long-description/{node}'
```
bryanpizzillo: We discussed this following the Drupal pattern of `/{entity}/id/{function}`, as in `/media/{media}/long-description`.  Did you find out this would not work, or there were other issues?

No matter what, `{node}` would be inappropriate in this case anyway since an infographic is of type media. So `{node}` needs to be `{media}`.

It looks like if you add the following to the route (after `requirements:`, it can load the entity for you:
```
  options:
    parameters:
      media:
        type: entity:media
```
If you truly just want the `{media}` to represent the id, then you can also add a requirement to validate the ID of:
```
requirements:
   _permission: 'access content'
  media: \d+
```

Finally, and this can be a backlog ticket, we should see if there is a requirement that says, only do this if the media has the accessible text field.

docroot/profiles/custom/cgov_site/modules/custom/cgov_infographic/src/Controller/CGovInfographicController.php
```diff
@@ -0,0 +1,67 @@
...
+   */
+  public function longDescription($node) {
+
+    // Check for invalid inputs.
+    if (!is_numeric($node)) {
+      throw new NotFoundHttpException();
+    }
+
+    // Load the entity, and verify that it's an infographic.
+    $infographic = $this->entityTypeManager()->getStorage('media')->load($node);
```
bryanpizzillo: if you use the "upcasting" configuration as defined in https://www.drupal.org/docs/8/api/routing-system/parameter-upcasting-in-routes and referenced in my comment on the routing.yml file, then this is not needed.  Once again, `getStorage('media')->load($node)` should never be written again, ever, never ever ever.

docroot/profiles/custom/cgov_site/modules/custom/cgov_infographic/src/Controller/CGovInfographicController.php
```diff
@@ -0,0 +1,67 @@
...
+   * {@inheritdoc}
+   */
+  protected function getModuleName() {
+    return 'cgov_infographic';
+  }
+
+  /**
+   * Return an infographic's long description in an HTTP Response.
+   *
+   * @param int $node
```
bryanpizzillo: This either should be `$media` or `$entity` calling a media item as a node hurts my head greatly. I am sure @bkline would agree on this one. I would say media as it is only for media.

Additionally, if it is an id, then $node DEFINITELY should not be the variable name. I thought this was the media entity until I got further down in the code.

docroot/profiles/custom/cgov_site/modules/custom/cgov_infographic/src/Controller/CGovInfographicController.php
blairlearn: Lots of finicky bits, but that's in place now.

docroot/profiles/custom/cgov_site/modules/custom/cgov_infographic/src/Controller/CGovInfographicController.php
blairlearn: Changed to $media.

docroot/profiles/custom/cgov_site/modules/custom/cgov_infographic/cgov_infographic.routing.yml
blairlearn: > we should see if there is a requirement that says, only do this if the media has the accessible text field

There was a Slack conversation with @lburack regarding required fields vs. governance for making sure there was either a body or an accessible text version.  Acting on the assumption that it wouldn't make sense to link to a blank long description, I coded the longdesc attribute to only be generated if accessible text existed.   (If you try to access a non-existant accessible text version, you'll get a 404.)

lburack:
@bryanpizzillo @blairlearn – the URLs are now showing as expecting. Thanks for fixing this!

blairlearn:
Ugh.  I just realized, the logic for setting up and resolving the longdesc attribute (accessible text field) is in the wrong place.  I've entered #1361.

blairlearn:
There's also some funkiness in the workflow which I haven't investigated yet to determine whether it's particular to infographics or everything else too.

1. Create a new infographic
1. Fill out the fields as usual
   * Set the site section to `/about`.
   * Set the pretty url to `pretty`.
1. Set the publishing state to `Published`.
1. The infographic's path will be `/node/{id}` (And it really is `/node`, not `/media`)
1. Edit the infographic.
1. Set workflow state to `Editing`.
1. Click "Save".
1. Edit the infographic again.
1. Set workflow state to `Published`.
1. Click "Save".
1. The infographic's path will only *now* be `/about/pretty`.

bkline:
Re-rolled patch for Drupal core bug #3045648 applied to address the use case of content which existed before the `content_moderation` module was installed and enabled.

https://www.drupal.org/files/issues/2019-04-07/3045648-15.patch

PR #1173, opened 2019-03-24
Added Inline Linking using Linkit

### LINKED ISSUE #1165: Implement Managed Linking Within the WYSIWYG
As an author I should be able to select some text, find a page on the site to link to, and insert an anchor around the selected text.

This link should automatically update the href whenever the page changes its URL.

bryanpnih:
This adds inline linking to all node content, and videos. This will allow a user to find the page they want to link to via its URL or title.

This closes #1165. 

ODE: ncigovcdode67.prod.acquia-sites.com

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,6 +40,7 @@
         "drupal/entity_browser": "^2.0",
         "drupal/entity_embed": "1.x-dev",
         "drupal/image_widget_crop": "^2.2",
+        "drupal/linkit": "^5.0",
         "drupal/metatag": "^1.8",
         "drupal/page_manager": "^4.0@beta",
         "drupal/panels": "^4.3",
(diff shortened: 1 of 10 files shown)


lburack:
@bryanpizzillo just informed me that inline linking cannot use a content browser, so the typeahead is working as expected for both admins and editors. Marking this as Passed IA Review.

I will say the functionality isn't so great for external links — the user has to select the dropdown option of "LinkIt could not find any suggestions. The URL will be used as is." If you press enter instead of purposefully selected that dropdown, it will tack your URL onto the path of the page you're editing (e.g., http://ncigovcdode67.prod.acquia-sites.com/about-cancer/coping/nytimes.com) This is something we'll need to make sure is clear in training @VictoriaSunNIH @mltnewman

PR #1214, opened 2019-03-26
More list styling

arcepaul:
- thumbnail list hovers are now independent, no dirty js
- dynamic lists given some styling traits of managed lists
- small styling tweaks to match prod


arcepaul:
I went for the former and classed items with thumbnails with 'has-media', allowing the styling to be used across dynamic and managed lists

bryanpizzillo:
With the .has-media addition, can you also remove the conditional around the .has-images in https://github.com/NCIOCPL/cgov-digital-platform/blob/feature/lists/docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/listing/paragraph--cgov-list.html.twig

PR #1240, opened 2019-03-27
Ticket/1034  order section nav by weight

BrendanBeltzNIH:
Update SectionNav child rendering to respect the weights of Terms by first performing a usort on the list.

ODE: http://ncigovcdode72.prod.acquia-sites.com


lburack:
I am VERY happy to report that this passes IA Review!

PR #1299, opened 2019-04-01
Implement Embedded Feature Cards

bryanpnih:
TBD - Just getting a build to start


arcepaul:
@lburack @VictoriaSunNIH 
ODE is available at http://ncigovcdode83.prod.acquia-sites.com/

lburack:
I tested this on the following content types using image and no image, as well as playing around with the alignments:
- article
- mini landing page
- blog post
- event
- press release

This passes IA review, but just to note, I created a backlog ticket (can happen post-migration) to look into removing the alignment options of 'None' and 'Center' since we don't have them now and they look weird on the front-end #1379

PR #1359, opened 2019-04-05
Feature/embedded feature card

arcepaul:


arcepaul:
@bryanpizzillo 
This is just to get your eyes on this.  I tried to minimize the number of changes this go around.  One question I had though, will topic feature ever show up now or will it always be an embedded entity?

bryanpizzillo:
>This is just to get your eyes on this. I tried to minimize the number of changes this go around. One question I had though, will topic feature ever show up now or will it always be an embedded entity?

I don't follow the question exactly, so I may be incorrect in my response. On a mini-landing page (e.g. /about-nci/organization/ccct (IIRC)) you will add the feature cards through Paragraphs, not the WYSIWYG.  

There will be Articles, Blog Posts, Content Blocks on Topic Pages, and others that will have Feature Cards inserted through the WYSIWYG.

Does that answer the question?

arcepaul:
@bryanpizzillo 
Kind of.  I think I was trying to nail down that feature cards would only be added as embedded entities, and that things like 'topic-feature' couldn't be added otherwise.

arcepaul:
@bryanpizzillo 
I just added mobile styling.  This is now ready to go up to ODE when you're ready.

PR #1365, opened 2019-04-08
OTP migration edge case handling.

BrendanBeltzNIH:
In the new system, OTP is drawn through twig. In the old system, it's done by NCI.page.js. Migrated articles may not be correctly transformed into the new paragraphs and instead be dumped raw into a  single paragraph. That edge case needs to be handled. NCI.page.js still runs and will correctly draw the OTP when the twig template fails to do so.

ODE: http://ncigovcdode82.prod.acquia-sites.com


lburack:
I copied our OTP code from percussion and pasted it into the first content wysiwyg (in source view). I saved the page/republished and OTP is displaying correctly. Marking this as Passed IA Review!

Issue #818, opened 2019-02-19
Front-end Buildout of Blue Microsite Palette


VictoriaSunNIH:
Ticket to build out the microsite blue palette.

This includes the blue palette, and use of the shared utility shades. See attachments.
[NCI-Microsites-Blue-Palette.pdf](https://github.com/NCIOCPL/cgov-digital-platform/files/2880189/NCI-Microsites-Blue-Palette.pdf)
[NCI-Microsites-Utility-Shades.pdf](https://github.com/NCIOCPL/cgov-digital-platform/files/2880190/NCI-Microsites-Utility-Shades.pdf)
[style-elements-microsite-themes-hex.xlsx](https://github.com/NCIOCPL/cgov-digital-platform/files/2880191/style-elements-microsite-themes-hex.xlsx)

Example: https://www.cancer.gov/nci/pediatric-adult-rare-tumor

Issue #817, opened 2019-02-19
Front-end Buildout of Microsite Purple Palette


VictoriaSunNIH:
Ticket to build out the microsite purple palette.

This includes the purple palette, and use of the shared utility shades.  See attachments.
[NCI-Microsites-Purple-Palette.pdf](https://github.com/NCIOCPL/cgov-digital-platform/files/2880182/NCI-Microsites-Purple-Palette.pdf)
[NCI-Microsites-Utility-Shades.pdf](https://github.com/NCIOCPL/cgov-digital-platform/files/2880183/NCI-Microsites-Utility-Shades.pdf)
[style-elements-microsite-themes-hex.xlsx](https://github.com/NCIOCPL/cgov-digital-platform/files/2880185/style-elements-microsite-themes-hex.xlsx)

Example: 
https://www.cancer.gov/nci/rare-brain-spine-tumor

PR #1411, opened 2019-04-10
cgov_common Sub Themes and Theme Variants

alifr:
Adding ability to create cgov_common sub themes and color variants. Updating front-end build to support both sub themes and theme variants.

Adding Purple palette color variant #817 
Adding Blue palette color variant #818 
Adding sub theme STARTERKIT


alifr:
Color variant updates for Purple and Blue are incoming. New Green variant will probably be completed as a separate ticket.

alifr:
The styles for the Blue and Purple variants have a few issues since they do not map perfectly from PROD to Drupal. However, I think this ticket should be considered complete since it accomplishes the task of setting up the Purple and Blue microsite palettes. We should open another ticket to fine tune the styles for the Drupal environment.

Issue #1397, opened 2019-04-10
Equal Heights for Multimedia Row is not working correctly


bryanpizzillo:
An example page is http://www.devbox/about-cancer 
<img width="1142" alt="Screen Shot 2019-04-10 at 12 00 09 AM" src="https://user-images.githubusercontent.com/11215649/55850751-b2027800-5b23-11e9-9903-65a70778e30f.png">

PR #1412, opened 2019-04-10
Ticket 1397: equal heights display in multimedia row

arcepaul:
#1397 : equalheights was working correctly but feature card wasn't filling the extra column height. 
- added height 100% to feature card under equalheight column


arcepaul:
Adding @BrendanBeltzNIH  b/c he was enlightening me on the history behind this...

BrendanBeltzNIH:
I would suggest replacing the slot item wrapper div rather than changing the the stylesheet if possible.

arcepaul:
ODE example available at:
http://ncigovcdode86.prod.acquia-sites.com/about-cancer

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/content/home_landing/field--paragraph--field-mm-feature-card.html.twig
```diff
@@ -1,5 +1,7 @@
 <div id="nvcgSlLayoutFeatureB" class="equalheight small-4 columns card">
-  {% for item in items %}
-    {{- item.content -}}
-  {% endfor %}
+  <div class="slot-item only-SI">
```
bryanpizzillo: Can you move this into the for each? The idea is that `nvcgSlLayoutFeatureB` is the slot, and each content item in the slot gets the wrapper.

`slot-item` is given to all items in the container.
`first-SI` is given to the first element when there is more than one
`last-SI` is given to the first element when there is more than one
`only-SI` is given when it is the only element.

Maybe if someone looks at the CSS to find first/only/last and sees nothing then we can dispatch with that class.

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/src/entrypoints/homelanding/Homelanding.scss
```diff
@@ -523,10 +523,10 @@ parent accordion in all breakpoints. */
 }
 .multimedia-feature-card,
 .multimedia .feature-card {
-  @extend %autoheight-fix;
-  
+  //@extend %autoheight-fix;
   a {
```
bryanpizzillo: I was under the assumption based on @BrendanBeltzNIH comment that adding the class back in would not need a change to Homelanding.scss. Was that incorrect?

arcepaul: True, it does no harm but just ends up being bloat .  This is just rolling it back to what it was already.  I believe Brendan just wanted to add the wrapper div back to the template and the style is already handled with what's there now.

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/content/home_landing/field--paragraph--field-mm-feature-card.html.twig
arcepaul: Ok, I have moved the slot-item wrapper into the for-each for the feature card and the media item.  The only-SI doesn't do anything in the context of the multimedia so I remvoed it.  I found no reference to the other SIs in the multimedia styles

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/src/entrypoints/homelanding/Homelanding.scss
bryanpizzillo: Cool!

PR #1419, opened 2019-04-11
(#512) Add DTM script tag to site

### LINKED ISSUE #512: Add DTM script tag to site
## Description 
Add the CancerGov DTM `<script>` tag to all pages:

```<script src="//assets.adobedtm.com/f1bfa9f7170c81b1a9a9ecdcc6c5215ee0b03c84/satelliteLib-5b3dcf1f2676c378b518a1583ef5355acd83cd3d.js">```

### PR DESCRIPTION
**Pull request by daquinohd**
(Closes #512)

### CODE REVIEW

File: docroot/profiles/custom/cgov_site/config/install/adobe_dtm.settings.yml [hunk 1]
```diff
@@ -0,0 +1,5 @@
+enabled: true
+property_embed_code_id: f1bfa9f7170c81b1a9a9ecdcc6c5215ee0b03c84
```
bryanpizzillo: Hrrmm... so this config will apply to all sites, but I am sure we will want multiple. So how can we handle this in a multi site situations?

docroot/profiles/custom/cgov_site/modules/custom/cgov_core/src/CgovCoreTools.php
```diff
@@ -367,4 +367,25 @@ public function renameContentTypePermissions(string $content_type, array $permis
     return $permissions;
   }
 
+  /**
+   * Installs the DTM configuration based on environment variable.
+   */
+  public function installDtmEnvironment() {
+    if ($this->cloudEnvironment() != 'prod') {
+      $dtmSettings = $this->configFactory->getEditable('adobe_dtm.settings');
```
bryanpizzillo: Similar question as above, how can this scale To multiple sites?

docroot/profiles/custom/cgov_site/modules/custom/cgov_core/src/Plugin/Block/CgovDisqus.php
```diff
@@ -1,7 +1,8 @@
 <?php
 
-namespace Drupal\cgov_blog\Plugin\Block;
```
bryanpizzillo: This looks good, but I would like the pieces to be separate commits. So make a separate pr for the core function, then one for the others

bryanpizzillo: If the pr for the core is just one commit, then you can easily merge it into your work branches

docroot/profiles/custom/cgov_site/config/install/adobe_dtm.settings.yml
daquinohd: Changed the embed_code_hash to a development property. The embed_code_id is the same for all NCI sites.

Issue #598, opened 2019-01-30
Syndication Epic


amydugan:
Includes tickets needed to support syndication of Cancer.gov content. 
- [ ] Create Syndicate field storage #437
- [ ] Create Syndicate and Syndication Keywords Common Fields #203
- [ ] Technical Item: Create the Syndication Keywords field #399
- [ ] Attach Syndication Fields to Article Page #210
- [ ] Update Syndication Default on Article (to be unchecked now) #548

PR #1460, opened 2019-04-15
Add templates for HHS syndication theme

bkline:
Here are the syndication templates Bryan requested be folded into `develop`. Note that they won't do anything without the `hhs_syndication` field and the other bits to connect the field to content types or do the massaging to hook up with the theme. The closest ticket I can find to link this to is the syndication epic (#598). I haven't run the unit tests locally, since all I'm adding are decoupled templates, so if this fails in the pipeline, `develop` should fail as well (he says, ducking).


bkline:
This PR will be re-rolled, following up on discussion with BP.

PR #1467, opened 2019-04-15
Add and enable the taxonomy manager module.

### LINKED ISSUE #1413: Taxonomy Management of Site Sections does not work
When a taxonomy is too large, Drupal's management tools are problematic.
* Using the management screen to reorder sections throws errors. This is due to the fact that the UI tries to post back an object that has all nodes and weights. (>1mb)
* Adding a node to a section requires you to traverse a very tiny box with the entire tree.

We need a solution that requires the smallest amount of LOE, but allows the IAs to effectively manage the taxonomy without the server breaking or causing other issues.

adriancofie:
Closes #1413

* Adds and enables the Taxonomy Manager module ( e.g:  for Site Sections)

Technical Note: This required a patch in order for the config

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -51,6 +51,7 @@
         "drupal/pathauto": "^1.3",
         "drupal/seckit": "^1.1",
         "drupal/shield": "^1.2",
+        "drupal/taxonomy_manager": "^1.0@alpha",
         "drupal/token": "^1.5",
         "drupal/token_filter": "^1.1",
         "drupal/twig_field_value": "^1.2",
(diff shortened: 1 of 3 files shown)


adriancofie:
Build in progress to : http://ncigovcdode94.prod.acquia-sites.com/

bryanpizzillo:
@adriancofie Can you close this please since we are not going with this approach? I would but you may want that ODE for something else.

adriancofie:
Different approach.

PR #1515, opened 2019-04-20
[WIP] POC Features Setup

doylejd:
# Description
Sets up initial POC for features integrations with cgov_article and cgov_core. 
- Adds new CGov features bundle (prefixes all packages with cgov_)
- Renames modules to "Article" and "Core" so that module naming works with bundle.
- Configures excludes to remove features detected dependencies that didn't already exist in the modules (So they remain consistent with current setup). 
- Re-export features to get the updated config based on current state. (This breaks workflow, so stripped out dependencies based on how its currently used). 
- Created custom features alter for Cgov to remove workflow config that makes it dependent on entity bundles.

### TODO

- [x] Build filter for workflow config to remove dependencies based on how its currently being used.
- [ ] Figure out how to handle subset of permissions.

### DIFF
--- a/docroot/profiles/custom/cgov_site/config/install/features.bundle.cgov.yml
+++ b/docroot/profiles/custom/cgov_site/config/install/features.bundle.cgov.yml
@@ -1,4 +1,4 @@
-uuid: f8a36c79-a87a-4e3d-9f43-a477c4811aa1
+uuid: abeaadad-9beb-4af4-8e29-76a372e2a1ba
 langcode: en
 status: true
 dependencies: {  }
@@ -6,7 +6,7 @@ _core:
   default_config_hash: Eu8CvZYqHh4Zx02qu1Xf9a7QC3JnDQTOZmSWkD8h4ic
 name: 'Cgov Bundle'
 machine_name: cgov
-description: 'Cgov Profile Features'
+description: 'Auto-generated bundle from package Cgov Bundle'
 assignments:
   alter:
     core: true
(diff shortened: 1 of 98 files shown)


docroot/profiles/custom/cgov_site/modules/custom/cgov_core/config/install/user.role.advanced_editor.yml
```diff
@@ -5,12 +5,3 @@ id: advanced_editor
 label: 'Advanced Editor'
 weight: -6
 is_admin: null
-permissions:
```
doylejd: Since these are only partial permissions, removed from config and placed in hook_install().

docroot/profiles/custom/cgov_site/src/Plugin/FeaturesAssignment/CgovFeaturesAlter.php
```diff
@@ -0,0 +1,55 @@
+<?php
```
doylejd: Cgov_alter FeatureAssignment - Allows us to alter the configurations being exported from features to remove the dependencies we do not want. In this case, we remove the dependencies on entities so that workflow can be imported and than configured per-entity if/when that entity is enabled.

docroot/profiles/custom/cgov_site/config/install/features.bundle.cgov.yml
```diff
@@ -0,0 +1,104 @@
+uuid: abeaadad-9beb-4af4-8e29-76a372e2a1ba
```
doylejd: Custom bundle for cgov. Changes some default settings to allow core config to be exported/overridden by our modules. Enables new custom FeaturesAssignment for cgov. (See files below)

bryanpizzillo:
Closing this out as we did it outside of this PR.

PR #1524, opened 2019-04-22
(#1519) Feature cards are not displaying Lead image correctly

### LINKED ISSUE #1519: Feature cards are not displaying Lead image correctly
In order to fix an issue with Embedded Images some code was added to the various templates where the Lead image is displayed. This left extra space, and that was fixed in #1478. The fix had an unintended side affect of changing how the lead image appears in all other templates.

Example Screenshot:
Screen Shot 2019-04-22 at 12 05 52 PM

bryanpnih:
Closes #1519 

The solution ended up added a suggestion for the Entity's view mode and then making field--field-image-article--full.html.twig to target only the lead field for the full view mode of the node.


bryanpizzillo:
ODE: http://ncigovcdode105.prod.acquia-sites.com/AH_VIEW

PR #1561, opened 2019-04-25
(#1206) Add moderated media tab to content page

### LINKED ISSUE #1206: Media Needs Moderation Tab
Right now, we are unable to view media that is not published. The moderated tab is necessary for videos and infographics which have workflow.

bkline:
closes #1206

ODE: http://ncigovcdode110.prod.acquia-sites.com

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_image/cgov_image.install
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_image/cgov_image.install
@@ -1,0 +1,18 @@
+<?php
+
+/**
+ * @file
+ * Contains cgov_image.install.
+ */
+
+/**
+ * Implements hook_install().
+ *
+ * Attach our images types to the right workflow.
+ */
+function cgov_image_install() {
+  $tools = \Drupal::service('cgov_core.tools');
+  foreach (['cgov_image', 'cgov_contextual_image'] as $image_type) {
+    $tools->attachMediaTypeToWorkflow($image_type, 'image_workflow');
+  }
+}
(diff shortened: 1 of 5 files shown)


lburack:
@bkline - This is working as expected for infographic, file, and video. But images (cgov_image and cgov_contextual_image) are not being pulled into the Moderated Media tab. They are media so they should be there too.

bkline:
Support for the image types has been added. (The original ticket had said "The moderated tab is necessary for videos and infographics which have workflow.")

bkline:
> I don't like the fact that the image workflow just moved into media.

There was a `digital_platform_dev` Slack conversation about where things should go for this ticket. The image workflow was in the `cgov_image` module, but we agreed it wasn't appropriate for the view created for this ticket to live in that module, because it supports other document types, and that would make the view dependent on the `cgov_image` module. So since the view needs to know about both the workflows they had to be accessible at least at the level of `cgov_media`, where the view lives. I originally considered moving the image workflow all the way up to `cgov_core`, but after discussions with Blair I decided to move it no further away from the types which used it than necessary, so it ended up in `cgov_media`, where the view lives. I agree that if you repurpose the workflow so that it is for other types than images `cgov_core` would be the right home. But that notion never surfaced in the discussions.

bryanpizzillo:
> > I don't like the fact that the image workflow just moved into media.
> 
> There was a `digital_platform_dev` Slack conversation about where things should go for this ticket. The image workflow was in the `cgov_image` module, but we agreed it wasn't appropriate for the view created for this ticket to live in that module, because it supports other document types, and that would make the view dependent on the `cgov_image` module. So since the view needs to know about both the workflows they had to be accessible at least at the level of `cgov_media`, where the view lives. I originally considered moving the image workflow all the way up to `cgov_core`, but after discussions with Blair I decided to move it no further away from the types which used it than necessary, so it ended up in `cgov_media`, where the view lives. I agree that if you repurpose the workflow so that it is for other types than images `cgov_core` would be the right home. But that notion never surfaced in the discussions.

For the record, we have opened up #1593 to address the image workflow

Issue #1533, opened 2019-04-23
Caching Issue with Blog Post Pagination


lburack:
When you add a new blog post to a blog series, the new post you just created does not appear as a Newer Post > link at the bottom of the previous post. 

Also, if you delete blog posts, they are still appearing in the pagination links, but they go to a page not found.

daquinohd:
@lburack @VictoriaSunNIH - this is ready for testing on http://ncigovcdode109.prod.acquia-sites.com.

lburack:
I added a new post and it appeared in pagination links. Also, I deleted several posts and they are now no longer showing in the Older/Newer Post pagination links.

This passes IA Review. Thanks, @daquinohd!

PR #1570, opened 2019-04-25
(#1533) Caching Issue with Blog Post Pagination

daquinohd:
* Added page_cache_kill_switch service, dependency injection, and killCache() to BlogManager class
* Added cache kill switch to blog block plugins
* Added getCacheMaxAge( to blog block plugins
* Cleanup & comments


daquinohd:
Closing this PR - there is a better way to do this.

Issue #1462, opened 2019-04-15
Set up HSTS


blairlearn:
We currently have HTTPS set up on `www-dev-ac` and `www-int-ac`.  This should be set up to always be on, with HSTS to enforce it.

blairlearn:
**Requirement** Need to enforce HTTPS on environments where it's present (dev and int), but not on environments where it's unavailable (local dev, ODEs).

**Approach** Add a line to the `post-drupal-scaffold-cmd` event in `composer.json` to fire a custom blt command. Copy the logic from `toggleModules()` in [`ToggleModulesCommand.php`](https://github.com/acquia/blt/blob/10.x/src/Robo/Commands/Setup/ToggleModulesCommand.php) (BLT project) to detect the environment and only apply the patch to environments listed in `blt.yml`

```
hsts:
  dev:
    enable: true
  int:
    enable: true
```

blairlearn:
## Requirements
After chatting with @doylejd, we came up with revised requirements:
* Enforce HTTPS on non-ODE environments (dev, int, other named `cancer.gov` environments).
* For non-ODE environments (particularly stage/prod), return a `403 unauthorized` status for the `acquia-sites.com` and `acsitefactory.com` names.
* Do not enforce HTTPS on ODEs, CI, or local environments.
* Redirect to the name with hostname instead of a naked domain (i.e. https://www.cancer.gov instead of https://cancer.gov)

## Approach
* Install the `drupal/seckit` module.
* Patch the `.htaccess` file to be identical on all environments. This makes for more complicated rules, but not overly so.

## Starting point patches
**HSTS**
```
diff --git docroot/.htaccess docroot/.htaccess
index 55ba88ad..383dfb5a 100644
--- docroot/.htaccess
+++ docroot/.htaccess
@@ -198,4 +198,6 @@ AddEncoding gzip svgz
   Header always set X-Content-Type-Options nosniff
   # Disable Proxy header, since it's an attack vector.
   RequestHeader unset Proxy
+  # Set the HSTS header on Acquia sites where the non-Acquia domains are used.
+  Header set Strict-Transport-Security "max-age=31557600; preload"
 </IfModule>
```

**Redirect HTTP to HTTPS** (This one needs some tweaks)
```
diff --git a/.htaccess b/.htaccess
index b2a1a29..0247fb0 100644
--- a/.htaccess
+++ b/.htaccess
[...]

PR #1581, opened 2019-04-29
Configure to use NIH Authentication

### LINKED ISSUE #505: Configure to use NIH Authentication
This story is to setup the system to use NIH Login Authentication.

### LINKED ISSUE #1466: Make SAML active on installation
SAML authentication is currently not activated when deployed. This allows us to continue using existing logins until more tiers are ready and any kinks have been worked out.

To make it active by default:

1. Navigate to `docroot/profiles/custom/cgov_site/modules/custom/cgov_saml_auth_config/config/install`
1. Edit `cgov_saml_auth_config.settings.yml`
1. Set the `activate:` property to `true`.

The module is already toggled on for the `dev`, `int` and `test` environments, this step just "turns it on."

blairlearn:
Closes #505 
Closes #1466

### DIFF
--- a/blt/blt.yml
+++ b/blt/blt.yml
@@ -28,15 +28,18 @@ drush:
 modules:
   local:
     enable: [dblog, devel, kint, seckit, views_ui, cgov_yaml_content]
-    uninstall: [acquia_connector, shield, acsf]
+    uninstall: [acquia_connector, shield, acsf, simplesamlphp_auth, cgov_saml_auth_config]
   ci:
     enable: {  }
-    uninstall: [acquia_connector, shield]
+    uninstall: [acquia_connector, shield, simplesamlphp_auth, cgov_saml_auth_config]
   dev:
-    enable: [acquia_connector, shield]
+    enable: [acquia_connector, shield, simplesamlphp_auth, cgov_saml_auth_config]
+    uninstall: {  }
+  int:
+    enable: [acquia_connector, shield, simplesamlphp_auth, cgov_saml_auth_config]
     uninstall: {  }
   test:
-    enable: [acquia_connector, shield]
+    enable: [acquia_connector, shield, simplesamlphp_auth, cgov_saml_auth_config]
     uninstall: [devel, views_ui]
   prod:
     enable: [acquia_connector, shield]
@@ -72,5 +75,7 @@ command-hooks:
   post-deploy:
     dir: '${docroot}'
     command: null
+simplesamlphp: true
+## Cancer.gov Specific Settings
 cgov:
   drupal_users_file: '${repo.root}/cgov-drupal-users.yml'
--- a/composer.json
+++ b/composer.json
@@ -54,6 +54,7 @@
         "drupal/redirect": "1.x-dev",
         "drupal/seckit": "^1.1",
         "drupal/shield": "^1.2",
+        "drupal/simplesamlphp_auth": "3.0",
         "drupal/token": "^1.5",
         "drupal/token_filter": "^1.1",
         "drupal/twig_field_value": "^1.2",
@@ -67,8 +68,7 @@
         "webflo/drupal-finder": "^1.0.0",
         "webmozart/path-util": "^2.3"
     },
-    "require-dev": {
-    },
+    "require-dev": {},
     "conflict": {
         "drupal/drupal": "*"
     },
@@ -118,6 +118,9 @@
         "cgov-clean": [
             "CGovPlatform\\composer\\Hoover::clean",
             "@composer clearcache --ansi"
+        ],
+        "post-drupal-scaffold-cmd": [
+            "cd docroot && patch -p1 <../patches/htaccess-saml.patch"
         ]
     },
     "scripts-descriptions": {
@@ -187,14 +190,17 @@
             "acquia/blt": {
                 "NCIOCPL#48": "patches/blt/issue-48-fix-behat-chrome-invocation.patch"
             },
-            "drupal/pathauto":{
-                "3003373 : Wrong verbose messenger method" : "https://www.drupal.org/files/issues/2018-09-30/3003373-2.patch"
+            "drupal/pathauto": {
+                "3003373 : Wrong verbose messenger method": "https://www.drupal.org/files/issues/2018-09-30/3003373-2.patch"
             },
-            "drupal/address":{
+            "drupal/address": {
                 "2856797: Missing theme suggestions for address_plain": "https://www.drupal.org/files/issues/2018-04-10/address-missing_theme_suggestions_for_address_plain-2856797-10.patch"
             },
-            "drupal/token_filter":{
+            "drupal/token_filter": {
                 "3036541: Missing token_filter schema file": "https://www.drupal.org/files/issues/2019-03-04/3036541-2-token_filter_schema.patch"
+            },
+            "drupal/simplesamlphp_auth": {
+                "2907182: Admin UI Pages not Accessible via Permission": "https://www.drupal.org/files/issues/simplesamlphp_auth-admin_ui_pages-2907182-30.patch"
             }
         },
         "drupal-scaffold": {
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_saml_auth_config/cgov_saml_auth_config.install
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_saml_auth_config/cgov_saml_auth_config.install
@@ -1,0 +1,36 @@
+<?php
+
+/**
+ * @file
+ * Install file for cgov_saml_auth_config.
+ */
+
+/**
+ * Implements hook_install().
+ */
+function cgov_saml_auth_config_install() {
+
+  // Copy SAML settings from cgov_simplesamlphp_auth.settings.yml into
+  // the simplesamlphp_auth module.
+  //
+  // Get the custom config.
+  $customConfig = \Drupal::configFactory()->getEditable('cgov_saml_auth_config.settings');
+
+  // Get the default configuration.
+  $saml_config = \Drupal::configFactory()->getEditable('simplesamlphp_auth.settings');
+
+  // Replace default configuration with ours.
+  $saml_config->setData($customConfig->get())->save();
+  // Custom config object is no longer needed.
+  $customConfig->delete();
+
+  // Set up permissions for roles allowed to interact with SAML config.
+  $samlPermissions = [
+    'administer simplesamlphp authentication',
+    'change saml authentication setting',
+  ];
+  // Get our helper.
+  $siteHelper = \Drupal::service('cgov_core.tools');
+  $siteHelper->addRolePermissions(['site_admin' => $samlPermissions]);
+
+}
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_saml_auth_config/config/install/cgov_saml_auth_config.settings.yml
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_saml_auth_config/config/install/cgov_saml_auth_config.settings.yml
@@ -1,0 +1,37 @@
+langcode: en
+default_langcode: en
+activate: false
+mail_attr: https://federation.nih.gov/person/Mail
+unique_id: https://federation.nih.gov/person/SamAccountName
+user_name: https://federation.nih.gov/person/SamAccountName
+auth_source: cancergov-sp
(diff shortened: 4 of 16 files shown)


blairlearn:
Ready for review.
Login at https://www-int-ac.cancer.gov/saml_login

* You should be prompted to login via NIH login (your normal credentials).
* Drupal login should only work for the `admin` user.
* SAML login has been set up for sunvw, buracklb, and blausteincm.

claireblaustein:
@blairlearn I get a message - unrecognized user name and password. Am I supposed to be redirected to the AMS, or putting my user and pass in the regular screen?

blairlearn:
@claireblaustein - You may have caught it right as I was fixing a bug and the login credentials reset.

You should log in via https://www-int-ac.cancer.gov/saml_login.  The `saml_login` part is important. If you go to that URL, you'll be redirected to an NIH login server (`authdev.nih.gov`). On the NIH server, you login with your regular NIH credentials. From there, you'll be sent back to the test server.

VictoriaSunNIH:
@blairlearn - I just tested this, and while it logged me in, it gave me an error message (see below):

image

blt/blt.yml
```diff
@@ -72,5 +75,7 @@ command-hooks:
   post-deploy:
     dir: '${docroot}'
     command: null
+simplesamlphp: true
```
bryanpizzillo: I feel like this should go into separate blt.yml file for the environments.  You can do <ENV>.blt.yml and override settings on a per environment basis. 

ODEs configs are impossible, but as we discussed previously, ODEs will not get SSO.
One final note to remember, `ci` is the pipelines environment.

docroot/profiles/custom/cgov_site/config/install/user.role.administrator.yml
```diff
@@ -38,6 +38,7 @@ permissions:
   - 'administer pathauto'
   - 'administer permissions'
   - 'administer seckit'
+  - 'administer simplesamlphp authentication'
```
bryanpizzillo: I think this would be better done in the cgov_saml_auth_config.install.

docroot/profiles/custom/cgov_site/modules/custom/cgov_saml_auth_config/cgov_saml_auth_config.install
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * @file
+ * Install file for cgov_saml_auth_config.
+ */
+
+/**
+ * Implements hook_install().
+ */
+function cgov_saml_auth_config_install() {
```
bryanpizzillo: This smells like it should just be an actual config override. https://www.drupal.org/docs/8/api/configuration-api/configuration-override-system

If it is really just setting values, then please do not use the word override.

docroot/profiles/custom/cgov_site/modules/custom/cgov_saml_auth_config/config/install/cgov_saml_auth_config.settings.yml
```diff
@@ -0,0 +1,37 @@
...
+mail_attr: https://federation.nih.gov/person/Mail
+unique_id: https://federation.nih.gov/person/SamAccountName
+user_name: https://federation.nih.gov/person/SamAccountName
+auth_source: cancergov-sp
+login_link_display_name: 'NIH Login'
+header_no_cache: true
+role:
+  population: null
+  eval_every_time: false
+register_users: true
```
bryanpizzillo: I would like it if we have to add users on our own. With tens of thousands of NIH users, I would like to avoid people hitting our servers and automatically getting accounts.

hooks/common/post-code-deploy/post-code-deploy.sh
```diff
@@ -33,4 +33,5 @@ blt cgov:user:load-all -D cgov.drupal_users_file=$users_file
 blt cgov:locales:translate
 blt custom:install_cgov_yaml_content_by_module cgov_yaml_content
 
+
```
bryanpizzillo: Is this file change just a line of whitespace? If so, can you revert this please?

scripts/pipelines/build_artifact
```diff
@@ -7,6 +7,7 @@ export PATH=${COMPOSER_BIN}:${PATH}
 # Generate artifact in a separate directory.
 export NEW_BUILD_DIR=/tmp/artifact
 blt artifact:build -D deploy.dir=${NEW_BUILD_DIR} --ansi --verbose
+
```
bryanpizzillo: same as above, if this file difference is only 1 line of whitespace then can you revert it please?

simplesamlphp/config/acquia_config.php
```diff
@@ -0,0 +1,125 @@
...
+elseif (getenv('AH_SITE_ENVIRONMENT')) {
+  // Set  ACE ad ACSF sites based on hosting database and site name.
+  $config['certdir'] = "/mnt/files/{$_ENV['AH_SITE_GROUP']}.{$_ENV['AH_SITE_ENVIRONMENT']}/saml-certificate/";
+  $config['metadatadir'] = "/mnt/www/html/{$_ENV['AH_SITE_GROUP']}.{$_ENV['AH_SITE_ENVIRONMENT']}/simplesamlphp/metadata";
+  $config['baseurlpath'] = 'simplesaml/';
+  // Setup basic logging.
+  $config['logging.handler'] = 'file';
+  $config['loggingdir'] = dirname(getenv('ACQUIA_HOSTING_DRUPAL_LOG'));
+  $config['logging.logfile'] = 'simplesamlphp-' . date('Ymd') . '.log';
+  $creds_json = file_get_contents('/var/www/site-php/' . $_ENV['AH_SITE_GROUP'] . '.' . $_ENV['AH_SITE_ENVIRONMENT'] . '/creds.json');
```
bryanpizzillo: What's the plan for ACSF?  Is it a post task, or what? (Per the docs this needs to be sites.json or something when running under ACSF)

simplesamlphp/config/authsources.php
```diff
@@ -0,0 +1,339 @@
+<?php
+
+$config = [
+
+    // This is a authentication source which handles admin authentication.
+    'admin' => [
+        // The default is to use core:AdminPassword, but it can be replaced with
+        // any authentication source.
+
+        'core:AdminPassword',
```
bryanpizzillo: Is this overwritten somewhere?

simplesamlphp/config/authsources.php
```diff
@@ -0,0 +1,339 @@
...
+        // The URL to the discovery service.
+        // Can be NULL/unset, in which case a builtin discovery service will be used.
+        'discoURL' => null,
+
+        // The private key and certificate for signing the SAML XML data.
+        'privatekey' => 'saml.pem',
+        'certificate' => 'saml.crt',
+        'signature.algorithm' => 'http://www.w3.org/2001/04/xmldsig-more#rsa-sha256',
+
+        /*
```
bryanpizzillo: can you please remove the example cruft? We can get it from the blt code if needed.

simplesamlphp/metadata/saml20-idp-remote.php
```diff
@@ -0,0 +1,85 @@
+<?php
+/**
```
bryanpizzillo: Can we put this file in the secrets/not source controlled location? Then have a generic PHP script that pulls in the correct file?

blt/blt.yml
```diff
@@ -33,10 +33,13 @@ modules:
     enable: {  }
```
doylejd: @blairlearn  It wont let me comment on code that wasn't changed but we should probably add these modules to the "uninstall" section of "local:" and "ci:". At some point you will start syncing databases instead of doing fresh installs and having SimpleSAML installed will break things.

simplesamlphp/config/acquia_config.php
doylejd: Agreed - It likely makes sense to do V1 for ACE and build on it for the ACSF conversion in a new PR.

composer.json
```diff
@@ -187,14 +190,17 @@
...
             },
-            "drupal/token_filter":{
+            "drupal/token_filter": {
                 "3036541: Missing token_filter schema file": "https://www.drupal.org/files/issues/2019-03-04/3036541-2-token_filter_schema.patch"
+            },
+            "drupal/simplesamlphp_auth": {
+                "2907182: Admin UI Pages not Accessible via Permission": "https://www.drupal.org/files/issues/simplesamlphp_auth-admin_ui_pages-2907182-30.patch"
             }
         },
         "drupal-scaffold": {
```
doylejd: @blairlearn I see you have a patch for the .htaccess changes but its not in your composer file. Can we add it in here? This can go under the "scripts:" section. Example below (and recommend creating separate patches for each set of changes required):

```
"scripts": {
        "post-drupal-scaffold-cmd": [
             "cd docroot && patch -p1 --forward <../patches/htaccess-ssl.patch",
            "cd docroot && patch -p1 --forward <../patches/htaccess-saml.patch",
            "cd docroot && patch -p1 --forward <../patches/htaccess-hsts.patch"
        ],
    ...
```

hooks/common/post-code-deploy/post-code-deploy.sh
blairlearn: Yeah, that didn't belong there.

scripts/pipelines/build_artifact
blairlearn: Also same as above.

simplesamlphp/config/acquia_config.php
blairlearn: The plan, such as it is, is to sniff whether the code is running on ACE or ACSF and load the appropriate credentials file accordingly.

docroot/profiles/custom/cgov_site/modules/custom/cgov_saml_auth_config/cgov_saml_auth_config.install
blairlearn: As discussed, this is just a one-time setting of the configuration.  Comments and code have been modified to clarify.

composer.json
blairlearn: The SAML patch is already there.  There's nothing for SSL or HSTS (yet), but they **must** not be applied that way. Since the ODEs and local machines will likely never have SSL (not without a great deal of ongoing pain and suffering). Current plan is to apply them conditionally via BLT invoked from the same spot. (See #1462 for the lowdown.)

blt/blt.yml
blairlearn: Seems reasonable, and done!

simplesamlphp/config/authsources.php
blairlearn: Done!

blt/blt.yml
blairlearn: This is a flag which controls whether the `simplesamlphp` library is copied into the deploy artifact, it does not enable any modules. Leaving in here avoids some complexity by keeping all of the BLT configuration for SAML in once place.

blairlearn: This is a flag which controls whether the `simpleSAMLphp` configuration files are copied into the expected location.  It does not directly toggle the authentication module on/off.

Moving this out would mean BLT configuration for the authorization module needs to be done in two places. Perhaps *all* configuration for a given environment (e.g. the module enable/uninstall list) should be moved out?

simplesamlphp/config/authsources.php
blairlearn: That first bit is the IDP to use when logging in to <site>/simplesaml.

simplesamlphp/metadata/saml20-idp-remote.php
blairlearn: Done.

blairlearn:
@bryanpizzillo - DO NOT MERGE.  This is lacking the Derpal 8.7 update.  Just trying to get the IA approval ahead of time..

@VictoriaSunNIH, @claireblaustein, and @lburack:

Let's try this again, eh?

Login at https://www-int-ac.cancer.gov/saml_login

* You should be prompted to login via NIH login (your normal credentials).
* Drupal login should only work for the admin user.
* SAML login has been set up for sunvw, buracklb, and blausteincm with the standard editor/author credentials.

VictoriaSunNIH:
@claireblaustein - I didn't have the same issues you encountered.  Once I logged in via SAML, everything worked for me.

Issue #1590, opened 2019-04-30
Blog Series translations breaking


daquinohd:
Blog Series translations are breaking when running the migration script.

daquinohd:
@lburack @VictoriaSunNIH - this was more of a migration/yaml content issue, but can you verify that the Blog Series Banner Image field translates as expected? The field gives the user the ability to translate the image, alt text, both, or neither.

PR #1619, opened 2019-05-03
(#1590) Blog Series translations breaking

daquinohd:
* Fixed translation_sync values in field.field.node.cgov_blog_series.field_banner_image.yml 
* Made 'Temas y relatos blog' a translation item


daquinohd:
> should `alt:` also be `alt: alt` in addition to file? or put another way, can you correctly translate the alt text & image of the banner?
Right now, the original image and alt text are copied over to the translation and the user has the option to keep/change/delete any combination of the image and alt text.
We don't need to add the `alt: alt` sync option - it actually breaks our build.

daquinohd:
#1590 is being committed to the *feature/blog-tirs* branch. Closing this PR.

PR #1633, opened 2019-05-06
(#1504) Make a homepage yml and make it work

### LINKED ISSUE #1504: Make a homepage in YAML and make sure we can make a homepage
Out of the box, drupal has / as the login page.

We need to make sure we can have a cancer.gov homepage that doesnt go to the log in page and has the https://www.cancer.gov/, and then make a homepage using YAML content.

bryanpnih:
## Testing Notes
* A "homepage" should appear at when you go to http://ncigovcdode123.prod.acquia-sites.com/
* The http://ncigovcdode123.prod.acquia-sites.com/user/login url should still work
* When you edit the English home site section and change the landing page, http://ncigovcdode123.prod.acquia-sites.com/ should point to the new page.
   * NOTE: You *should* change the existing landing page's SiteSection/Pretty to reflect it is not the homepage first. This update will not change the Landing Page's site section, nor pretty url. All it does is makes this page the system's front page, and makes sure the URLs pointing to that node display as /.
* Aliases for the content has a site section of "Home" and an empty pretty url will still be generated as `/-0`. This should not show up on the front-end. Places to test:
   * Link It
   * Lists
   * Cards
* The URLs in the head metadata should not show `/-0` anywhere either.

Closes #1504

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/cgov_site_section.module
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/cgov_site_section.module
@@ -109,6 +109,16 @@ function cgov_site_section_taxonomy_term_update(EntityInterface $entity) {
   if ($entity->bundle() !== 'cgov_site_sections') {
     return;
   }
+
+  // We only set the front page for English nodes without a parent.
+  // NOTE: this does not account for the case when the root becomes a
+  // child of another node. It is possible, but the amount of complexity
+  // that check would add is not worth it.
+  // NOTE2: target_id can be an int and it can be a string.
+  if ($entity->parent->target_id == 0 && $entity->langcode->value === 'en') {
+    _cgov_site_section_attempt_set_front_page($entity);
+  }
+
   // TODO: This comparison should be done somewhere else and less messy.
   // Do not do unnessesary updates, so only cascade if a change was
   // made to a significant field.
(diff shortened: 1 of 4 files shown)


VictoriaSunNIH:
@bryanpizzillo , @lburack - I'm testing this now, but i still get directed to the log in page...am I doing something wrong?

Putting in http://ncigovcdode122.prod.acquia-sites.com/ redirects me to http://ncigovcdode122.prod.acquia-sites.com/user/1 if I'm logged in, and if I'm not logged in, I get the regular log in page

lburack:
Same issue is happening for me, @VictoriaSunNIH @bryanpnih. I tried navigating to the page from the Content Overview — I'm able to view the Spanish homepage (http://ncigovcdode122.prod.acquia-sites.com/espanol), but when I click the toggle, it takes me to the login (or /user/1) page.

bryanpizzillo:
So in hindsight, a user should NEVER use linkit to link to the Home Page node. They should always use '/' or '/<front>'. If we ever change the node ID of the home page, Drupal would generate that URL as whatever the new pretty url is for the old homepage.

bryanpizzillo:
Also I am checking something with how the home page gets created. It looks like the YAML content loader may save a landing page using a slightly different method than the back end. I was able to change the homepage though on the site, it is just the initial load of content that did not work.

bryanpizzillo:
I am going to recreate the ODE

VictoriaSunNIH:
@bryanpizzillo - So when i added the homepage via linkit, I got /node/136.  Is that right?

bryanpizzillo:
I loaded it up again and it set the home page correctly. So it seems intermittent.

Issue #1197, opened 2019-03-25
Align Percussion content type specific styles to Drupal Content Types Styles


amydugan:
Inventory was completed in #1339.

alifr:
Relates to #473. See spreadsheet posted there by Mark for complete list of page specific styles in CSS.

alifr:
This ticket can not be finished until PDQ, Blogs and CTHP Content Types are added to develop.

VictoriaSunNIH:
@alifr - What does this affect on the front end?  Please list out what is being targeted so that we can do the IA review properly.

PDQ, Blogs and CTHP are all on develop now.

amydugan:
Stories to be broken out by 4/30.

alifr:
Tasks for this ticket:
- There were variations between home page and landing page accordions in `accordion.scss`, but now they share the same template so these styles can be consolidated and scoped to homelanding entry point scss
- scope margin on `.ncilandingpage p, .ncihome p:not(.site-name):not(.site-link)` elements in `homelanding.scss` to the homelanding entry point scss
- scope margin on `.ncilandingpage h1` in `homelanding.scss` to homelanding entry point scss

- ensure layout styles seen on `#nvcgSlSectionNav` in `blogs.scss` is now handled by twig templates.
- scope blogs post specific padding on `.contentzone` in `blog.scss` to blogPost entry point scss
- medium sized images are smaller on blogs. Adjust `figure.image-right-medium` in `blogs.scss` to be properly scoped to the blogPost entry point scss file

alifr:
Acceptance Criteria:

1. ensure margins around accordions at mobile breakpoints on the home and landing pages are correct and flush to the edge of the browser without clipping the +/- expanding icon
2. ensure margins on `<p>` elements in home and landing pages are just `.5em` on the top and bottom
3. ensure the top margin on `<h1>` elements on home and landing pages remains `.58em` even on mobile breakpoints
4. ensure layout looks correct on BlogPosts
  1. ensure that `<div id="nvcgSlSectionNav">` does not exist in the markup
  2. ensure that if `<div id="nvcgSlListBlogRTRail">` is removed from the page `<div id="main">` spans the full width of the page
  3. ensure that main content (`<div id="main">`) has a left padding of `1.6em`
5.

alifr:
Tasks have been broken out into the following tickets:

Home/Landing: https://github.com/NCIOCPL/cgov-digital-platform/issues/1608
[...]

PR #1656, opened 2019-05-07
Ticket/1197 content type specific styles

alifr:
With regards to ticket #1197


alifr:
So I merged the three branches into one and got six commits on the PR?

BrendanBeltzNIH:
One of those branches should have been squashed into the one issue one commit format. You'll need to do an interactive rebase now.

PR #1665, opened 2019-05-07
File Enhancements (Direct Downloads and All File Extensions)

### LINKED ISSUE #1166: Alias for files should take you directly to the download
When clicking on the file alias in drupal, users should be taken directly to the download.

The view for it /media/ID should also take users to the download

### LINKED ISSUE #1521: Allow all file types to be uploaded to File content type
Find a way to whitelist all file content types for the File content type.

bryanpnih:
This will be on http://ncigovcdode125.prod.acquia-sites.com/ whenever the build finishes and I deploy it there.

Closes #1166 
Closes #1521

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_file/cgov_file.module
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_file/cgov_file.module
@@ -1,0 +1,39 @@
+<?php
+
+/**
+ * @file
+ * Contains the cgov_file functions.
+ */
+
+use Drupal\file\FileInterface;
+
+/**
+ * Implements hook_form_alter().
+ */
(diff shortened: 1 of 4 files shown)


lburack:
@bryanpizzillo @VictoriaSunNIH – This seems to be passing.

I tested the following file types:
- PDF
- Word doc (.docx)
- Excel (.xlsx)
- Text file (.rtf)
- .zip (from CCG's TCGA content)

I was able to link all of these in the wysiwyg (with automatic download/open PDF). Excluding the zip files, I was able to add these to related resources and to a list (with automatic download/open PDF). FWIW, the rich text file doesn't have an icon, so it looks a little different. This is not an issue, just putting this here for reference.

Screen Shot 2019-05-08 at 5 19 43 PM

PR #1685, opened 2019-05-09
(#1668, #1657) Makes Image media/{id} go directly to raw file and fix mobile nav language flag

### LINKED ISSUE #1657: Main/Mobile Menu is in English when in Spanish Interface, it Should be Spanish
When the user interface is in English, but the menu cannot find a URL, the Main/Mobile Menu displays in English. You can see this by going to /espanol/user/login.

The Spanish menu should always show when the user interface language is Spanish.

### LINKED ISSUE #1668: Alias for images should take you directly to the raw image file
The route /media/{media} for cgov_image and cgov_contextual_image should return a 404. We don't want junky URLs laying around.

This is what would have happened if we left the setting for "Enable View Media URLs" turned off. But we need URLs for Video & Infographic pages, as well as aliases for download files.

HOWEVER, the Media Overview view and the Media Moderation view generate URLs to the view controller for the media items. It would be silly to generate a link for editors that would knowingly return a 404, so what to do?

Our best url to pick? The path to the file in the field_media_image field. (e.g. the image)

Implementation:
* A EventSubscriber that returns a 404 if the request is for /media/{media} and the bundle is cgov_image or cgov_contextual_image.
  * This would be similar to 
https://github.com/NCIOCPL/cgov-digital-platform/blob/develop/docroot/profiles/custom/cgov_site/modules/custom/cgov_core/src/EventSubscriber/TranslationAvailableSubscriber.php, except simpler in that it would:
    * only handle the route for `entity.media.canonical`
    * Only need to check the bundle to determine if it needs to throw.
* An [OutboundPathProcessor](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21PathProcessor%21OutboundPathProcessorInterface.php/interface/uses/OutboundPathProcessorInterface/8.7.x) that would replace the /media/123 url with the path to the image file. This would need to be higher in the priority than alias.

This should handle the views listing AND give the 404 on request.

BrendanBeltzNIH:
Closes #1668, closes #1657.

ODE: http://ncigovcdode133.prod.acquia-sites.com

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.services.yml
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.services.yml
@@ -13,7 +13,7 @@ services:
       - '@entity_type.manager'
   cgov_core.cgov_navigation_manager:
     class: Drupal\cgov_core\Services\CgovNavigationManager
-    arguments: ['@path.current', '@path.alias_manager', '@entity_type.manager', '@entity_field.manager', '@logger.channel.cgov_core']
+    arguments: ['@path.current', '@path.alias_manager', '@entity_type.manager', '@entity_field.manager', '@logger.channel.cgov_core', '@language_manager']
   cgov_core.twig_extensions:
     class: Drupal\cgov_core\CgovCoreTwigExtensions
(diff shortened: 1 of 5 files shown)


lburack:
good to go!

PR #1754, opened 2019-05-17
(NCIOCPL#1466) Make SAML active on installation

### LINKED ISSUE #1466: Make SAML active on installation

blairlearn:
When deploying to a (non-ODE) environoment, make SSO mandatory without
a manual activation step.

Closes #1466


blairlearn:
Moved into #1773.

PR #1759, opened 2019-05-18
(#792) Generate site maps

### LINKED ISSUE #792: Implement automatic generation of sitemap xml
When we implement the automatic generation of sitemap XML, this should generate sitemaps for:
* Cancer.gov
* Each microsite under cancer.gov (nano, nciconnect, mypart, dceg)

bkline:
Closes #792 

- adds the `simple_sitemap` module
- overrides the default configuration for that module
- configures which content types should be mapped with what settings
- implements a controller which provides a site index XML file for www.cancer.gov
- adds a route subscriber which serves up the index at `/sitemap.xml` if the site is `cgov`
- implements a configuration interface to manage the site ID and production URL

We don't need the production URL for this task, but we will need it for HHS syndication, and it fits well here. I won't be shocked or surprised if you decide you want a different approach to capturing the information about which site we're on, but this works, and I figured it would be easier (for me, at least) to have the conversation about what we really want to do if we actually have something to look at as a starting point. Or, you might decide we'll use this as is for the initial launch, and swap in something you like better after the dust settles. The only thing in the whole system that depends on this config information is the new `cgov_sitemap` module (and all it needs to know is the answer to the question "are we on the cgov web site?").

Let's first see if the tests pass in the cloud. They do on my local machine.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -55,6 +55,7 @@
         "drupal/seckit": "^1.1",
         "drupal/shield": "^1.2",
         "drupal/simplesamlphp_auth": "3.0",
+        "drupal/simple_sitemap": "^3.1",
         "drupal/token": "^1.5",
         "drupal/token_filter": "^1.1",
         "drupal/twig_field_value": "^1.2",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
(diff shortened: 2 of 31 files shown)


bkline:
Loaded onto ODE 149. The cron job has run, generating the site map. I set the `cgov_core.site_config` values and cleared the cache (the changes don't take effect until the cache is cleared.

* http://ncigovcdode149.prod.acquia-sites.com/sitemap.xml
* http://ncigovcdode149.prod.acquia-sites.com/content/sitemap.xml

bkline:
Replaced by PR #2049

Issue #1781, opened 2019-05-21
Blog categories are showing an empty URL for topic when migrated in


VictoriaSunNIH:
On the migration ODE, blog categories are showing an empty URL for topic.

URL: https://www-test-ac.cancer.gov/news-events/cancer-currents-blog?topic=

topic seems to be empty for each category.

daquinohd:
May not be related, but some categories are no longer showing Blog Posts that are tagged to that topic.

daquinohd:
Category filter issue happens between 
`6c500e4 Load site sections from migration data` (topics not appearing) 
`cedcdd8 Feature/core prod check` (latest 'good' build where topics appear)

PR #1787, opened 2019-05-21
(#1781) Blog categories show empty URL for topic when migrated

daquinohd:
* Set taxonomy/tid URLs as default before setting the pretty URL
* Template cleanup


daquinohd:
(Closes #1781)

daquinohd:
This will be part of a blog migration feature branch

Issue #1507, opened 2019-04-17
Bug: Public Use text should reference the content title on Blog Post page


popkhadzeg:
## Description 
Public Use text should reference the content title on Blog Post page, but it says "Node view" for English and Spanish pages.
Fine on other content type pages where public use text appears.

example: http://www-dev-ac.cancer.gov/news-events/cancer-currents-blog/2019/vitamin-d-supplement-cancer-prevention

<img width="1260" alt="Screen Shot 2019-04-17 at 3 09 47 PM (2)" src="https://user-images.githubusercontent.com/46539699/56314459-27bda380-6123-11e9-8a13-4ef5e651606e.png">

alifr:
I've noticed that "Node view" is seen in the breadcrumbs of the Blog Post editing page, but in an article page it's using the article title. I think the disconnect is here somewhere.
<img width="483" alt="Screen Shot 2019-04-23 at 10 33 19 AM" src="https://user-images.githubusercontent.com/5579274/56589634-a26b4080-65b3-11e9-9f0a-a70f31898061.png">
<img width="481" alt="Screen Shot 2019-04-23 at 10 32 57 AM" src="https://user-images.githubusercontent.com/5579274/56589635-a26b4080-65b3-11e9-807d-e44332386241.png">

PR #1803, opened 2019-05-22
(#1507) Public Use text should reference the content title on Blog Post page

daquinohd:
* Added patch for page_manager title issue


lburack:
Working on the English and Spanish blog posts I created. Passes IA Review!

PR #1821, opened 2019-05-27
NCI-000: Updating configuration of features to not exclude curated co…

doylejd:
…nfig.


bryanpizzillo:
Manually pulled into feature/features and squashed into the initial features config commit.

PR #1881, opened 2019-06-04
NCI-000: Adding new site inc to view diff.

doylejd:
### Changes
- Adds pre-sites-php hook provided by Acquia to allow for multi level domains to be setup.


doylejd:
@bryanpizzillo Since you mentioned we are getting rid of nested structures here - I will close this PR.

Issue #1067, opened 2019-03-15
Output Syndicate message when Syndication is selected


VictoriaSunNIH:
On Articles and PDQ CIS, when syndicate is selected, we need to output the following message and icon at the bottom of the page (below public reuse)

![image](https://user-images.githubusercontent.com/31222859/54437875-f6a91800-470b-11e9-8085-75ddc1f6d075.png)

Spanish version:
![image](https://user-images.githubusercontent.com/31222859/54437904-07f22480-470c-11e9-9aff-ef2f8873bd31.png)

Example:
https://www.cancer.gov/types/head-and-neck/head-neck-fact-sheet
https://www.cancer.gov/espanol/tipos/cabeza-cuello/hoja-informativa-cabeza-cuello

PR #1883, opened 2019-06-04
(#1067) favicon fingerprint

alifr:


alifr:
also removed the manifest file

PR #1923, opened 2019-06-07
Patch for XML migrations.

adriancofie:
https://www.drupal.org/project/drupal/issues/3048464


adriancofie:
Merged changes into another PR.

PR #1937, opened 2019-06-08
(#1936) Create an Embeddable Image Mode for Inline Images

### LINKED ISSUE #1936: Create an Embeddable Image Mode for old Percussion Inline Images
Percussion inline links do not have the figure wrapper around them, no do they have any sort of size constraint.
Example Markup: 
```
<img src="<url>" alt="<field_media_image.alt>" />
```

This will require a new View Mode
Label: Image Display Inline

This will require a template/display view for that view mode:
The media_image_field should use the Article Image Style.
(It should just be able to be {{content}}

Enable in Image Embed...

bryanpnih:
- Updated Entity Embed to RC1
- Add Entity Embed suggestions to allow for wrapper-less inline images.
- Changed migration rule for inline images.

Closes #1936 
ODE:  ncigovcdode161.prod.acquia-sites.com


bryanpizzillo:
Upgrading to Entity Embed RC1 broke linking, so reverted back to older version.

bryanpizzillo:
Moved this commit to #1939. Closing.

PR #1939, opened 2019-06-08
Inline images and Internal External Links

### LINKED ISSUE #1936: Create an Embeddable Image Mode for old Percussion Inline Images

### LINKED ISSUE #1938: Allow external links to have internal links.
In the migration we have a lot of external links with anchors and query parameters. The content migrates in, but the pages blow up. We need to allow external links to contain internal links. (Without the stupid autocomple)

This applies to cgov_external_link, cgov_card_external and cgov_cthp_feature_card_external.

bryanpnih:
Create an Embeddable Image Mode for Inline Images (Closes #1936)
Allow external links to have internal links. (Closes #1938)

ODE: ncigovcdode161.prod.acquia-sites.com

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -213,9 +213,6 @@
             },
             "drupal/simplesamlphp_auth": {
                 "2907182: Admin UI Pages not Accessible via Permission": "https://www.drupal.org/files/issues/simplesamlphp_auth-admin_ui_pages-2907182-30.patch"
-            },
-            "drupal/address": {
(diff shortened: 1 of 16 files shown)


bryanpizzillo:
NOTE: Upgrading to Entity Embed RC1 broke linking, so reverted back to older version. (Tried it as part of inline images)

Issue #1630, opened 2019-05-03
Remove drupal form login for administrator role.


blairlearn:
Leaving the `administrator` role and `admin` user able to login using the Drupal authentication system until we're sure we've got the kinks worked out SSO.

In `docroot/profiles/custom/cgov_site/modules/custom/cgov_saml_auth_config/config/install/cgov_saml_auth_config.settings.yml`

* Set `allow.default_login` to `false`
* Set `allow.default_login_roles.administrator` to `'0'`
* Set `allow.default_login_users` to `'0'`

blairlearn:
A minor addition to the approach: Create a `cgov_saml_auth_config_dev` module which `blt drupal:toggle:modules` will activate on the dev tier in place of `cgov_saml_auth_config`.

The new `_dev` module will allow users in the `administrator` role to use Drupal login.

bryanpizzillo:
I have also set the admin user as blocked when a site is installed in prod.

VictoriaSunNIH:
@blairlearn - this still needs to be available on lower tiers, just not on prod.  Needs to be implemented as a staging factory hook to reenable admin on a dev tier

PR #2041, opened 2019-06-18
ticket/1986 siteadmin perms

blairl-nih:


blairlearn:
Replaced by 2024.

#1630 to be handled atomically.

PR #2049, opened 2019-06-19
(#792) Generate site maps

### LINKED ISSUE #792: Implement automatic generation of sitemap xml

bkline:
Closes #792 

* add the `simple_sitemap` module
* override the default configuration for that module
* configure which content types should be mapped with which settings
* register the configuration changes with the `features` module

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -59,6 +59,7 @@
         "drupal/role_delegation": "^1.0",
         "drupal/seckit": "^1.1",
         "drupal/shield": "^1.2",
+        "drupal/simple_sitemap": "^3.2",
         "drupal/simplesamlphp_auth": "3.0",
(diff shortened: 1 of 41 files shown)


bryanpnih:
@bkline I deployed this to http://ncigovcdode196.prod.acquia-sites.com.  I tried visiting http://ncigovcdode196.prod.acquia-sites.com/sitemap.xml and it is giving a page not found error. What is the URL for the sitemap?

bkline:
That's the right URL, but the sitemap doesn't get created until `cron` has run.

PR #2073, opened 2019-06-22
(#2074) Blog Post pagination issue

daquinohd:
* Added sort utility function for Blog Post pager link issues


daquinohd:
### ODE: http://ncigovcdode202.prod.acquia-sites.com

lburack:
It seems like this is working now. Not all the Spanish blog posts were available, but the first few that I checked had the correct titles displaying for "< Artículo anterior" and "Artículo siguiente >"

Issue #2186, opened 2019-07-10
Print Style Issues - Images Printing Too Large (backlog issue to investigate)


lburack:
During the content freeze, we received feedback from content owners that the print styles of the new site are different than what has been on the live site.

See screenshots received from content owners below:
<img width="836" alt="Screen Shot 2019-07-10 at 11 52 31 AM" src="https://user-images.githubusercontent.com/29731392/60984181-42aa4900-a309-11e9-8f70-6124a886a869.png">

<img width="860" alt="Screen Shot 2019-07-10 at 11 52 37 AM" src="https://user-images.githubusercontent.com/29731392/60984190-450ca300-a309-11e9-9ba8-eb5052f3fbea.png">

amydugan:
Closing - update to print stylesheet to include embedded entity sizing rules went live 7/18/19 (v 1.1.0)

PR #2207, opened 2019-07-15
(#2186) embedded entity sizing rules added to print stylesheet

arcepaul:
added embedded entity sizing rules that were previously hidden because of media query


arcepaul:
@VictoriaSunNIH @lburack 

ODE: http://ncigovcdode231.prod.acquia-sites.com/

bryanpizzillo:
You branch is behind whatever you branched from. I rebased from develop, where it appears you branched from. Since its commit is the same as develop, this should not have an issue being pulled into the release branch. Obviously, you cannot just pull your branch anymore.

Issue #2256, opened 2019-07-25
The content_previewer role needs the "View latest version" permission.


blairlearn:
The `View latest version` permission is not set for the `content_previewer` role.  This results in users with this role getting an access forbidden message when they attempt to preview a page, e.g. https://www-cms-dev.cancer.gov/node/1026837/latest

amydugan:
Went live 7/30 with Hotfix 1.1.2. Closing

PR #2270, opened 2019-07-26
(#2256) Add 'view latest version' permission to content_previewer role.

blairl-nih:
Add previewer (content_previewer) to set of users for local testing.

### DIFF
--- a/cgov-drupal-users.yml
+++ b/cgov-drupal-users.yml
@@ -36,3 +36,7 @@ additional_users:
       - 'image_manager'
       - 'admin_ui'
       - 'blog_manager'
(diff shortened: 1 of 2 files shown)


blairlearn:
# How to test:

1. Login as user `preview`.
2. Browse to: http://ncigovcdode240.prod.acquia-sites.com/node/821/latest
3. Expect to see the body text "Newer version"

As a not-logged-in-user (incognito/private browsing window)
1. Browse to http://ncigovcdode240.prod.acquia-sites.com/test/preview
2. Expect to see the body text "Public revision."

PR #2278, opened 2019-07-31
Ticket/2249 stuck pdq drafts

### LINKED ISSUE #2249: Some PDQ content is stuck in the draft state
The Drupal entity API returns the wrong revision number for the latest revision of an English summary when it is queued together with its Spanish translation for transition to the `published` state. As a result, the English summary in such as case is left in the `draft` state.

bkline:
* Replace the use of the flawed entity query API with a direct call to get the latest revision
* Explicitly set the `revision_translation_effective` flag when publishing
* Add missing `@throws` statements
* Replace `t(...)` with `$this->t(...)`
* Closes #2249


blairlearn:
Replaced by #2280 to resolve git issues.

Issue #2291, opened 2019-08-02
Correct entity ID for Test Tier IDP


blairlearn:
In `simplesamlphp/config/authsources.php`, the code for deciding which IDP to use includes this block.

```
  case '01test':
  case '01testup':
    // This may not be correct.
    $idp = 'https://authtest.nih.gov/SAML2/IDP';
    break;
```

The correct value is `https://authtest.nih.gov/IDP`.

PR #2292, opened 2019-08-02
(#2291) Correct entity ID for Test Tier IDP

blairl-nih:
- Correct the IDP entity ID for the ACSF test tier.
- Make the selection of IDP for ACSF dev tier explicit.


blairlearn:
Tested on www-cms-test with `build-20200330_1221`.  Still working through IDP issues for the other cms sites on that tier, but it works.

PR #2325, opened 2019-08-10
PDQ Restful API patches

### LINKED ISSUE #2318: Multiple simultaneous publish preview requests break publish preview for PDQ
If two requests for publish preview of the same PDQ summary hit the same site at the same time, it is likely that the calls to delete the temporary nodes will fail with an error message complaining that the request is ambiguous (because the node ID lookup for the negative CDR ID used for the PP request finds more than one node). This can happen if the user gets impatient and clicks the publish preview link a repeatedly. Once that condition occurs, subsequent attempts to preview the same summary will fail, because none of the temporary documents get deleted and they start piling up.

I’ll have to give some thought to how to prevent this from happening. I could just modify the code to not worry about ambiguous delete requests if the CDR ID is negative (which it is for the PP requests, to distinguish from the permanent summary content nodes). The downside for that would be that if two users request PP for the same summary at exactly the same time, one of them (likely the second user) will get an error. Very unlikely to happen, though, and a much less problematic result than a summary getting stuck so that all future attempts to run publish preview on it will fail until the temporary nodes are cleared out.

### LINKED ISSUE #2319: Watchdog entry is created when a PDQ POST request comes in with a missing nid index
The primary CDR publishing code for PDQ content ensures that the `$summary` array passed to the end point has a `nid` index, even if the value is `NULL` (as it will be for a new English summary). However, the publish preview code does not behave the same way, the that index is missing. A Jira ticket will be filed to fix that problem on the CDR side, but the Drupal PHP code for the CIS and DIS POST end points should assign the `$nid` variable defensively:

`$nid = $summary['nid'] ?? '';`

bkline:
* Assign `$nid` defensively in case the 'nid' index is not present in `$summary`
* Allow simultaneous deletion of multiple temporary publish preview nodes
* Closes #2318 
* Closes #2319

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/pdq_cancer_information_summary/src/Plugin/rest/resource/PDQResource.php
+++ b/docroot/profiles/custom/cgov_site/modules/custom/pdq_cancer_information_summary/src/Plugin/rest/resource/PDQResource.php
@@ -195,7 +195,7 @@ class PDQResource extends ResourceBase {
   public function post(array $summary) {
 
     // Extract the bits we'll use frequently.
-    $nid = $summary['nid'];
+    $nid = $summary['nid'] ?? '';
     $language = $summary['language'];
     $cdr_id = $summary['cdr_id'];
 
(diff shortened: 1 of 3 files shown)


bkline:
Copied from Slack:

> As for the PR, I’m not sure what an IA would do to test on the ODE, beyond what Volker and I can do (and perhaps the users) to hammer away at publish preview on DEV. I’m mostly relying on my own push test to confirm that publishing is unaffected and testing of publishing preview by CDR devs and users, as well as the inspection of the patch itself by Blair and/or Bryan. Your thoughts, @blairlearn?

Issue #2180, opened 2019-07-09
Blog Category - Cache Issues


lburack:
A Blog Category (Taxonomy Term) was updated, but the change did not apply to the Categories block on all the pages within the blog. (This was Cancer Currents)

I'm creating this issue because I:
1) had a tab open on https://www-prod-acsf.cancer.gov/news-events/cancer-currents-blog?topic=fda-approvals. The page title was "FDA Drug Approvals - Cancer Currents Blog" and the right rail category was "FDA Drug Approvals"
2) In another tab, I had www-cms.cancer.gov open and I changed the blog taxonomy term to remove 'Drug' so the category was "FDA Approvals"
3) On prod-ascf, I see my update when I'm on any of the other category pages (e.g., https://www-prod-acsf.cancer.gov/news-events/cancer-currents-blog?topic=disparities), but **not** when I'm on the fda-approvals page itself.

If this doesn't make sense, @bryanpizzillo can better explain/update this.

bryanpizzillo:
A change to a the `name` of a Blog Category should trigger a cache invalidation of all the pages displaying the Categories block. Since the Categories block is dynamic and shared across all blogs, you need to be careful about what gets invalidated. (If you invalidated just the block ID, you would potentially clear out all blogs from the cache)

Since all pages within a blog show the block, it would be simplest to invalidate the blog series if a category is updated. (Just make sure the blog series node id is a cache tag on every page)

kalidindir:
@lburack @VictoriaSunNIH
http://ncigovcdode256.prod.acquia-sites.com

VictoriaSunNIH:
@lburack - I tested this by changing the biology category.  I see that the right nav changed on all pages, as well as the page title.  I also created a new post and tagged it and saw the correct name there.  Waiting on you to retest!

lburack:
@VictoriaSunNIH @kalidindir — this also passes my IA review. I'll test the other tickets related to the PR #2346 in the morning.

PR #2336, opened 2019-08-16
(#2180) Add cache tag node:blog_series_nid for blog node page. Invali…

kalidindir:
…date blog series tag node:blog_series_nid when blog category taxonomy term is updated.

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_blog/cgov_blog.module
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_blog/cgov_blog.module
@@ -8,6 +8,7 @@
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\taxonomy\Entity\Term;
 use Drupal\views\ViewExecutable;
+use Drupal\Core\Cache\Cache;
 
 /**
  * Implements hook_form_alter().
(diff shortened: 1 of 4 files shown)


docroot/profiles/custom/cgov_site/modules/custom/cgov_blog/cgov_blog.module
```diff
@@ -218,3 +219,32 @@ function cgov_blog_preprocess_block(&$variables) {
   }
 
 }
+
+/**
+ * Implements hook_preprocess_node().
+ */
+function cgov_blog_preprocess_node(&$variables) {
```
bryanpizzillo: @kalidindir We try not to use preprocess hooks as there are usually better ways. The blog series cache tag should really be set the the various block plugins in the cgov_blog module. Those blocks are what actually depend on the series. If I were to remove those blocks from the blog_post displays, then the blog_post no longer has a cache dependency on the blog series. Additionally, by adding it on the blocks, you do not have to specifically check for the view mode of the post. This is even better if we ever create a separate view_mode that contains one of those blocks we would not have to update this logic.

The taxonomy_term_update looks pretty good and has nice comments. There is just one change I would make and I will add it there.

docroot/profiles/custom/cgov_site/modules/custom/cgov_blog/cgov_blog.module
```diff
@@ -218,3 +219,32 @@ function cgov_blog_preprocess_block(&$variables) {
...
+
+/**
+ * Implements hook_taxonomy_term_update().
+ */
+function cgov_blog_taxonomy_term_update($term_entity) {
+  /* When taxonomy term for blog categories is updated invalidate the
+   *  corresponding blog series page.
+   */
+  if ($term_entity->getVocabularyId() === "cgov_blog_topics") {
+    $owner_blog = $term_entity->get('field_owner_blog')->getValue();
```
bryanpizzillo: Change this to:
`$owner_blog = $term_entity->field_owner_blog`

Using the magic __get functions reduce the possibility of an exception being thrown. This could break if the term was programatically created and someone missed the owner target. (Migration does not check if things are created properly) So we try and use it whenever we can... some instances of get predate a bunch of bug fixes because the code did NOT use the magic __get.

docroot/profiles/custom/cgov_site/modules/custom/cgov_blog/cgov_blog.module
```diff
@@ -218,3 +219,32 @@ function cgov_blog_preprocess_block(&$variables) {
...
+    // Invalidate the cache tag for blog series node.
+    if ($owner_blog != NULL && $owner_blog[0]['target_id'] > 0) {
```
bryanpizzillo: If you know that $owner_blog should have only 1 owner, you can actually do:
`$owner_blog->target_id` 
The magic __get method gets the first item in a list.

docroot/profiles/custom/cgov_site/modules/custom/cgov_blog/cgov_blog.module
```diff
@@ -218,3 +219,32 @@ function cgov_blog_preprocess_block(&$variables) {
...
+      Cache::invalidateTags(['node:' . $owner_blog[0]['target_id']]);
```
bryanpizzillo: Same as above, just use
`$owner_blog->target_id`

PR #2385, opened 2019-08-29
[POC] NCIOCPL-000: Prototype for CGov Vocab Manager.

### LINKED ISSUE #1413: Taxonomy Management of Site Sections does not work

### LINKED ISSUE #2273: Can't reorder site sections
Within www-cms.cancer.gov Structure > Taxonomy > Site Sections, I tried to reorder site sections. I clicked saved at the bottom of the site sections page, but my changes are not actually saving. 

Users with access to the taxonomy should be able to reorder site sections and save those changes, to be seen on the front-end.

doylejd:
### Description
* Implements cgov_vocab_manager to provide an updated Vocabulary Management Experience.
* Adds core patch to allow default term overview pages to render and not blow up.  

### User Story Outline: 
Rebuild the Taxonomy Vocabulary Overview Page: 
* The Vocab Overview page should display only one level of terms at a time. 
  * Each term on the page should have a new button: “Manage Children” if the term has children. 
  * The user should have the same functionality as the current Taxonomy Vocab overview page:
    * Edit/Create/Delete/Reorder (based on perms)
* Once a user clicks on “Manage Children” of a term:
  * The page reloads with the next level of children. 
  * The breadcrumb is updated with the Parent Path so that there is a clear hierarchy visible on the page 
  * Each term on the page should have a new button: “Manage Children” if the term has children. 
  * The user should have the same functionality as the current Taxonomy Vocab overview page:
    *  Edit/Create/Delete/Reorder (based on perms)
* Repeat Step 2 for however many levels deep.
  * A user can change the parent term via the UI present in the edit page. This will not be a requirement of the Taxonomy Vocab Overview. (This may be a future enhancement). 

### Brief Architecture Overview
* Replicate the Drupal\taxonomy\Form\OverviewTerms Form into a custom module. 
* Add configurable logic to tree load so it doesn’t load every term:
  * vid - Vocabulary
  * parent - Parent term ID (Default to 0)
  * max-depth - Max depth of load (should be hard-coded to 1 for now)
  * load_entities	- Should it load full entities (should be hard-coded to false for now)
* Alter Listing to;
  * Add a new column for “Children”. This will link to a new listing of a new level of taxonomy. 
  * Only display one level of terms at a time. 
  * Only save terms that are present on the given form view. (This could be problematic if weights are manually changed to something outside of the scope of the current page. May want to consider removing the ability to manually reassign weights). 

### Vocab Management URL: 
* http://www.devbox/admin/structure/cgov_taxonomy_manager/<Vocab ID> 
* ie. http://www.devbox/admin/structure/cgov_taxonomy_manager/cgov_site_sections

### TODO:
* Enable module through config (need to re-export config)
* Automated Test Suite/Coverage
* Documentation
* Add Term functionality on a given level is not functioning. Currently just links to normal add form. 
* Consider UI/UX Improvements

Closes #1413
Closes #2273

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -205,7 +205,9 @@
                 "3029219 : State transition validation ignores language when determining \"first time moderation\"": "https://www.drupal.org/files/issues/2019-03-24/3029219-8.patch",
                 "3045648 : Use language for determining initial moderation state": "https://www.drupal.org/files/issues/2019-04-07/3045648-15.patch",
                 "3048288 : Avoid bogus MethodNotAllowedException; see https://github.com/NCIOCPL/cgov-digital-platform/issues/1716": "https://www.drupal.org/files/issues/2019-04-15/route-method-exception-path-fix-3048288.patch",
-                "3048464 : SubProcess migrate process plugin does not play nicely with XML-based source data": "https://www.drupal.org/files/issues/2019-04-30/3048464-8.patch"
+                "3048464 : SubProcess migrate process plugin does not play nicely with XML-based source data": "https://www.drupal.org/files/issues/2019-04-30/3048464-8.patch",
+                "763380  : Do not use \\Drupal\\taxonomy\\TermStorageInterface::loadTree() in \\Drupal\\taxonomy\\Form\\OverviewTerms::buildForm()" : "https://www.drupal.org/files/issues/2019-07-08/763380-61.patch"
+
             },
             "drupal/ctools": {
                 "2785499 : Entity Bundle Condition Plugin does not work in all situations": "https://www.drupal.org/files/issues/2019-03-01/ctools_fix_condition_plugin_entity_bundle-2785499-12.patch"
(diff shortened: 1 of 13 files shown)


bryanpizzillo:
There is an issue with commit 491a565 where the siteadmin user can login to a local install, but with all the ODE modules they cannot. When all the ODE modules are installed on the local machine, the siteadmin user cannot login. So there is some conflict between the vocab manager and the ODE modules.

bryanpizzillo:
acquia_connector adds an external link to the menu bar, Urls that are external links can not have getRouteName called. The site section code to replace the taxonomy link was not checking if the link was routed or not.

PR #2409, opened 2019-09-06
Fix click event analytics on CTHP and PDQ pages

### LINKED ISSUE #2327: Fix click event analytics on CTHP and PDQ pages so that -health_professional or -patient doesn't stack
Go to any CTHP or PDQ page, look at the analytics debugger and you will see that the page name is set to a value like "www.cancer.gov/types/liver/hp - health_professional". Click the toggle between patient and HP and observe the same. As you go back to the HP page or if you elicit click events on the page, you'll see that the "- patient" or "- health_professional" begin to stack and add multiple times over. This was previously fixed for post events (#2201 ) BUT click events are still an issue. 

The fix that was previously applied to for posts was:  

- Cleaned up setNciPageNameAdditions() and set pageName to omit concatenated local pn values

This is messing up analytics reports and needs to be fixed.

markecramer:
closes #2327

### DIFF
--- a/docroot/profiles/custom/cgov_site/themes/custom/cgov/src/libraries/analytics/AppMeasurement.custom.js
+++ b/docroot/profiles/custom/cgov_site/themes/custom/cgov/src/libraries/analytics/AppMeasurement.custom.js
@@ -432,11 +432,6 @@ var AppMeasurementCustom = {
             let s = this;
             let additions = [];
 
-            // Add audience.
-            if (s.prop7) {
-                additions.push(s.prop7);
-            }
-
             // Add dictionary types.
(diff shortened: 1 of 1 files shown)


lburack:
@markecramer - I'm not sure how to test this. Should I ask @welshja to do the "IA Review"?

welshja:
@markecramer - when I go to these pages and click on some of the links (like links with in a CTHP card or the on this page nav for pdq), I do see the call happening, but the page name is not set at all. We wanted to remove the "addition" to the pagename, but still need to see the pagename fire in those click events.

Also, I'm seeing an issue on http://ncigovcdode262.prod.acquia-sites.com in general where all click events are passed as "exit clicks' and "other clicks".

welshja:
@markecramer updated the DTM property to address the issue. Passes "IA" review on the ODE.

PR #2418, opened 2019-09-10
Analytics loading order issue

markecramer:
Changed javascript loading order to have DTM page_bottom loaded last.


markecramer:
ODE to test, http://ncigovcdode263.prod.acquia-sites.com/

PR #2448, opened 2019-09-16
(#2315, #2206) - Add CT cache tags, Update Site Section tag logic

### LINKED ISSUE #2206: Site Section Navigation Caching Issues
When pages within a section reside in the Akamai cache, and an advanced editor makes changes to the navigation for that section (by adding a taxonomy item, editing a taxonomy item, etc.)  - those pages are not updated within the Akamai cache and do not display the changes made to the section nav on the front end.

### LINKED ISSUE #2315: Implement Content Type Cache Tags
We need cache tags for the specific entity/bundle types so that we can easily bulk clear all content of that type from CDN/Cache. This is very useful on metadata changes. This would reduce the burden on the servers by not requiring the clearing of the entire cache.

adriancofie:
ODE: ncigovcdode268.prod.acquia-sites.com

Closes #2206 
Closes #2315 

Note: Acceptance Criteria are on the individual tickets.

- Adds the node bundle ID to the nodes cache tags. e.g:
   - cgov_article:full 
   - cgov_article:all-view-modes 
   - cgov_blog:all-view-modes 

- Adds the Section NavRoot ID to the Section Nav blocks cache tags. e.g:
    -  site_section:11844

- Adds logic to invalidate a given Section Nav root ID -- under the right circumstances.
    -  When relevant Site Section Fields are updated
    -  When Site Section or Pretty URL changes on a node 

- Cleans up / extracts the term re-save logic into a function.

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.module
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.module
@@ -601,4 +601,19 @@ function cgov_core_preprocess_image_formatter(&$variables) {
   ) {
     $variables['image']['#alt'] = "";
   }
+
+}
+
+/**
+ * Implements hook_entity_build_defaults_alter().
+ */
+function cgov_core_entity_build_defaults_alter(array &$build, EntityInterface $entity, $view_mode) {
+  // Add the bundle as a cache tag to help simplify cache invalidation.
+  if ($entity->getEntityTypeId() === 'node' || $entity->getEntityTypeId() === 'media') {
+    $tags = $view_mode == 'full' ? [$entity->bundle() . ':full'] : [];
+    $tags[] = $entity->bundle() . ':all-view-modes';
+    $build['#cache'] = [
+      'tags' => Cache::mergeTags($entity->getCacheTags(), $tags),
+    ];
+  }
 }
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_core/src/Plugin/Block/SectionNav.php
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_core/src/Plugin/Block/SectionNav.php
@@ -9,6 +9,7 @@ use Drupal\cgov_core\Services\CgovNavigationManager;
 use Drupal\Core\Session\AccountInterface;
 use Drupal\Core\Access\AccessResult;
 use Drupal\cgov_core\NavItem;
+use Drupal\Core\Cache\Cache;
 
 /**
  * Provides a 'Section Nav' block.
@@ -179,14 +180,16 @@ class SectionNav extends BlockBase implements ContainerFactoryPluginInterface {
    * {@inheritdoc}
    */
   public function build() {
+    $build = [];
+
     $navTree = $this->getSectionNav();
     if ($navTree) {
       $build = [
         '#type' => 'block',
-        'nav_tree' => $navTree,
+        '#nav_tree' => $navTree,
       ];
-      return $build;
     }
+    return $build;
   }
 
   /**
@@ -194,6 +197,32 @@ class SectionNav extends BlockBase implements ContainerFactoryPluginInterface {
    */
   public function getCacheMaxAge() {
     return 0;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheTags() {
+    $tags = [];
+
+    $navTree = $this->getSectionNav();
+    if ($navTree) {
+      // Set the cache tag if a nav root exists for this section.
+      $navRoot = $this->navMgr->getNavRoot('field_section_nav_root');
+      if ($navRoot) {
+        $section_root_tid = $navRoot->getTerm()->id();
+        $tags[] = 'site_section:' . $section_root_tid;
+      }
+    }
+
+    return Cache::mergeTags(parent::getCacheTags(), $tags);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheContexts() {
+    return Cache::mergeContexts(parent::getCacheContexts(), ['url.path']);
   }
 
 }
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/cgov_site_section.module
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/cgov_site_section.module
@@ -10,6 +10,7 @@ use Drupal\cgov_site_section\FieldStorageDefinition;
 use Drupal\Core\Field\FieldDefinition;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\taxonomy\Entity\Term;
+use Drupal\Core\Cache\Cache;
 
 /**
  * Implements hook_entity_field_storage_info().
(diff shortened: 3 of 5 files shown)


docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.module
```diff
@@ -601,4 +601,28 @@ function cgov_core_preprocess_image_formatter(&$variables) {
...
   }
+
+}
+
+/**
+ * Implements hook_entity_build_defaults_alter().
+ */
+function cgov_core_entity_build_defaults_alter(array &$build, EntityInterface $entity, $view_mode) {
+
+  // Add the bundle as a cache tag to help simpflify cache invalidation.
```
adriancofie: When an item is rendering as the full view-mode it will have both the :full and :all-view-modes cache tag.

Otherwise the rendered item will just contain :all-view-modes tag.

docroot/profiles/custom/cgov_site/modules/custom/cgov_core/src/Plugin/Block/SectionNav.php
```diff
@@ -179,14 +179,28 @@ public function renderNavElement(NavItem $navItem, int $renderDepth, int $curren
...
       $build = [
         '#type' => 'block',
-        'nav_tree' => $navTree,
+        '#nav_tree' => $navTree,
       ];
-      return $build;
+
+      // Set the cache tag if this if a nav root exists for this section.
+      $navRoot = $this->navMgr->getNavRoot('field_section_nav_root');
+      if ($navRoot) {
```
adriancofie: Logic for adding the section root as a cache tag.

docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/templates/block/section_nav/block--cgov-section-nav.html.twig
```diff
@@ -1,3 +1,5 @@
+{% set nav_tree = content['#nav_tree']%}
```
adriancofie: Modifications for this file are related to adding the section root as a cache tag.

docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/cgov_site_section.module
```diff
@@ -119,35 +164,178 @@ function cgov_site_section_taxonomy_term_update(EntityInterface $entity) {
...
+ * Determines if a section should trigger a re-save of its children terms.
+ *
+ * @param \Drupal\Core\Entity\EntityInterface $updated_section_entity
+ *   The site section entity with its pending value updates.
+ *
+ * @return bool
+ *   A TRUE or FALSE as to whether to trigger a re-save.
+ */
+function will_trigger_save_children(EntityInterface $updated_section_entity) {
+
```
adriancofie: Returns whether or not a site section  should update it children, based on if certain fields have changed.

docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/cgov_site_section.module
```diff
@@ -119,35 +164,178 @@ function cgov_site_section_taxonomy_term_update(EntityInterface $entity) {
...
+ *   The entity just updated.
+ *
+ * @return bool
+ *   A TRUE or FALSE as to whether to trigger a cache clear.
+ */
+function _will_trigger_cache_tag_invalidation(EntityInterface $updated_entity) {
+  $values = [];
+
+  // Invalidation Check For Nodes/Media.
+  if ($updated_entity->getEntityTypeId() === 'node' || $updated_entity->getEntityTypeId() === 'media') {
```
adriancofie: Whether or not a node/media should invaildate it's term taxonomy.

docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/cgov_site_section.module
```diff
@@ -119,35 +164,178 @@ function cgov_site_section_taxonomy_term_update(EntityInterface $entity) {
...
+    }
+    // Set entity Fields which trigger invalidation when modified.
+    $values =
+      [
+        'field_site_section',
+      ];
+  }
+
+  // Invalidation Check for Site Sections.
+  if ($updated_entity->getEntityTypeId() === 'taxonomy_term' &&
```
adriancofie: Whether or not a term should invalidate it's term taxonomy.

docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/cgov_site_section.module
```diff
@@ -119,35 +164,178 @@ function cgov_site_section_taxonomy_term_update(EntityInterface $entity) {
...
+ * @param array $fields
+ *   The fields which trigger a cache invalidation.
+ * @param \Drupal\Core\Entity\EntityInterface $entity
+ *   The updated entity.
+ *
+ * @return bool
+ *   A TRUE or FALSE as to whether the fields were modified.
+ */
+function entity_fields_modified(array $fields, EntityInterface $entity) {
+  // Store the original entity for comparison.
```
adriancofie: Compares entity fields in a flexible manner

docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.module
bryanpizzillo: I think this will work. I would request that the actual build array be modified only in one place...
```
$tags = $view_mode == 'full' ? [$entity->bundle() . ':full'] : [];
$tags[] = $entity->bundle() . ':all-view-modes';
$build['#cache'] ...
```

docroot/profiles/custom/cgov_site/modules/custom/cgov_core/src/Plugin/Block/SectionNav.php
bryanpizzillo: I think you are also going to have to add a context here, right? To tell it to vary based on ID right now... oh right, the max-age is still 0, so it would not get cached anyway.

Also, I have stared at the sun and adding tags to the build array is not actually the best way in a block's build. The build is actually called as a `#pre_render` function AFTER the thing has been rendered. Technically it should be done in getCacheTags().

docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/cgov_site_section.module
bryanpizzillo: This should also check pretty_url_name since that makes up the HREF in the nav.

docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/cgov_site_section.module
bryanpizzillo: Nice bit of magic here!

docroot/profiles/custom/cgov_site/modules/custom/cgov_core/src/Plugin/Block/SectionNav.php
adriancofie: I held off on adding the context since I reverted back to max-age = 0, but I guess it doesn't hurt to add it now.

The plot thickens...I'll give it a look.

docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/cgov_site_section.module
adriancofie: Does it really though? This is where I got thrown a bit. 

If a page is in the Section Nav, doesn't that mean it's always a landing page? And landing pages do not set a pretty URL on the node/media , their path is derived from the Site Section pretty URL (via the computed path field). 

Is there a scenario where a page is set as a landing page but doesn't have an empty pretty URL field on the node/media?

docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/cgov_site_section.module
bryanpizzillo: Looks good.

docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/cgov_site_section.module
bryanpizzillo: I see what you are getting at, but technically any node can go in as a landing page. Which means that whatever makes up its URL needs to be checked for changes. (Also, Media does not need to be checked because only nodes can be landing pages.)

VictoriaSunNIH:
Looks good - I changed the site section order, renamed a site section, and I updated a site section pretty URL and republished a node associated with that site section to make sure it took on the new site section.  All of it worked as expected.  Passes Review!

PR #2488, opened 2019-10-17
(#839) Prevent deletion of Cgov Image media that have ever been published

### LINKED ISSUE #839: Prevent users from deleting already published cgov_images
Basically, the same workflow rule as content. An image_manager should be able to delete any images that are draft and never have been published. However, once they have been published they may be in use, and therefore no one should be allowed to willy-nilly delete them. 

There may be limitations with Drupal core WRT media items and being able to hook into the permission checks like we do with content.

adriancofie:
Closes #839
ODE: http://ncigovcdode278.prod.acquia-sites.com/

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_image/cgov_image.module
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_image/cgov_image.module
@@ -118,7 +118,12 @@ function _cgov_image_filter_view_image(
   }
 
   // Only certain image types will have overrides.
-  if (in_array($img_type, ['featured', 'panoramic', 'social_media', 'thumbnail'])) {
+  if (in_array($img_type, [
+    'featured',
+    'panoramic',
+    'social_media',
+    'thumbnail',
+  ])) {
 
     $override_field_name = 'field_override_img_' . $img_type;
     $override_field = $entity->get($override_field_name);


VictoriaSunNIH:
@adriancofie - @lburack and I are unable to test this because the ODE says "The website encountered an unexpected error. Please try again later."

Can you please redeploy and tag us when this is ready for testing?  Thanks.

adriancofie:
@lburack @VictoriaSunNIH This has been re-rolled and is on: http://ncigovcdode278.prod.acquia-sites.com/

lburack:
@adriancofie @VictoriaSunNIH  - the following happened, as expected:
- I was able to delete a draft, unpublished image
- I was unable to delete a published image

### **BUT** 
we still need a way to delete images (just not "willy-nilly", as @bryanpizzillo said in #839). 

Site Admins can archive content, which they can then delete. This is the same workflow rule that images should have. We should have specified these requirements in the ticket #839. I'd like to talk with @mworrest and @blilianyu tomorrow about permissions, as we only have a handful of Site Admins, but not sure we want all Image Managers to have this capability. I will update the requirements after I speak with them.

VictoriaSunNIH:
Moving this back to needs requirements due to the additional comments

adriancofie:
For legacy code see:  https://github.com/NCIOCPL/cgov-digital-platform/commit/c4cd0318f10f921e559f392789380913c40ab290

VictoriaSunNIH:
Closing this PR for now, taking care of this in future PRs

PR #2576, opened 2019-11-22
Update CTS to v0.6.0

arcepaul:
- fix for populating stores based on queries -- _looks good, but there are still outstanding bugs_
- re-ordered delighter on trial description page -- _some issues with styling on mobile, noted in box_
- search criteria box header width adjusted per request _looks good_
- clear form button added to sticky block _looks good, has some mre styling needed_
- extra label on cancer type keyword removed _looks good_


VictoriaSunNIH:
This looks good - there are new issues being tracked separately for Clear Form, and the delighter on mobile/tablet needs to be centered but we'll also track that separately.

PR #2579, opened 2019-11-25
Feature/cts update v0.7.2

arcepaul:
- incorporates Bryan's tests for formatting a trial search query
- fixed issue with display of nearby locations on trial description page
- fixed various issues with repopulating store from pasted in query string
- fixed issue with store updating when paging through results


arcepaul:
@VictoriaSunNIH @lburack 
An ODE for this PR is available at: http://ncigovcdode300.prod.acquia-sites.com/about-cancer/treatment/clinical-trials/search

bryanpizzillo:
Issue:
Enter `Advanced Breast Carcinoma` (ABC) into the basic form. Search. Modify results. The primary cancer type is ABC on the advanced form. That is not possible because ABC is not a main type. On both prod and 283 on the advanced form it should be Primary Cancer: Breast Cancer, Subtype: ABC.

VictoriaSunNIH:
@arcepaul  - I'm still not able to use the back button after I use the pager to navigate back through the pages on ODE 300.

VictoriaSunNIH:
@arcepaul  - another bug I'm noticing, when I select a state from the dropdown list (once I've filtered the search by state), there is no state heading appearing above the list

PR #2587, opened 2019-12-05
Update CTS app to v0.10.0

arcepaul:
This method converts a query string to a form object
that can then be used to update the form store.

Integrated queryStringToFormObject with app. Removed
old query builder code.

Also implemented fetcher utilities to be used by
queryStringToFormObject.

Implemented many many tests.


arcepaul:
v0.10.1 fixes include:

- [ ] Meta descriptions are swapped: basic is showing advanced search description copy

- [ ] Update "NCI Contact Center" to be "Cancer Information Service"
- [ ] Clinical**T**rials.gov
- [ ] Location field doesn't have an option for "District of Columbia" and when you search for it, results show everything in "United States" (this is on the live site and shows 245 results)
- [ ] Need to add coverage on TrialStatusIndicator for more statuses(Administratively Complete, Complete, and Withdrawn) 
- [ ] On Trial Type helper text, move the period outside the quotation mark ("All." should be "All". )
- [ ] Help icon on the Age field is linking to https://www.cancer.gov/about-cancer/treatment/clinical-trials/search/help#basicsearch instead of https://www.cancer.gov/about-cancer/treatment/clinical-trials/search/help#age
- [ ] The open all, close all option also applies to the show search criteria box as well as the accordions, and should only apply to the content within the accordions.
- [ ] When i search for all results, I only get a start over option, with no "Back to search results". This should also appear on the trial description page when i have searched for all results, and not only for searches with criteria.
- [ ] Font is incorrect on headers on accordions. Should be Montserrat like live site.

PR #2589, opened 2019-12-05
update CTS app to v0.11.x

arcepaul:
Updating the CTS app package to v0.11


VictoriaSunNIH:
0.11.0 Issues I'm seeing on 305:

- @arcepaul - When modifying search, the VA and Healthy Volunteer options which were previously selected to "yes" are reset to "no" (visually), but when I resubmit the search, I still see that the parameter is there in the URL

@slonge - For Drugs, there are odd bullets in the list, as well as stray bullets.  It also looks like the area that I can select to input is not the entire field, but only part of it

@bryanpizzillo - Zipcode - the number of locations is working in results, but the locations filter on trial description is showing me all locations instead of just NIH ones (but I'm realizing that was a bug that is also on 283 and on dev acsf)

Issue #2648, opened 2020-02-28
Activate fast_404 handling


blairlearn:
# Turn on Drupal fast_404 handling.

By default, Drupal renders a fully themed error page for all 404 errors. This is computationally expensive and when many errors occur (e.g. an appscan), it can slow the server. The fast 404 handler skips the theming and returns a simple "page", the contents of which are stored inline in a settings file (in our case, `docroot/sites/settings/cgov_core.settings.php`).

A [fast_404 module](https://www.drupal.org/project/fast_404) is available, but as of this writing, it is still in alpha and is likely overkill.

> **ESTIMATE** 2

### Resources:
* [Acquia fast 404 documentation](https://support.acquia.com/hc/en-us/articles/360005246714-Using-Fast-404-to-improve-Drupal-performance)  Note that it talks about modifying the core settings.php which is not the correct file to edit.
* [Fast 404 notes in settings.php](https://github.com/NCIOCPL/cgov-digital-platform/blob/develop/docroot/sites/default/settings.php#L658-L683)

### Prerequisites
Determine what to handle
* Static files: txt, png, gif, jpeg, jpg, css, js, ico, swf, flv, cgi, bat, pl, dll, exe, asp, aspx, svg
* Pages?  (Requires putting friendly error page in Akamai)
* Are there specific paths which should get themed 404s?  (Exclude paths)

blairlearn:
Fast 404 is active "out of the box".

Compare:
https://www.cancer.gov/chicken (themed 404 page) vs
https://www.cancer.gov/chicken.jpg (ugly 404 page)

blairlearn:
Expected behavior:

These non-existent pages should display themed "Page not found" messages.
* /ErrorPages/404
* /espanol/ErrorPages/404
* /rare-brain-spine-tumor/ErrorPages/404
* /rare-brain-spine-tumor/espanol/ErrorPages/404
* /nano/ErrorPages/404
* /pediatric-adult-rare-tumor/ErrorPages/404

All other non-existent pages (e.g. /chicken or /espnaol/pollo) should display an unthemed ("ugly") message saying the path cannot be found.

(On production, the unthemed messages will be replaced by themed messages via Akamai rules.)

PR #2649, opened 2020-02-28
(#2648) Activate fast 404 handling.

blairl-nih:


blairlearn:
Abandoning.  All this does is duplicate functionality which is already in place.

Issue #2644, opened 2020-02-26
CSS is not minified in production


blairlearn:
### Issue description
The production CSS files are either not being minified, or only set for 'lite' minification.

### Steps to reproduce the issue

1.  Open browser to https://www.cancer.gov/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/dist/css/Common.css?q5mc6m

### What's the expected result?
- Multiple CSS properties on a single line.
- CSS selectors reduced to a single line.
- No (or fewer) comments.
- Little readability.

### What's the actual result?
- One line per property.
- Most selectors have individual lines
- Comments noting sections of the source SCSS files.
- Leading whitespace.

### Additional details / screenshot
- ![image](https://user-images.githubusercontent.com/21063180/75365218-dd509b80-588a-11ea-9cc4-ccf1446260f5.png)

blairlearn:
Common.css went from 302KB to 219KB
Common.js went from 838KB to 326KB

PR #2652, opened 2020-03-02
(#2644) Turn minification back on.

blairl-nih:


docroot/profiles/custom/cgov_site/themes/custom/cgov/webpack.base.config.js
```diff
@@ -30,7 +31,10 @@ const config = {
 					{
 						loader: MiniCssExtractPlugin.loader
 					},
-					'css-loader',
```
arcepaul: this block can be reverted as it doesn't do anything.  The minify option was deprecated after v1.0.0

blairlearn: Done!

PR #2738, opened 2020-04-03
(#2729) Patched Metatags Module

### LINKED ISSUE #2729: Enabler: Update drupal/metatags module
# Update drupal/metatags

The `drupal/metatags` module has an outstanding security issue.  This will cause builds to fail.

It is also necessary to review our use of the "2981793: Allow metatag configuration to be disabled" patch.  (Has it been made part of the module?)

> **ESTIMATE** xx

image

bryanpizzillo:
There was an issue that was found after we pulled metatags into the hotfix
branch. Added a patch for metatag 1.1.13, which is in development.

Closes #2729


bryanpizzillo:
I botched this and pulled in the wrong PR. There will be one to replace this.

Issue #2762, opened 2020-04-21
Story: Support YAML Content Embeds


bryanpizzillo:
## As a tester, I want the ability to add embedded elements into the YAML markup, so that I can build automation tests for our embedded content

> **ESTIMATE** NN

### Acceptance Criteria

```gherkin
Scenario: [title]
  Given [context]
    And [some more context]
  When  [event]
  Then  [outcome]
    And [another outcome]
```

### Notes
* ...

---

## Technical Notes
* Embedded items use the UUID of content which is not something we can predict. We need to have some sort of way to set the `data-entity-uuid` attribute to a string that would map to a lookup.
* Entity Embed example (This is for an embedded image using the "right medium" template
   ```html
     <drupal-entity
        data-embed-button="media_entity_embed" 
        data-entity-embed-display="view_mode:media.image_display_article_medium"
        data-entity-type="media"
        data-entity- uuid="084f90fa-2aef-47d1-a377-e1a72d297801"
        data-align="right"
        data-caption=""
     ></drupal-entity>
   ```
* YAML Content Module Reference Field Process
   * Node
      ```
      field_featured_item:
         - target_type: 'node'
            '#process':
                 callback: 'reference'
                    args:
                       - 'node'
                       - type: 'cgov_article'
                          title: 'Frisco - The Roo Roo Kid'
      ```
   * Media
      ```
      field_image_article:
         - target_type: 'media'
            '#process':
                 callback: 'reference'
                    args:
                       - 'media'
                       - bundle: 'cgov_image'
                          name: 'Frisco - King of Shenandoah' 
      ```
   * Term ID
      ```
      field_site_section:
         - '#process':
            callback: 'reference'
            args:
                - 'taxonomy_term'
                - vid: 'cgov_site_sections'
                   computed_path: '/pets/frisco'
      ```

### Possible Solution 1: Use Regex
Use regex to replace the data-entity-uuid attribute with the correct UUID. This will follow the #process directive use for entity references.
```
<drupal-entity data-embed-button="media_entity_embed" data-entity-embed-display="view_mode:media.image_display_article_medium" data-entity-type="media" data-entity-uuid="#process('<PROCESS_TYPE>', '<ENTITY_TYPE>', '<BUNDLE>', '<CRITERIA>')" data-align="right" data-caption=""></drupal-entity>
```
Where the following parameters are:
* **PROCESS_TYPE** - The YAML content module's process type. This should probably just be `reference`
[...]

PR #2771, opened 2020-04-29
(#2762) Implement Entity Embed Support in Yaml

bryanpizzillo:
- It is slightly dirty. There was no direct way to add it to our
   yaml_content event subscriber class. So I had to add a
  entity_presave hook. I do check that it is in the CLI at
  least so it does not fire off when in the UI.
- Added a test YAML to ensure it works.

Closes (#2762)

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_yaml_content/cgov_yaml_content.module
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_yaml_content/cgov_yaml_content.module
@@ -1,0 +1,43 @@
+<?php
+
+/**
+ * @file
+ * Contains cgov_yaml_content.module.
+ */
+
+use \Drupal\Core\Entity\EntityInterface;
+
+/*
+ * This is going to be a fairly short treatise on the limitations of yaml_content module's
+ * event system. Basically the events like FIELD_IMPORT would lead to believe that you
+ * can manipulate the data coming in (getContentData), however, you cannot as the content
+ * data is immutable. After putzing for a while to find ways for us to replace the
+ * <drupal-entity> UUID attributes with the correctly loaded UUID, there appeared to be
+ * no sane way to accomplish this. (The yaml_content authors agreed it is a known limitation)
+ * SO... What we are going to do is to write some presave hooks that fire only when we
+ * are running the yaml_content module. (Oh, I should note that Paragraph entities got
+ * insanely complex to even attempt to fix pre-save as the entity revision was already
+ * saved at the point of pre-save for the node)
+ */
+
+/**
+ * Implements hook_entity_presave().
+ */
+function cgov_yaml_content_entity_presave(EntityInterface $entity) {
+  // Are we running from the command line?
+  if (PHP_SAPI !== 'cli') {
+    return;
+  }
+
+  /* TODO: Only run if we are in a YAML_Content command */
+  /* This may not be possible. */
+
+  // Only do this for entities implementing ContentEntityInterface.
+  $class = new \ReflectionClass($entity);
+  if (!in_array('Drupal\Core\Entity\ContentEntityInterface', $class->getInterfaceNames())) {
+    return;
+  }
+
+  $entity_embed_processor = \Drupal::service('cgov_yaml_content.yaml_entity_embed_processor');
+  $entity_embed_processor->processEntity($entity);
+}
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_yaml_content/src/Service/YamlEntityEmbedProcessor.php
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_yaml_content/src/Service/YamlEntityEmbedProcessor.php
@@ -1,0 +1,209 @@
+<?php
+
+namespace Drupal\cgov_yaml_content\Service;
+
+use Drupal\Core\Entity\ContentEntityInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Field\FieldItemListInterface;
+use Drupal\Core\TypedData\Exception\MissingDataException;
+use DOMWrap\Document;
+
+/**
+ * Helper service for YAML content loading for <drupal-entity> tags.
+ *
+ * This service will replace <drupal-entity> UUID attributes with the UUIDs of
+ * content that had been previously loaded in our content loading.
+ *
+ * NOTE: For sanity, this only supports ContentEntities as they can have fields.
+ */
+class YamlEntityEmbedProcessor {
+
+  /**
+   * Drupal Entity Type Manager.
+   *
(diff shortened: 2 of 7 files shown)


bryanpizzillo:
ODE: http://ncigovcdode343.prod.acquia-sites.com/
The sample content can be viewed at:
* English: http://ncigovcdode343.prod.acquia-sites.com/test/embed-poc
* Spanish: http://ncigovcdode343.prod.acquia-sites.com/espanol/test-es/embed-poc-es

Information on how to make your own can be found in the ticket under "Possible Solution 2: Use Dom Wrap"

docroot/profiles/custom/cgov_site/modules/custom/cgov_yaml_content/cgov_yaml_content.module
```diff
@@ -0,0 +1,43 @@
...
+ * insanely complex to even attempt to fix pre-save as the entity revision was already
+ * saved at the point of pre-save for the node)
+ */
+
+/**
+ * Implements hook_entity_presave().
+ */
+function cgov_yaml_content_entity_presave(EntityInterface $entity) {
+  // Are we running from the command line?
+  if (PHP_SAPI !== 'cli') {
```
adriancofie, line 28: @bryanpizzillo How do the PDQ imports run? Is there a performance concern, presuming every entity will make it to `$entity_embed_processor->processEntity($entity);`

docroot/profiles/custom/cgov_site/modules/custom/cgov_yaml_content/src/Service/YamlEntityEmbedProcessor.php
```diff
@@ -0,0 +1,209 @@
...
+   *   The string to transform.
+   */
+  public function transformEmbeds($value) {
+    $doc = new Document();
+    $doc->html($value);
+    $allEmbeds = $doc->getElementsByTagName('drupal-entity');
+
+    // Note: Don't use a foreach; it doesn't work. Must loop backwards.
+    // Stolen from The Developer Formerly Know As Adrian's Migrate code.
+    // Ask ‽ why it does not work. :)
```
adriancofie, line 86: 🤦

adriancofie:
Sidenote: Is the YAML content used anywhere in ACSF for testing purposes? Is there any reason to have the yaml_content or cgov_yaml_content enabled at all in ACSF? Are they currently?

docroot/profiles/custom/cgov_site/modules/custom/cgov_yaml_content/cgov_yaml_content.module
bryanpizzillo: PDQ import run in the RESTApi, so that will not be `cli`.
But yeah, every entity will make it to the processor. I do not trust the hook_HOOK_presave hooks as I have experienced them getting missed. I REALLY did not like this approach, but there was little that could be done with `yaml_content` without rewriting it. In the Drupal slack channel, the maintainers acknowledge that updating the content is a gap...

bryanpizzillo:
> Sidenote: Is the YAML content used anywhere in ACSF for testing purposes? Is there any reason to have the yaml_content or cgov_yaml_content enabled at all in ACSF? Are they currently?

It is not. It is actually only enabled when the installer runs, which does not happen on ACSF (yet). The idea (someday) is to have 2 microsites in ACSF to test the YAML content. But the modules would only be turned on for those sites.

docroot/profiles/custom/cgov_site/modules/custom/cgov_yaml_content/src/Service/YamlEntityEmbedProcessor.php
```diff
@@ -0,0 +1,209 @@
...
+    }
+  }
+
+  /**
+   * Function for processing <drupal-embed> tags in a text_long field.
+   *
+   * @param \Drupal\Core\Field\FieldItemListInterface $field
+   *   The Field to process.
+   */
+  public function handleEmbedForTextLong(FieldItemListInterface $field) {
```
blairlearn, line 63: A style question -- what's the reason for this method being declared public? I don't think it's a problem per se, but it's only used by this class.  Not even from a unit test.  Not the first place I've seen this in our code base, but it seems like a potential point of confusion for future consumers of the class.

docroot/profiles/custom/cgov_site/modules/custom/cgov_yaml_content/cgov_yaml_content.module
bryanpizzillo: Ok, so I am just going to uninstall cgov_yaml_content after installation on the ODEs. It will stay on for local installs, since, well how else can you test content...

docroot/profiles/custom/cgov_site/modules/custom/cgov_yaml_content/src/Service/YamlEntityEmbedProcessor.php
```diff
@@ -0,0 +1,209 @@
...
+   *   The entity type.
+   * @param array $filter_params
+   *   The filters for the query conditions.
+   */
+  protected function throwParamError($error_message, $entity_type = "ENTITY TYPE UNKNOWN", array $filter_params = []) {
+    // Build parameter output description for error message.
+    $error_params = [
+      '[',
+      '  "entity_type" => ' . $entity_type . ',',
+    ];
```
blairlearn, line 199: The value for the `$entity_type` variable will not be enclosed in quotation marks.  E.g. if `$entity_type` contains the value "foo", this will result in the second array value being `"entity_type" => foo,`. 

I believe the intent is for that value to be `"entity_type" => "foo",` (or, more likely, "node" or "media" or some real entity type).

blairlearn:
It works as expected. 😀  Please address that one issue and I'm good with it.

docroot/profiles/custom/cgov_site/modules/custom/cgov_yaml_content/src/Service/YamlEntityEmbedProcessor.php
bryanpizzillo: In case/when we add tests for this at some point. It is a fair point, but I was just going to leave them public for now.

docroot/profiles/custom/cgov_site/modules/custom/cgov_yaml_content/src/Service/YamlEntityEmbedProcessor.php
bryanpizzillo: It is fixed. I also fixed where I copied it from.

docroot/profiles/custom/cgov_site/modules/custom/cgov_yaml_content/src/Service/YamlEntityEmbedProcessor.php
blairlearn: 👍

PR #2803, opened 2020-06-16
Ticket/2796 glossary app config updates

sarinapadilla:


bryanpizzillo:
@sarinapadilla I assume this is outdated and should be deleted correct?

sarinapadilla:
@bryanpizzillo yes, somehow this PR still exists though I thought I changed it's base to be 0624...which already has the commits. This can be deleted.

PR #2520, opened 2019-10-24
(#2519) Integrate R4R App

bryanpnih:
Created cgov_r4r module & pulled in JS.

Not really tested.

Closes #2519

PR #2836, opened 2020-07-07
R4R app module

arcepaul:
Closes #2520 .

This effort involves moving R4R into the app module ecosystem. 

- Created libraries and entrypoints
- published R4R package
- updated page options for special R4R case
- created entry for r4r in admin


arcepaul:
First iteration of R4R up!
http://ncigovcdode285.prod.acquia-sites.com/research/resources

PR #2856, opened 2020-07-22
(#2851) Update glossary app package to v1.0.3

### LINKED ISSUE #2851: Update Glossary App Package for GlossaryApp Issue 192
Ticket to update the glossary app package for https://github.com/NCIOCPL/glossary-app/issues/192, for a fix for the bug where User is not directed to page with list of term definitions when there is one term definition for a letter.

arcepaul:
Closes #2851 .

- fix single result listing for expand


arcepaul:
ODE will be available at: 
http://ncigovcdode371.prod.acquia-sites.com/

VictoriaSunNIH:
I tested the genetics dictionary (expand K is the only one I know of right now), and this works as expected. Marking as passed

lburack:
@VictoriaSunNIH  - 'K' in the Genetics Dictionary is the only one I'm aware of too. Looks good to me as well!

VictoriaSunNIH:
Special characters, and the no results found dropdown issue seems resolved too.  Marking as passed!!

PR #2870, opened 2020-08-03
(#2869) Enable Caching of Main Menu

### LINKED ISSUE #2869: The main/mobile/mega menu does not cache and is SLOW
### Issue description
Most of the time rendering a page is spent rendering the Main/Mobile/Mega menu (i.e. The MainNav block) . This is not cached so EVERY page renders the main menu. This has slowly become an issue during deployments to the point that clearing the Akamai cache makes the CMS portion of the web site impossible to use, and has started affecting the rendering of pages too.

It is ok if the menu is slow, as long as it only has to be be built a few times after a cache clearing. (It would be nice if it was generated quickly too...)

> **ESTIMATE** NN

Sample New Relic Log
image

### LINKED ISSUE #2875: Memcache for ACSF became disabled in last BLT update
### Issue description
The memcache settings injected into acsf in `<DRUPAL_ROOT>/factory-hooks/post-settings-php/memcache.php` was updated to Acquia's new contents as part of BLT 10.x update. This change in 10.x moved the Acquia recommended settings out of the BLT package and into its own acquia/memcache-settings package. The memcache.php script references this. `acquia/memcache-settings` was never added to our composer.json. 

This is why we have been having horrible performance issues on prod upon cache clearing. (I mean the nav was horrible, but this is more horrible.

> **ESTIMATE** NN

bryanpizzillo:
In addition to caching the Main Menu it was found that memcache was not enabled, which was the underlying cause of everything being slow. Now we should be back to normal slow AND faster because of the nav. (The nav will still be slow to generate for now)

Closes  #2869
Closes #2875

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,6 +41,7 @@
     "require": {
         "acquia/acsf-tools": "dev-9.x-dev",
         "acquia/blt": "^10",
+        "acquia/memcache-settings": "^1.0",
         "behat/mink-selenium2-driver": "dev-master as 1.3.x-dev",
         "composer/installers": "^1.2",
         "cweagans/composer-patches": "^1.6.5",
@@ -62,7 +63,7 @@
         "drupal/image_widget_crop": "^2.2",
         "drupal/imagemagick": "~2.6",
         "drupal/linkit": "^5.0",
-        "drupal/memcache": "2.0-alpha7",
+        "drupal/memcache": "^2.1",
         "drupal/metatag": "^1.12",
         "drupal/migrate_file": "^1.1",
         "drupal/migrate_plus": "4.x-dev",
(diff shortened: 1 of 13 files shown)


bryanpizzillo:
ODE: http://ncigovcdode380.prod.acquia-sites.com/

alinai:
The accordion for El Cancer always appears expanded on loading the Spanish version of Home Page

image

alinai:
The following PDQ summary is not displaying in accordions on mobile in all browsers (Chrome, FF, Edge, Safari, IE)
https://www-test-acsf.cancer.gov/about-cancer/advanced-cancer/planning/end-of-life-pdq

image

alinai:
The following issue is resolved on test ACSF. (I had commented it in another ticket)

The mobile menu looks like the following on navigating to News & Events-> Conferences (+) -> ASCO 2020 (note: the screen on Prod was refreshed to see the Collapse (+) buttons Expanded (-) [otherwise all the Expand (-) buttons appears Collapsed (+) on Prod]

PR #2931, opened 2020-10-01
(#2929) Generic App Module Support

### LINKED ISSUE #2888: Breadcrumb is displaying on non-home/landing-page-type children of homepage on non-English Cancer.gov sites
Currently, the breadcrumbs are displaying on direct children of the homepage on non-Cancer.gov English sites. This is inconsistent with the behavior on Cancer.gov English and DCEG, where the breadcrumbs are not displayed on any direct children of the home page.

> **ESTIMATE** 5 (including testing across all microsites and Spanish!!)

1.  Go to any of the following sites: 
https://www.cancer.gov/espanol/publicaciones
https://www.cancer.gov/espanol/acerca-sitio
https://www.cancer.gov/pediatric-adult-rare-tumor/research
https://www.cancer.gov/nano/grants-training
https://www.cancer.gov/rare-brain-spine-tumor/blog
2.  See that the breadcrumb is displaying, with only the homepage link in it.

- The breadcrumb should not display for any of the direct children of the homepage of the microsite or Spanish site.

### What's the actual result?
- The breadcrumb displays only the homepage link.

### Here's how to fix it
Update Breadcrumb.php:
```
// Requirement: If the only breadcrumb is the root, we
// don't want to render any breadcrumbs.
$homeUrl = Url::fromRoute('<front>')->toString();

if (count($formattedBreadcrumbs) === 1 && $formattedBreadcrumbs[0]['href'] === $homeUrl) {
  $formattedBreadcrumbs = [];
}
return $formattedBreadcrumbs;
```

Also remove the temporary fix from sitewideSearchApp.js:
```
// This is a temporary fix for removing the breadcrumb element that displays on Spanish and microsite search results
// pages. REMOVE THIS when the sitewide fix for the breadcrumbs is implemented!
let breadcrumbs = document.querySelector(".breadcrumbs");
if(breadcrumbs !== null && breadcrumbs.parentNode) {
  breadcrumbs.parentNode.removeChild(breadcrumbs);
}
```

### LINKED ISSUE #2929: Enabler: Implement Generic JS App Module
# Implement Generic JS App Module
The initial ticket is to implement the Drug Dictionary App Module. However, we have transformed this into a "create a generic JS App Module" that would allow us to release and update apps as just JS files.

 ## JS-Only App Module Plugin.
 
 This app_module/plugin is used for those React-based app modules that do not need any back-end Drupal support. This does require that the app modules follow a specific config format for the JSON data. Until app modules have the ability to draw a UI for their options, OR at least some validation method exists, the following will have to do.
**(Please be careful)**
 
### Example Config
```
 {
   "drupalConfig": {
      "appName": "your-apps-name-here",
      "rootId": "NCI-app-root",
      "initFnName": "window.appModuleInit",
      "appCssUri": "https://someserver/app-name/app-name.vX.Y.Z.css",
      "appJsUri": "https://someserver/app-name/app-name.vX.Y.Z.js",
      "removeHeadElements" : [
        ...
      ],
      "validAppPaths": [
        ".*"
      ]
   },
   "frontEndConfig": {
     ...This will be passed into the initialize function...
   }
 }
```
 
**The config has two parts:**
- drupalConfig -- this is configuration for Drupal to know how to load the app.
- frontEndConfig -- this is configuration for the app itself. This will be passed into the initialize method. (It will be combined with some other Drupal path information.) You can make this whatever you want.
 
**The drupalConfig is made up of the following:**
- appName -- This is a unique ID for your app. (e.g. 'glossary-app-doct') try and make multiple instances of the app have a unique name please.
   >**_NOTE:_** This must be appropriate for being the key of an array.
- rootId -- this is the ID of the div tag for this instance, and what will get passed into the initialize function.
- initFnName -- The full name of the init function, we assume the app's init function will be attached to the window, but please include that too. The following items will be passed into the init method:
  * `baseHost` - Source: the `<current>` route's Scheme and host.
  * `basePath` - Source: path from the node's canonical url as fetched from `$node->uriRelationships`; if canonical does not exist, then use the path from the `<current>` route.
  * `canonicalHost` - Source: Scheme + host from the node's canonical url as fetched from `$node->uriRelationships`. Otherwise, if that is not available, the value used for `baseHost`.
  * `language` - Source: the current interface's language.
  * `rootId` - the value of `$config['drupalConfig']['rootId']`
  * All the values of `$config['frontEndConfig']` processed using the Front-end Config Value Processing rules below.
- pathToCss -- This is the full URL to the CSS file for the app. Make sure it has a version number, and never update that asset. (Or caching issues will occur)
- pathToJs -- This is the full URL to the JS file for the app. Make sure it has a version number, and never update that asset. (Or caching issues will occur)
- removeHeadElements -- The Drupal head elements your module will handle and should not be generated by Drupal lest Google get the wrong information. This list maps to keys in `$attachments['#attached']['html_head']`, except for `alternate` which will remove any `<link rel="alternate">` elements. A key for the `<title>` tag has been added, `title_tag`.
   * A good list of tags for Google niceness is `canonical_url`, `robots`, `title_tag`, `alternate`. **MAKE SURE YOUR APP ACTUALLY GENERATES THESE IF THEY ARE REMOVED.**
- validAppPaths -- This is a list of regular expressions that will ensure that your app has a route for that path. '/' is always allowed, so this is for any URL OTHER than '/'. If the requested path does not match, then a 404 is retured. The path is RELATIVE TO YOUR APP. So if the url of your app is /foo/bar and the route is /chicken/:someID, then enter "\\/chicken\\/.*". (NOTE: You need to escape backslashes in JSON)

### Front-end Config Value Processing
As strings are not good enough for values such as endpoint URLs or system wide settings we have implemented a processing directive for config values. These directives should precede the actual value to process. The following are available for processing config values:
* `@@LITERAL@@` - this will take the value as a literal value.
* `@@TOKEN@@` - this will run the string through the token service's replace function.

#### Example in action
App module config
```
{
  drupalConfig: { ... },
  frontEndConfig: {
    "normalString": "this is a string",
    "literalString": "@@LITERAL@@window.some_var",
    "tokenString": "@@TOKEN@@[node:title]",
  }
}
```

Config values passed to app
```
window.config = {
  ... default urls ...,
  normalString: "this is a string",
  literalString: window.some_var,
  tokenString: "The node's title"
}
```

> **ESTIMATE** NN

bryanpizzillo:
ODE: ncigovcdode393.prod.acquia-sites.com

Closes #2929 
Closes #2888


bryanpizzillo:
@lburack This now includes #2888 (the breadcrumbs)

lburack:
Tested English and Spanish, to show and hide the root site section in the breadcrumb. Working as expected!

sarinapadilla (approved):

PR #2971, opened 2020-11-19
(2970) Make it possible to preserve environment on code update.

### LINKED ISSUE #2970: Allow ODEs to Redeploy without Reinstalling
### Issue description
Everyone has asked about this before, but we need some long term prototyping with real content.

So we will allow it under some special circumstances:
1. This will only work on redeploying the same branch. So you need to deploy once fresh, then restore your DB, then run drupal update.
2. This will only happen if an environment variable exists. The variable will be `PRESERVE_ON_REDEPLOY` and it must be set to `"true"`.
3. This will need to call drupal update on redeploying.

> **ESTIMATE** TBD

bryanpizzillo:
Closes #2970

### DIFF
--- a/hooks/common/post-code-update/post-code-update.sh
+++ b/hooks/common/post-code-update/post-code-update.sh
@@ -40,51 +40,64 @@ if [[ $target_env =~ ^ode\d* ]]; then
   target_env="ode";
 fi
 
-## Clear last install out of Memcache.
-## Drush CR will not work on a non-existant site. Drush CC will not find the memcached
-## backed bins if the site is not installed. So we made our own to work with empty sites
-## to clean out memcache. (DB backing stores disappear as soon as the DB is dropped.)
-drush cgov:destroy-cache
+# This variable must be the word true for it to update instead of reinstall
+if [[ $PRESERVE_ON_REDEPLOY == "true" ]]; then
+  ## Just do a drupal update and update translations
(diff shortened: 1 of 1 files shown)


hooks/common/post-code-update/post-code-update.sh
```diff
@@ -40,51 +40,64 @@ if [[ $target_env =~ ^ode\d* ]]; then
...
+  ## to clean out memcache. (DB backing stores disappear as soon as the DB is dropped.)
+  drush cgov:destroy-cache
 
-## Reload translation pack.
-blt cgov:locales:translate -D drush.ansi=false
+  ## Perform a fresh install.
+  blt artifact:install:drupal --environment=$target_env -v --no-interaction -D drush.ansi=false
 
-## Setup some default JS globals.
-cat FrontendGlobals.json | drush config:set cgov_core.frontend_globals config_object -
```
blairlearn, line 59: The old code reloaded translations after loading the users.  The equivalent code in the `else` block has reversed those  two steps.

I do not know if that is a significant difference.

blairlearn (commented):
I see a difference in the ordering of steps in the new code, however I do not know that it is significant.

blairlearn (approved):
Looking at the actual files instead of the diff, I see I'm mistaken.  No ordering changes, and the new logic looks good.

PR #2991, opened 2020-12-15
(#2989) Require HTTPS to access the site.

### LINKED ISSUE #2989: Enabler: Enforce HTTPS on non-local environments
# Enforce HTTPS on ODEs and ACSF environments.

The acquia-sites.com and acsitefactory.com domain originally didn't have valid TLS certificates. HTTPS should be enforced on these environments.

blairl-nih:
- Remove 403 for non HTTPS access attempt, simply redirect to HTTPS. This will enforce HTTPS in all environments, regardless of domain.
- Update line numbers for file offsets occuring after .htaccess.https.patch.

Close #2989

### DIFF
--- a/docroot/.htaccess
+++ b/docroot/.htaccess
@@ -115,19 +115,17 @@ AddEncoding gzip svgz
   # uncomment the following line:
   # RewriteBase /
 
+  # Redirect HTTP to HTTPS, but only on Acquia servers.
+  RewriteCond %{ENV:AH_SITE_ENVIRONMENT}  ^(.)+$
+  RewriteCond %{HTTP:X-Forwarded-Proto} !https
+  RewriteRule ^(.*)$ https://%{HTTP_HOST}%{REQUEST_URI} [L,R=301]
+
   # Return Status 403 (Forbidden) on non-ODE server access in the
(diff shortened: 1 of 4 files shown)


docroot/.htaccess
```diff
@@ -115,17 +115,8 @@ AddEncoding gzip svgz
   # uncomment the following line:
   # RewriteBase /
 
-  # Return Status 403 (Forbidden) on non-ODE server access in the
```
bryanpizzillo, line 118: It seems like removing this would allow people to bypass dev,test,int urls such as www-dev-ac.cancer.gov.

bryanpizzillo (commented):
I would like to know the answer to my comment below before requesting changes. This change looks like it will have additional side effects not discussed in this ticket.

docroot/.htaccess
blairlearn: Yeah, I removed too much.  That does indeed allow the lower tiers to be accessed by their native names.  (But it does force HTTPS! 😁)

blairlearn (changes requested):
Keep the enforcement of cancer.gov host names.

PR #3118, opened 2021-07-06
(#3117) Don't care how many times assertValidTokens is called.

### LINKED ISSUE #3117: Tests fail when running locally but not in Pipelines
### Issue description
This is the  mirror of #3094.  Because of the way PHP and Drupal do asserts, the code which runs in "developer mode" (`zend.assertions = 1`) doesn't run in "production mode" (`zend.assertions = 0`).

When we run tests locally, we run in "production mode".  When  they run in pipelines, the tests currently run in "developer mode." Prior to #3094, Acquia had the pipeline running in production mode but then they switched it.

[`ZipCodeLookupResourceTest` asserts that the `assertValidTokens`](https://github.com/NCIOCPL/cgov-digital-platform/blob/develop/docroot/modules/custom/cgov_cts/tests/src/Unit/ZipCodeLookupResourceTest.php#L65-L67) method executes exactly once.  In developer mode, it executes once.  In production mode, it doesn't execute.

So, our choices are to either always run locally in developer mode, and then hope Acquia doesn't randomly change how it works in pipelines again, or else not care how many times it executes.

We're going to do the latter.

> **ESTIMATE** TBD

1.  In the container, run `blt custom:paratest`

- All tests pass.

### What's the actual result?
- `ZipCodeLookupResourceTest::testGetWithMatch` fails because `assertValidTokens` was not executed.

blairl-nih:
Close #3117

### DIFF
--- a/docroot/modules/custom/cgov_cts/tests/src/Unit/ZipCodeLookupResourceTest.php
+++ b/docroot/modules/custom/cgov_cts/tests/src/Unit/ZipCodeLookupResourceTest.php
@@ -62,7 +62,7 @@ class ZipCodeLookupResourceTest extends UnitTestCase {
     // tests for these items. Mocking up the entire cache structure is really
     // problemmatic. So we are going to just have to always hope that url.path
     // is one of the valid Drupal tokens.
-    $this->mockCacheContextsManager->expects($this->exactly(1))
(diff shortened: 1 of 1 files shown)


bryanpizzillo (approved):
This looks good to me. I will leave it up to @blairlearn to determine what order to pull it in.

Issue #3062, opened 2021-05-18
A new site section should not be able to be created at the root


VictoriaSunNIH:
### Issue description
A new site section should never be allowed to be created at the root - as this breaks our cancer.gov homepage. 

We need to block this from happening.

> **ESTIMATE** TBD

### Steps to reproduce the issue

1.  A new site section is created and the landing page was set, with everything checked to hide in navigation. 
2. The parent page (in the Relations drop down) was not selected. 
3. Because there was no URL path,  the site section was added to the root
4. This caused our main homepage (cancer.gov) to not render and show a page not found status.

### What's the expected result?
-We need to make sure that a new site section is not able to be created at the root, and that a relations is set for any new site section that is not the root.

### What's the actual result?
- breakage of cancer.gov

### Additional details / screenshot
- ![Screenshot]()

### Related Tickets
* Issue #9999
* Issue #9999

bryanpizzillo:
Added the constraint `You SHOULD never be allowed to have a term with 2 parents` to the requirements above. Currently, when saving, you could have multiple parents, one of which being Term 0, or indicating the root.

sarinapadilla:
@alinai @VictoriaSunNIH @suryatheja-muppalla I've added the AC document to the ticket description above!

lburack:
I updated the error message in the ACs per comments on the PR (https://github.com/NCIOCPL/cgov-digital-platform/pull/4304#issuecomment-2091163772)

Also this scenario should **NOT** be a thing. We have MANY, MANY site sections that have all navigation display options checked. This would break a ton of content on the enterprise and microsites. This needs to be removed from ACs and not a functional thing. Site sections are complicated and it's imperative that ACs be reviewed with me (or another lead IA) before development. 

**THIS IS BAD! THIS SCENARIO SHOULD NEVER HAVE BEEN CREATED. REMOVE FROM ACS!**
```gherkin
Scenario: User tries to save a new site section with invalid navigation display options
  Given the user has Site Admin permissions
  When a new site section is created 
    And the landing page was set 
    And all three check boxes in “Navigation Display Options” field is checked to hide in navigation 
  When the user clicks 'Save' on the site section
  Then the following error message appears: "The site section cannot be saved. A minimum of one (1) display option must be visible."
```

@andyvanavery31 @sarinapadilla @bryanpizzillo @navsunka @santwuma

andyvanavery31:
[...]

PR #3134, opened 2021-07-15
(#3062) A new site section should not be able to be created at the root

suryatheja-muppalla:
Closes: #3062 

Fixed issue and an error message will be displayed when the user try to save the new site section.


adriancofie (approved):

docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/cgov_site_section.module
```diff
@@ -38,6 +39,43 @@ function cgov_site_section_entity_bundle_field_info(EntityTypeInterface $entity_
...
+ * Implements term_cgov_site_sections_parent_validate(), validate default root as parent.
+ */
+function term_cgov_site_sections_parent_validate(&$form, \Drupal\Core\Form\FormStateInterface $form_state) {
+   $langcode =  $form_state->getValue('langcode');
+   $tid =  $form_state->getValue('tid');
+   $vid = 'cgov_site_sections';
+    $terms =\Drupal::entityTypeManager()->getStorage('taxonomy_term')->loadTree($vid);
+    $enParentTermCount = $esParentTermCount = 0;
+    foreach ($terms as $term) {
+      if($term->depth == 0 && $term->langcode == 'en' && $tid != $term->tid) {
```
bryanpizzillo, line 52: Can you please do this so it is not ONLY Spanish and English? I.E. it should allow for any number of languages, or it should at least allow for those languages specified.

(We had to add Chinese and Portuguese to our old system (and yes, it was never used) and code like this added many hours to that project)

docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/cgov_site_section.module
```diff
@@ -38,6 +39,43 @@ function cgov_site_section_entity_bundle_field_info(EntityTypeInterface $entity_
...
+          $form_state->setErrorByName('parent', t('The site section cannot be saved. Please set the parent site section under the Relations tab.'));
+       }
+
+}
+}
+
+/**
+ * Implements hook_form_alter().
+ */
+function cgov_site_section_form_alter(&$form, FormStateInterface &$form_state, $form_id) {
```
bryanpizzillo, line 73: I initially typed the following, but there may be a better way:
>There should be a `hook_entity_presave` implementation. `hook_form_alter` only covers using the UI. Using the Drupal API, I can create multiple taxonomy terms through the backend API.  We should not be able to do bad things through the backend. 

However, a quick bit of Googling to see how an exception in presave is handled, I see that you can add a Constraint and handle both UI and API.

See 
* https://drupal.stackexchange.com/questions/183685/how-do-i-add-a-custom-validation-handler-to-an-existing-form-field.
* https://www.drupal.org/docs/drupal-apis/entity-api/entity-validation-api/providing-a-custom-validation-constraint#add_constraint_other_entity_base_field

The key take away is that I should not be able to create multiple root site sections ever. Not through the UI, not through the backend.

bryanpizzillo (changes requested):
Request 1 is 100% mandatory - do not hardcode languages. I spelled out the reasons below in the review comment, so I won't repeat here.

Request 2 is 99% mandatory unless there is sample code showing where Drupal does something dumb and creates a term before associating a parent and thus we cannot do it. (It should not be *that* dumb)

Request 3 is not in the comments below and is 80% mandatory - if you can implement the constraint, then a Kernel test should suffice.

docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/cgov_site_section.module
```diff
@@ -38,6 +39,43 @@ function cgov_site_section_entity_bundle_field_info(EntityTypeInterface $entity_
...
+/**
+ * Implements term_cgov_site_sections_parent_validate(), validate default root as parent.
+ */
+function term_cgov_site_sections_parent_validate(&$form, \Drupal\Core\Form\FormStateInterface $form_state) {
+   $langcode =  $form_state->getValue('langcode');
+   $tid =  $form_state->getValue('tid');
+   $vid = 'cgov_site_sections';
+    $terms =\Drupal::entityTypeManager()->getStorage('taxonomy_term')->loadTree($vid);
+    $enParentTermCount = $esParentTermCount = 0;
+    foreach ($terms as $term) {
```
adriancofie, line 51: Please keep an eye on performance and code efficiency when making the updates @bryanpizzillo mentioned, particularly when using the presave() hook. 

presave() is used recursively when updating certain cgov_site_section fields and loadTree() is expensive in certain circumstances --- we have A LOT of site sections in production so this is potentially nightmare fuel.

bryanpizzillo (changes requested):
`$entity->save()` does not call `$entity->validate()`, which is what triggers the SectionParentTermConstraint. This means that it is possible to use the API to create a bad site section.

I see in the docs that [Term might have](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21ContentEntityBase.php/property/ContentEntityBase%3A%3Avalidated/8.6.x) a `->validated` property.

So one approach might be to check:
```
if (!$entity->validated) {
  $validation_errors = $entity->validate();
  if (count($validation_errors) !== 0) {
    // This is pseudo code here.
    throw new Error($validation_errors[0]);
  }
}
```

Once this is added, there must also be a test for the presave hook.

docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/cgov_site_section.module
```diff
@@ -171,10 +172,18 @@ function cgov_site_section_taxonomy_term_update(EntityInterface $entity) {
   if ($entity->parent->target_id == 0 && $entity->langcode->value === 'en') {
     _cgov_site_section_attempt_set_front_page($entity);
   }
-
   // Save the term children if this term update meets the criteria.
   if (will_trigger_save_children($entity)) {
-    save_term_children($entity);
+    // Checking for any validation related errors.
```
bryanpizzillo, line 177: Why are you calling validate here? Also, why are you calling validate twice?

bryanpizzillo: Also this is not being done in presave. So new terms completely bypass this, or at least should.

smitahantziosps:
New PR for this is here -https://github.com/NCIOCPL/cgov-digital-platform/pull/3631

PR #3163, opened 2021-09-13
Change of Identity Providers

### LINKED ISSUE #3162: Enabler: Make the  SAML Identity Provider configurable
# The SAML Identity Provider (IDP) should be configurable.

Presently, the IDP is selected from a hard-coded list of providers in `simplesamlphp/config/authsources.php`. An environment-specific provider should instead be specified in the `SamlSecrets` class in `/mnt/files/${AH_SITE_GROUP}.${AH_SITE_ENVIRONMENT}/secrets.settings.php`

> **ESTIMATE** 1 day

blairl-nih:
Closes #3162


blairlearn (changes requested):
This is not the right approach. In the the process of setting up a second site to use the Okta identity provider, it was discovered that Okta uses different metadata, and a different IDP url, for each application.

blairl-nih:
This wasn't the right approach.

PR #3169, opened 2021-09-16
(#2102)As a content editor I would like to know if my URL will be lon…

### LINKED ISSUE #2102: As a content editor I would like to know if my URL will be longer than 255 characters.
When creating a site section, or a page, we enter content into field_pretty_url. This then gets appended to the parent sections path, and a URL alias is created. When editing a page (or section) it would be good to know the URL being created is longer than 255 characters and would be truncated.

(Other thoughts, maybe warn at 200 characters. Maybe error on 255 characters...)

The alias limit was increased from 100 to 255 as part of #2100 when we first noticed the truncation issue.

suryatheja-muppalla:
…ger than 255 char

closes #2102 

fixed the change requested to warn for 200 char limit.

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.libraries.yml
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.libraries.yml
@@ -1,0 +1,3 @@
+cgovcorejs:
+  js:
+    js/cgov-core.js: {}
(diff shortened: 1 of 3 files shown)


alinai:
@suryatheja-muppalla , @VictoriaSunNIH ,  I created an article
https://ncigovcdode434.prod.acquia-sites.com/test/test-and-test-test-and-test-test-and-test-tes[…]-test-test-and-test-test-and-test-test-and-test-te and then published it. It seems currently the pretty url gets truncated at 100 characters in the the pretty url field. So I was unable to see any message when character count crosses 200 in this field.

ramu-nih:
This id added in https://github.com/NCIOCPL/cgov-digital-platform/pull/3504

PR #3173, opened 2021-09-27
(#3121) Create new Adobe Launch module

### LINKED ISSUE #3121: Module Blocker: Adobe DTM
### Issue description
The `drupal/adobe_dtm` module does not support 9.x, and it was a [PITA to point to the dev branch](https://github.com/NCIOCPL/cgov-digital-platform/blob/develop/composer.json#L29). So this is not going to be a viable long term solution.

There is an `adobe_launch` module that now supports 9.x, but it is minimally maintained and on an RC version. (Although at least it has valid configuration schemas)

However, both of these modules have such limited usage that there must be some other way people are injecting the launch JS. (Which is not JUST importing a JS file.)

At this point, we should probably just roll our own version with our minimal requirements. (What is the launch URL, and should it be loaded async or not, and clear the drupal caches on update so changes reflect everywhere...)

> **ESTIMATE** TBD

bryanpizzillo:
ODE: ncigovcdode432.prod.acquia-sites.com

- This is to remove adobe_dtm, which does not support Drupal 9.
- This borrows from some of the logic, what is necessary for our requirements.
- We addeed a config:cgov_adobe.settings cache tag to the build array, so updating the analytics config will invalidate all pages and the Drupal cache. I.E. No longer a need to drush cr and the like when the config is updated. This did not exist previously.
- This also enables cgov_adobe, copies the launch URL from the adobe_dtm config, and disables adobe_dtm.  
- THIS DOES REQUIRE AN SUBSEQUENT RELEASE TO REMOVE THE adobe_dtm MODULE FROM COMPOSER!

Closes #3121

### DIFF
--- a/docroot/modules/custom/cgov_adobe/cgov_adobe.info.yml
+++ b/docroot/modules/custom/cgov_adobe/cgov_adobe.info.yml
@@ -1,0 +1,7 @@
+name: 'Application Module'
+type: module
+package: 'Application Support'
+description: 'Support for embedding applications on nodes and other routable entities'
+core: 8.x
+# dependencies:
+#  - drupal:toolbar
--- a/docroot/modules/custom/cgov_adobe/cgov_adobe.links.menu.yml
+++ b/docroot/modules/custom/cgov_adobe/cgov_adobe.links.menu.yml
@@ -1,0 +1,5 @@
+cgov_adobe.settings_form:
+  title: 'Cgov Adobe'
+  description: 'Configure the integration with Adobe Launch and the web site.'
+  route_name: cgov_adobe.settings_form
+  parent: system.admin_config_system
--- a/docroot/modules/custom/cgov_adobe/cgov_adobe.module
+++ b/docroot/modules/custom/cgov_adobe/cgov_adobe.module
@@ -1,0 +1,164 @@
+<?php
+
+/**
+ * @file
+ * Contains cgov_adobe.module.
+ *
+ * This is heavily borrowed from drupal/adobe_dtm, which does not support
+ * D9, and has a lot of extra features we do not need. So we needed to
+ * implement tag management for our needs in order to get to D9.
+ *
+ * NOTE: We do not support async, and will only implement it when we have
+ * the need for it. (Since we can't really test async right now)
+ */
+
+use Drupal\Core\Routing\RouteMatchInterface;
+
+/**
+ * Implements hook_help().
+ */
+function cgov_adobe_help($route_name, RouteMatchInterface $route_match) {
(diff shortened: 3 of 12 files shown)


bryanpizzillo:
## Test Notes
### Automated tests
Browser Tests were created for:
1. exclude paths working correctly. We can exclude a page by specifying its alias or route; once set, visiting a page by its alias or route will be excluded. YOU DO NOT NEED TO SPECIFY BOTH. The test checks this.
   * If no paths are excluded everything works. 
2. Tests to ensure the enable settings works (setting enable to false disables the page tags everywhere)
3. The config settings page is tested, in that you can save the config correctly if you post the correct data. It assumes the UI works properly as all Drupal browser tests usually assume.

We test that the URL and the page bottom tag appear on all pages where it is allowed.

### Manual tests
* I change the launch URL and made sure it worked.
* I tested the paths
 
### Upgrade Test
I tested this locally following these steps.
1. `blt cgov:reinstall --no-interaction`
2. `blt sql-drop --no-interaction`
3. `drush sql-cli < ./files-private/LAST_NIGHTS_BACKUP.sql`
4. `blt drupal:update`
5. `drush pml | grep adobe` -- Validate the cgov_adobe module was turned on and adobe_dtm was turned off.
     ```
     Statistics                             Adobe DTM (adobe_dtm)                                             Disabled                   
     Application Support                    Application Module (cgov_adobe)                                   Enabled                 
     ```
6. `drush cget` (selecting cgov_adobe.settings, or in my case number 33) -- Validate the launch URL was copied from the adobe_dtm config to the cgov_adobe config.
     ```
     enabled: true
     exclude_paths:
         - '/admin*'
         - '/batch*'
         - '/node/add*'
         - '/node/*/edit'
         - '/node/*/delete'
         - '/node/*/revisions*'
         - '/user/*/edit*'
         - '/user/*/cancel*'
     launch_property_build_url: //assets.adobedtm.com/6a4249cd0a2c/949e98979342/launch-5f2a5357bf66.min.js
     _core:
         default_config_hash: kGdGScFqH_-dpwxneiSwmhd-JFJ9eBt145vV8z8IjIY
     ```

alinai:
@KateMashkinaNIH , Java Regression Tests passed on ODE432.

SupriyaSaxena0605:
@KateMashkinaNIH  All the automated Acceptance tests passed on ODE432

kate-mashkina:
Tested UI - disabling Adobe Launch Tags works, excluding path works as well. I did not have to clear cache and/or varnish - everything was taken care of by clicking 'Save Configuration' button!

PR #3183, opened 2021-09-30
(#3180) Remove Adobe DTM Module

### LINKED ISSUE #3180: Enabler: Remove adobe_dtm from codebase
# Remove adobe_dtm module from codebase
> **ESTIMATE** TBD

The adobe_dtm module had issues with Drupal 9 and was replaced in #3121. Part of that ticket was "uninstalling" the module from our platform sites. We cannot remove the module at the same time, so this is an addon ticket to remove it from composer and our codebase. This should be done before the composer 2 upgrade as this module required some hacky repository settings as well in our composer.json that I would rather not deal with in the composer upgrade.

bryanpizzillo:
This removes the module from the codebase -- it is assumed this will be deployed immediately after release 1.2.14.

Closes #3180

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,19 +25,6 @@
         "clinical-trial-search": {
             "type": "vcs",
             "url": "https://github.com/NCIOCPL/clinical-trials-search-client.php"
-        },
-        "bad-adobe-module": {
-            "type": "package",
-            "package": {
-                "name": "drupal/adobe_dtm",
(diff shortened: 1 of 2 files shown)


bryanpizzillo:
Merging this PR as it has been tested locally more than once, and we really need to test this in ACSF. So we will pull it into the 1.2.15 release branch. If it does not work, then we create a new release branch, cause this is the only thing for that branch.

PR #3205, opened 2021-10-06
(#2782) Add fingerprinting to generated svg sprite

### LINKED ISSUE #2782: The NCI Sprite does not have fingerprinting
### Issue description
The NCI Sprite does not have any fingerprinting and as such if we update the sprite, then user's browsers keep it cached, and we end up with a lot of questions as to why it is not updated.

> **ESTIMATE** 20

N/A (We don't want to reproduce it... since you must break the world)

- Any reference in the CSS to `../images/sprites/svg-sprite.svg` should be `../images/sprites/svg-sprite.svg?v=<FINGERPRINT>` where FINGERPRINT is a unique identifier of that version of the sprite.

arcepaul:
Closes #2782 .

Added fingerprinting to svg using webpack and 8 digit hash.


bryanpizzillo:
This was fixed in a different way. Closing this PR

PR #3234, opened 2021-11-05
(#3227) Revert purge module to 3.0

### LINKED ISSUE #3227: "Something" is not clearing out the cache correctly.
Since going live we have had 2 issues reported where a change to the content on the back end was not reflected on the web site.
1. A Drug Information Summary was deleted via a hotfix, and after the delete occurred it was still appearing on the front-end.
2. A Spanish translation of a blog post was created and published. The English post did not show the toggle after the Spanish version was published.

In both cases the following steps triggered the right content to appear:
1. `drush pinv url <URL_TO_BE_CLEARED>` 
2. A URL purge through Akamai

#### NOTES:
* We do not know yet if this happens for all content, or just happens sometimes.
* When the issue occurred with the DIS, I was able to see the 404 when going directly to the Origin server, indicating the Drupal cache was correctly updated.
* When the issue occurred with the DIS, it was not enough to just clear the Akamai cache to get it to fully stick.
   * `drush pinv url <URL>` indicates that the Varnish cache was not getting cleared. (You can clear Akamai all you want, but if the varnish cache is still holding onto the stale version, then it will show the stale version.)
   * I did see at one point the page gave me a 404 through Akamai as if it worked, but others still saw the page. Ultimately a couple of hours later when others said that only clearing Akamai did not work, I once again saw the page and not a 404.
     * There may be some funny business because we have 2 web servers. There are also 2 nginx/varnish servers.

That is for the person doing this to figure out. What are the exact steps to reproduce.

When content changes are published on the back end (www-cms), they should show up on the web site (www).

### What's the actual result?
They are not appearing correctly

bryanpizzillo:
There is an issue with a change in 3.2 regarding how they determine if a page is cacheable or not. This is causing issues.

Closes #3227

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -71,7 +71,7 @@
         "drupal/paragraphs": "^1.6",
         "drupal/paragraphs_asymmetric_translation_widgets": "^1.0",
         "drupal/pathauto": "^1.6",
-        "drupal/purge": "^3.2.0",
+        "drupal/purge": "3.0.0",
         "drupal/redirect": "^1.6",
         "drupal/role_delegation": "^1.0",
         "drupal/seckit": "^2.0",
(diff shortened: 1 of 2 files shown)


KateMashkinaNIH:
Verified cancergov, dceg and nano on ACSF by edited every content type, added translation where possible- no issues to report
ODE445 passed automated regression tests

Issue #3245, opened 2021-11-16
Spike: Prototype removing Sticky Nav


bryanpizzillo:
## In order to for our team to feel more comfortable with removing the sticky nav ahead of the rest of the updates, developer needs to work on a prototype that does not have sticky nav.
> **Estimate** *# point* 

The new NCIDS will no longer make the navigation "sticky", as such the implementation of the NCIDS navigation on WWW will need to disable the Sticky Nav and any related logic.

The Sticky Nav is a very complicated bit of code that also affects many other features of the site given how it breaks deep linking. There are many elements within the site that must also detect the scroll position to ensure they do not "crash" into the sticky nav. (The nav is taller on an upward scroll than it is when scrolling downward) 

It will be easier to manage the new NCDS Transitional Theme and the current theme side by side with the sticky nav removed. This is because the new transitional theme will be a lightweight copy of the current theme in which we will just be referencing the current JS & CSS for most of the features. At the point we remove the sticky nav, and if the current site still uses it, we will have to duplicate all the JS & CSS that references or handles sticky nav offsets into the transitional theme. At this point we will have to maintain two sets of these files. Examples of code that "know" about the sticky nav are Deep Linking, however there are probably many more styles that account for the spacing.

We have been requested to prepare a prototype of the existing site so we can see what the site would look like without the sticky  nav, as well as be able to demo it to other microsite owners. The idea being that the sticky nav will be removed from all sites when they are updated, so can we just accelerate this removal to reduce maintenance costs.

### Expected Deliverables:
* A working prototype to demo.

bryanpizzillo:
## Testing Scenarios
These scenarios are documenting what currently happens on production, not what makes sense or what "works", but the state of things now. These links are based on "ODE" content. For the purposes of this scenario, "top of the window" means right below the sticky nav on current codebase and the top of the window on the prototype.

### Enter top-level section URL directly
#### Desktop
1. Open a browser and enter `/types/breast/patient/breast-treatment-pdq#_185`
    * This is the "Treatment Option Overview" section 
**Result:** You should be taken to the "Treatment Option Overview" section and it should be at the top of the window.

#### Mobile
1. Open a browser and enter `/types/breast/patient/breast-treatment-pdq#_185`
    * This is the "Treatment Option Overview" section 
**Result:** You should be taken to the "Treatment Option Overview" heading in the accordion, with the contents collapsed.

### Enter deep section URL directly
#### Desktop
[...]

PR #3274, opened 2022-01-05
Remove the Sticky Nav

### LINKED ISSUE #3239: Enabler: Remove Sticky Nav Cruft and Anchor Scrolling
# Remove Sticky Nav Cruft and Anchor Scrolling
> **ESTIMATE** TBD

So this ticket is not going to be anywhere as easy as just doing it. The current front-end uses Headroom to manage the sticky navigation when scrolling a page. Since the navigation appears at the top of the page, this breaks deep linking to anchors. So there are additional hacks to scroll the page down to ensure the anchor appears under the nav. There is a lot of effort to handle deep linking, but not all is related to sticky nav, some is for dealing with accordions. (So be careful) Additionally there are other elements where we need to make sure they to not crash into the sticky navigation, so there hacks around that.

A quick [search](https://github.com/NCIOCPL/cgov-digital-platform/search?p=1&q=headroom) shows a number of elements for headroom. There may be more that do not reference headroom itself, but something else.

This was prototyped by in https://github.com/NCIOCPL/cgov-digital-platform/pull/3263. There were a number of underlying bugs identified while prototyping, so it should not be just copied wholesale.

### Acceptance Criteria
Various criteria were documented on #3245

### LINKED ISSUE #3272: Content types using the pdq.js entry point may throw Javascript errors under certain circumstances.
PDQ Cancer Information Summaries throw a Javascript error when the [pdq.js](https://github.com/NCIOCPL/cgov-digital-platform/blob/develop/docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/src/entrypoints/pdq/PDQ.js#L78-L92) code attempts to initialize pronunciations. This probably also occurs for any Drug Information Summary without a pronunciation. This all comes down to no checking to see if elements exist before manipulating them.

Console Error Attached:
image

### LINKED ISSUE #3273: Urls with anchor fragments break NCIAnalytics.buildPageDetail
Just visit a PDQ summary page with an anchor fragment such as https://www.cancer.gov/types/colorectal/patient/colon-treatment-pdq#_135. This will cause the error. The root cause is an [undefined variable](https://github.com/NCIOCPL/cgov-digital-platform/blob/develop/docroot/profiles/custom/cgov_site/themes/custom/cgov/src/libraries/analytics/nci-analytics-functions.js#L1890).

However, after additional investigation, defining the variable is not enough as the CSS selector needs to also have any special characters escaped. (Since this code probably blew up anyway, we never would have gotten to that next case)

This does lead us to the question that if this code never worked, why do we need to fix it? 

image

### LINKED ISSUE #3282: Enabler: Remove cgov_common theme JS building from cgov folder
# Remove cgov_common theme JS building from cgov folder
> **ESTIMATE** TBD

In order to more cleanly implement #3237 we should just move out any `cgov_common` building logic from the `cgov` folder. This affects things like package.json, webpack configs, etc. When all is said and done `cgov` should just contain:
* `cgov_common`
* `src`
* `static`

bryanpizzillo:
This ticket fixes a few bugs that were interfering with removing the Sticky Nav, as well as removing the sticky nav itself.

ODE: http://ncigovcdode448.prod.acquia-sites.com/AH_VIEW
Test Cases: See #3245

Closes #3272
Closes #3273 
Closes #3239 
Closes #3282


VictoriaSunNIH:
This looks good to the Product team!

PR #3289, opened 2022-01-26
(#3288) Hack to fix legacy links in PDQ

### LINKED ISSUE #3288: Some PDQ links do not work on Desktop
A bug was introduced with the sticky nav to prevents some legacy PDQ anchors (`#link/_123`) from working correctly. These anchors do not appear in the actual content, so the browser cannot handle it as a normal anchor. There have been reports that these links *sometimes* work when pasted into a browser. The quickest fix for this is add a click handler on these links to strip the `#link/` part off the URL.

An example can be found at https://www-dev-ac.cancer.gov/types/breast/hp/breast-ovarian-genetics-pdq
Look for the link for "Moderate-Penetrance Genes", it will have an href of "#link/_88".

The more permanent fix its to remove these bad links from the source.

NOTE: This needs to be retested on mobile to ensure we don't break anything.

bryanpizzillo:
This fix rewrites (via JS) any #link/_123 style links in content.
This hack should be removed once the PDQ content is no longer publishing anchors in this old format.

ODE: http://ncigovcdode453.prod.acquia-sites.com/AH_VIEW

Closes #3288

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_yaml_content/content/911_cct_chatbot_pages.content.yml
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_yaml_content/content/911_cct_chatbot_pages.content.yml
@@ -1,0 +1,112 @@
+## This is mock content for allowing the CBIIT team to test their CCT chatbot
+## Root URL for Testing (Very long)
+- entity: "node"
+  type: "cgov_article"
+  langcode: en
+  status: 1
+  moderation_state:
+    value: 'published'
+  title: "Funding for Cancer Training"
+  field_browser_title:
+    value: "Funding for Cancer Training"
+  field_page_description:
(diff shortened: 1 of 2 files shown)


kate-mashkina:
https://lion.box.com/s/w5s8xepx3vla0syfhf4y6avd42sksss0 - box note with the ODE results 

scenario Enter **deep section URL directly, but have scrolled** is weird
Nadia and myself saw one behavior (which actually matched Victoria's test notes from https://github.com/NCIOCPL/cgov-digital-platform/issues/3245) , where as Supriya reports another (which matched results from test- acsf )
Noting it here as potential result of a  race condition even though both behaviors are acceptable and not a blocker( tagging @VictoriaSunNIH )

arcepaul (approved):
LGTM - confirmed that the links are modified by the script as expected.

PR #3319, opened 2022-03-15
Pin Highcharts to major version 9

arcepaul:
Same as blair's changes but leaving mapdata to pull latest


blairlearn:
Additional work required.  See #3329

Issue #3314, opened 2022-03-08
Create raw HTML block for information sessions


bennettcc:
Please create a raw HTML block to be used for information session on the SBIR custom homepage.  The addition/management of text and imagery should be done through this custom block (editable within the CMS)

SBIR Homepage: https://publicis.invisionapp.com/share/D512EWYA3GTE#/screens

acsnedegar:
This looks good to me on dev-acsf and cms view

Issue #3311, opened 2022-03-08
Update the full wide card within the cancer.gov template


bennettcc:
1) Change the Cool theme button color, background color, and text hoverstate for use within cancer.gov on full wdith borderless cards:

Accent button: #123E57
Button Hover: #1D567E
Background color: #2A72A5 (as is)
Text hoverstate color: stay white and get an underline

2) Change the Warm theme button color, background color, and text hoverstate for use within cancer.gov on full width borderless cards:

Accent button: #123E57
Button Hover: #1D567E
Background color: #FFFFFF
Text hoverstate color: #BB0E3D (red as is)

SBIR Homepage Comps: https://publicis.invisionapp.com/share/D512EWYA3GTE#/screens

acsnedegar:
This passes for me, thanks!

alinai:
Looks good to me on ODE 446. Thanks @erikscottmeyer and @acsnedegar!

alinai:
Looks good to me

alinai:
Looks good to me on dev-acsf.

acsnedegar:
Also looks good to me there

PR #3328, opened 2022-03-24
Tickets/sbir custom homepage

erikscottmeyer:
Added commits for tickets #3311 and #3314 regarding the SBIR Custom Homepage.


blairlearn:
Did not result in anything which could be deployed.

PR #3332, opened 2022-03-25
Enable Dart Sass

### LINKED ISSUE #3293: Fix hardcoded uses of tokens in CSS
There are hard coded instances of tokens that will make it impossible to re-use sass modules from the current theme. For example we have:
* `$montserrat-font-stack` which is hard coded in the `_profile-panel.scss` partial as `font-family: Montserrat, Avant Garde, Arial, sans-serif;`. This should be `font-family: $montserrat-font-stack;`
   Instances:
   * `docroot/profiles/custom/cgov_site/themes/custom/cgov/src/styles/_tableSort.scss`
   * `docroot/profiles/custom/cgov_site/themes/custom/cgov/src/styles/_profile-panel.scss`
* ... more to come.

### LINKED ISSUE #3296: Refactor mixins and functions out of sass partials that define styles
### Issue description

> **ESTIMATE** 3

On of the approaches we are planning to take to mix USWDS and NVCG CancerGov is to wrap legacy content in a `<div class="legacy-content">` element. We can then take all of our inner content styles and wrap the SASS includes with `.legacy-content { @import ... }` directives. The issue is that when a mixin or function is defined in one of these style classes AND that mixin/function is used outside of that partial, the mixin/function cannot be found. This ticket is for ensuring that there is a clear delineation such that we can include only what is necessary.

### LINKED ISSUE #3307: We should be using Dart Sass (sass) instead of Node Sass (node-sass)
Currently we are using node-sass for our Sass compiler for the front-end. The ncids will be moving to Dart Sass as part of https://github.com/NCIOCPL/ncids/issues/353. We should also do it because node-sass is deprecated.

### LINKED ISSUE #3326: CTHP Guide Card View More Info has Contrast Issues
While investigating an issue with nested SASS and the View more information list, we discovered some other issues with the display. 

**Non-hovering Display**
image

**(BAD) Hovering over an External Link with exit disclaimer**
image

**Hovering over a file link text**
image

**(BAD) Hovering over a file link outside of link text**
image

### LINKED ISSUE #3327: Remove arrow-link-white style from code
In our code, "arrow-link-white" style was once used for the "Resources For" cards that showed on home and landing pages.  We no longer use this style, and all cards are white, using the "arrow-link" style. 

Additionally, in some sort of bizzaro universe, the `guide-card--home` class also references it with a comment `Class previously used to target homepage only with blue tint on 'Resources for'`. This class is still on the site -- but the arrow-link-white references need to be removed.

We should cleanup the code and remove arrow-link-white from our code.

image

### LINKED ISSUE #3357: Common.scss - relocate "non-common" rules and partials
Some rules, hacks and partials have been added to Common.scss leading to unnecessary bloat.  As a developer, please move these rules out of common.scss and into their respective content-type stylesheets so that Common.scss is trimmed down to only common styles across pages of the website.

> **ESTIMATE** 2

### Notes

* There is a glut of hacks at the bottom of Common.scss that can be moved to other stylesheets such as ones for CTHP, blog posts, and homelanding pages
* Things like special-report, alternating lists, borderless cards can be moved out and into the content type stylesheets that they more specifically belong to

bryanpizzillo:
ODE: https://ncigovcdode459.prod.acquia-sites.com/

NOTE: I broke out 3307 so that it would be easier to review. Switching to dart sass is the last commit, but there were a bunch of issues that needed to be fixed prior to that working.

- Lots of changes to sprites because our sprites mixin was bad and breaking dart sass.
- Moved to using Symbols for sprite backgrounds instead of offsets in the sprite file. 
- Also fixed all the deprecated division warnings that Dart Sass is now reporting.

Closes #3307
Closes #3326
Closes #3327 
Closes #3293 
Closes #3296 
Closes #3357


bryanpizzillo:
I cannot approve a PR I "own" so I will leave the approval as a comment. This has passed product testing; any major issues have been resolves; there may be minor issues, but QA has tested this to the best of their ability given the scale and have not found anything else major. So pulling this into develop. (and there was much rejoicing)

PR #3352, opened 2022-04-08
(#3351) Supply theme in Yaml blocks

### LINKED ISSUE #3351: Enabler: Allow Yaml Content to Add Blocks to Any Theme's Region
# Allow Yaml Content to Add Blocks to Any Theme's Region
> **ESTIMATE** TBD

We have extended the `yaml_content` module to allow `block_content` custom blocks to be added to a theme's region. This is done by adding a `region__CONFIG` property in the YAML. The value of the property is set as `<region_id>`. Currently this only allows blocks to be added to a region of the active theme. For the initial development of the redesign theme we need to be able to load up the legacy blocks into the `cgov_common` theme, and then load the new NCIDS blocks into the `ncids_trans` theme. We should be able to add blocks to the regions of any theme by supplying the region as `<theme_id>:<region_id>`. This is in addition to using the current `<region_id>` value that uses the active theme.

bryanpizzillo:
ODE: https://ncigovcdode463.prod.acquia-sites.com/
Closes #3351

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_yaml_content/content/000_mobile_menu.content.yml
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_yaml_content/content/000_mobile_menu.content.yml
@@ -6,7 +6,7 @@
   langcode: en
   status: 1
   region__CONFIG:
-    value: "mobile_nav"
+    value: "cgov_common:mobile_nav"
   field_raw_html:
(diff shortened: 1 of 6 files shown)


bryanpizzillo:
This passed regression testing.

blairlearn (approved):
Mobile Nav, Search Bar, and the footer all appear to be populated for both English and Spanish.

PR #3356, opened 2022-04-12
(#3353) Update dmore/behat-chrome-extension and dmore/chrome-mink-driver

### LINKED ISSUE #3353: Build breaks due to Behat changes
### Issue description
The `DMore/behat-chrome-extension` package has been renamed to `behat-chrome/chrome-mink-driver`. This is part of our dependencies for behat.

Something in the transition causes `composer install` to stop and prompt for GitLab credentials. When none are provided (because we don't use GitLab), the install fails.

> **ESTIMATE** TBD

1.  `composer cgov-clean && composer install`

- All packages should install.

### What's the actual result?
- A password prompt appears.

```
Could not fetch https://gitlab.com/api/v4/projects/DMore%2Fchrome-mink-driver/repository/archive.zip?sha=f85c8f86ca2e9000119c310577a6942683f7e280, enter your gitlab.com credentials to access private repos
```

blairl-nih:
Newer versions, created to resolve the password prompt.

Better, cleaner fix than in #3354

Fixes #3353

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,6 +40,8 @@
         "brianium/paratest": "^6.3",
         "composer/installers": "^1.9.0",
         "cweagans/composer-patches": "^1.7",
+        "dmore/behat-chrome-extension": "^1.4",
+        "dmore/chrome-mink-driver": "^2.8",
         "drupal/acquia_connector": "^3.0.4",
         "drupal/acquia_purge": "^1.0@beta",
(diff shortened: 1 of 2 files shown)


bryanpizzillo (approved):
Ok, I like this a lot more that we have pinned compatible versions with what is in blt-behat.

Issue #3382, opened 2022-05-19
HSTS not set in Drupal


blairlearn:
### Issue description
The strict-transport-security header is missing in responses from the www-*-ac sites.

> **ESTIMATE** TBD

### Steps to reproduce the issue

1. Open the browser
2. Open the F12 tools
3. Switch to the network tab
4. Navigate to www-dev-ac.cancer.gov (or test-ac, int-ac is behind Akamai)
5. View the response headers for the page. 

### What's the expected result?
- The `strict-transport-security` header should be set to a value of `max-age=31557600; preload`

### What's the actual result?
- The `strict-transport-security` header is not present.

blairlearn:
This appears to be across-the-boards.  Running `drush cget seckit.settings seckit_ssl` in any environment, shows the seckit defaults.

This may be reproduced locally by editing `blt.yml` to load `cgov_https_config` for local dev.

blairlearn:
This appears to be due to `docroot/profiles/custom/cgov_site/modules/custom/cgov_https_config/config/install/seckit.settings.yml` being incorrectly named as `seckit.settings.settings.yml` and thus not loading.

blairlearn:
Closed in 910ba0895aaf1ea40b9bb328d5d6f91dc911a33d

PR #3391, opened 2022-05-23
(#3382) Fix seckit/HSTS

blairl-nih:
Correct name of seckit.settings.yml.


blairlearn:
Fix verified on www-dev-ac with SSL Labs

image

bryanpizzillo (approved):

PR #3424, opened 2022-06-28
(#1700) Change session timeout.

### LINKED ISSUE #1700: Configure inactivity timeout to 120 minutes
CGDP CMS instances need an activity timeout. Currently this is the default Drupal timeout of around 2 days. The security POAM says we should set a timeout at 60 minutes but the Okta timeout is at 2 hours. Setting less than 2 hours results in the user being logged out but then immediately logged back in again, which is an awkward user experience. We are currently planning to set the timeout to 120 minutes to align with Okta.

**Tasks**

- [x] Install and configure [autologout](https://www.drupal.org/project/autologout) module
- [x] Set autologout to only be available on SAML environments
- [ ] Verify timeouts work 

**Testing**

_Scenario 1_

- Log in to environment via SSO and go to a CMS admin page (i.e. all content)
- Set a timer for two hours
- Don't do anything else in the browser during those two hours
- Expected result: You are logged out and redirected to `/`
    - Note, if your computer goes to sleep during that two hour period you may not be redirected. But if you try to navigate to another CMS page you should find yourself logged out

_Scenario 2_

- Log in to environment via SSO and go to a CMS admin page (i.e. all content)
- Set a timer for 1 hour
- At 1 hour navigate to another CMS admin page
- Set a timer for 1 hour
- After another hour (2 hours total) navigate to another CMS admin page
- Expected result: You are still logged in and able to navigate because your session was extended

**Original from @VictoriaSunNIH:** 

> Security requirements require that the system will log a user out if they are inactive for 60 minutes.  Please confirm and configure this setting.

blairl-nih:
Closes #1700

### DIFF
--- a/simplesamlphp/config/acquia_config.php
+++ b/simplesamlphp/config/acquia_config.php
@@ -125,3 +125,9 @@ elseif (getenv('AH_SITE_ENVIRONMENT')) {
   $config['store.sql.password'] = $creds['pass'];
   $config['store.sql.prefix'] = 'simplesaml';
 }
+
+/**
+ * Set Session Timeout.
+ */
+//$config['session.duration'] = 1 * (60 * 60); // 1 hours.
+$config['session.duration'] = (5 * 60); // 5 minutes
--- a/simplesamlphp/config/acquia_config.php
+++ b/simplesamlphp/config/acquia_config.php
@@ -130,4 +130,6 @@ elseif (getenv('AH_SITE_ENVIRONMENT')) {
  * Set Session Timeout.
  */
 //$config['session.duration'] = 1 * (60 * 60); // 1 hours.
-$config['session.duration'] = (5 * 60); // 5 minutes
+//$config['session.duration'] = (5 * 60); // 5 minutes
+// we try  'session.state.timeout' next
+


jfrank-nih:
So @blairl-nih, the default appears to be... nothing? Nothing happened to my session at 4 hours. After 8 hours I was logged out because `session.duration` kicked in.

That said, I think we just got too focused on the word "timeout" when we were looking for "lifetime".

```
 /*
     * Expiration time for the session cookie, in seconds.
     *
     * Defaults to 0, which means that the cookie expires when the browser is closed.
     *
     * Example:
     *  'session.cookie.lifetime' => 30*60,
     */
    'session.cookie.lifetime' => 0,
```

So I think what we want is `session.cookie.lifetime = (60 * 60) // 1 hour` and we leave `session.duration` alone. **_Unfortunately_**, if we set the cookie lifetime I think it will preserve across browser sessions, and we don't want that. And JavaScript cookies don't really let you do "1 hour or browser close".

Looking through the simplesamlphp code I have a _guess_ at what's going on.

- If the browser is still open then the cookie is still alive (`session.cookie.lifetime`)
- If we are within our valid cookie duration (`session.duration`) then our session persists
- The auth state (`session.state.timeout`) might expire but if within the session duration, that doesn't expire the session, just the state ([saveState](https://github.com/simplesamlphp/simplesamlphp/blob/131c640e5343c62d01038633504710445be3501a/src/SimpleSAML/Auth/State.php#L203) goes to [setData](https://github.com/simplesamlphp/simplesamlphp/blob/131c640e5343c62d01038633504710445be3501a/src/SimpleSAML/Session.php#L881) which will set or extend the timeout for the state)

Without actually debugging the values over time to see what's going on, that's my guess. 

If this is the case and we can't really do what we want here in simplesaml then we could interact with it from drupal ([like we do here](https://github.com/NCIOCPL/cgov-digital-platform/blob/develop/docroot/profiles/custom/cgov_site/modules/custom/cgov_saml_auth_config/cgov_saml_auth_config.module)). Not quite sure how this would work but it would be something along the lines of looking into simplesaml from Drupal to see if the auth state had expired and, if so, forcing a Drupal logout or killing the simplesaml session.

jfrank-nih:
simplesamlphp recommended means of doing a session check function: https://github.com/simplesamlphp/simplesamlphp/blob/master/docs/simplesamlphp-advancedfeatures.md#session-checking-function

blairlearn:
@jfrank-nih The cookie behavior is pretty much what I'd expect. You can have a cookie until browser close, or until a fixed time; I'm unaware of a mechanism for the lesser of the two.

But, my question is, what happens when the session/auth state expires? If Drupal no longer knows the state, it seems to me that it would have to force you to login in again, which is the desired behavior.

I think finding out what happens then is the next thing to check.

jfrank-nih:
That's my point--nothing seems to happen. I left a browser window open for four hours and tried going to a page while watching the network requests. I was not logged out and not passed to Okta--just went to the page as normal. Then when I hit eight hours I was logged out as expected by the `session.duration`. So... unless you have a better way to check that I think the answer is debug it if you want to know.

As to the cookies, yes, it is expected behavior. You could get around it by setting _two_ cookies--one a session cookie and one a timed cookie. Then verifying that _both_ cookies were still around. But that would be a heavy mod to simplesamlphp.

jfrank-nih:
I did a little more digging on this one. Found the simplesamlphp mailing list which was helpful. Here's [a good explanation of the different options](https://groups.google.com/g/simplesamlphp/c/ERlbkoFpntw/m/Kld7MKGHTdsJ), which shows that we basically figured it out.

And [here's someone with essentially our problem](https://groups.google.com/g/simplesamlphp/c/sRYMxOQNKa4/m/-Yw5zGn4AwAJ). Basically `session.duration` is what we want, but we want it to extend.

So it looks like a good option would be:

- Set `session.duration` to 1 hour
- Somewhere in the Drupal bootstrapping, if logged in, call simplesamlphp's `setAuthorityExpire` function and extend the duration by another hour from the current time. Maybe do this only on full page loads or maybe on any call. I'm inclined to the latter.

bryanpizzillo:
Closing as:
1. we hit a wall and there is nothing in the code of importance, and the issue has the comments.
2. we don't use acquia pipelines and forks...

PR #3440, opened 2022-07-11
TEST -- Updating only packages

blairl-nih:
Instead of  updating drupal, update the dependent packages


blairlearn:
This was a debugging build

Issue #3451, opened 2022-07-19
Intermittent Issue with json_data field in configurations


bryanpizzillo:
### Issue description
> **ESTIMATE** TBD

There is an intermittent issue where installing or updating a Drupal instance since v1.3.14 can result in the errors listed below. This occurred on one environment (www-test-ac) leading up to v1.3.14, and on ACSF Prod. This issue has not occurred on other ODEs, nor on ACSF Test. The error is occurring during a `features:import:all`.

### Errors
```
In FieldConfigStorageBase.php line 31:
[Drupal\Component\Plugin\Exception\PluginNotFoundException]
Unable to determine class for field type 'json_data' found in the 'field.field.block_content.ncids_mega_menu_content.field_yaml_content' configuration 

DiscoveryTrait.php line 53:
[Drupal\Component\Plugin\Exception\PluginNotFoundException]
The "json_data" plugin does not exist. Valid plugin IDs for Drupal\Core\Field\FieldTypePluginManager are: address, address_country, address_zone, app_module_reference, schema_org_data, hhs_syndication, datetime, daterange, entity_reference_revisions, file, file_uri, image, link, metatag, list_float, list_integer, list_string, path, redirect_source, text, text_long, text_with_summary, viewsreference, boolean, changed, created, decimal, email, entity_reference, float, integer, language, map, password, string, string_long, timestamp, uri, uuid
```

bryanpizzillo:
So, I performed the following steps on www-test-ac:
1. Deployed the 1.3.14 tag and saw it fail.
2. I logged into the server and:
    1. Checked drush pml to verify json_data_field was enabled. _It was enabled._
    2. Tried `blt drupal:update` which is basically what failed in the deployment. _It failed._
    3. Ran `drush cr`
    4. Ran `blt drupal:update` again. **_It worked._**
3. Redeployed the 1.3.14 tag and it succeeded. 

I am going to have to chalk this up to some sort of cache state issue with maybe memcache. IDK why the plugin discovery cache did not get cleared out when Drupal installed. This is `field_types_plugins` in the `cache_discovery` bucket.

santwuma:
Closing ticket - No longer an issue, didn't happen again with subsequent deployments.

santwuma:
Closing ticket - No longer an issue, didn't happen again with subsequent deployments.

PR #3452, opened 2022-07-19
(#3451) Tweak dependencies for json_data

bryanpnih:
- This is really a shot in the dark. These changes are really just the
   the things we normally do with modules, that have never exhibited
   these issues.


bryanpizzillo:
Closing as this should be part of the Drupal 9 upgrade.

PR #3456, opened 2022-07-22
(#3455) Point Adobe Launch Config to Cgov Property

### LINKED ISSUE #3455: Change the Adobe Launch config to dev environment of main property
Currently the CGDP uses the Production environment of the Digital Platform: Extension Development property when it installs a new instance of Drupal. This is for ODEs and for new production ACSF sites. Using the extension development property has become a challenge as:
* There is no way to easily move rules and other configuration between a property configured for extension development and a property not configured for extension development, e.g., the one for www.cancer.gov.
* We have an ever more increasing number of rules for redesign that would require copy/pastes.

So the request is to change the default Adobe Launch configuration to use the **development** environment of the Digital Platform: CGov: EDDL-Enabled property. (`<script src="https://assets.adobedtm.com/6a4249cd0a2c/949e98979342/launch-d90dca90a30f-development.min.js"></script>`)

This will allow the analytics team to easily create and test rules, reducing the time for duplication, as well as not **requiring** the Adobe Experience Platform Debugger to test rules. (It is still helpful to debug broken rules though)

bryanpnih:
ODE: https://ncigovcdode485.prod.acquia-sites.com/
Closes #3455

### DIFF
--- a/docroot/modules/custom/cgov_adobe/config/install/cgov_adobe.settings.yml
+++ b/docroot/modules/custom/cgov_adobe/config/install/cgov_adobe.settings.yml
@@ -8,4 +8,4 @@ exclude_paths:
   - '/node/*/revisions*'
   - '/user/*/edit*'
   - '/user/*/cancel*'
-launch_property_build_url: //assets.adobedtm.com/6a4249cd0a2c/663bb44c2e69/launch-9ed25e181790.min.js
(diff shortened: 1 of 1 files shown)


welshja:
Confirmed that the test ODE has Digital Platform: CGov: EDDL-Enabled developement environment set as the launch tag.

bryanpizzillo (approved):
It has been tested and works as noted in the comments.

PR #3523, opened 2022-09-09
(#3511) Run Front-end Tests in Pipelines

### LINKED ISSUE #3511: Enabler: add front end tests to build process
# Add frontend tests to build 

> **ESTIMATE** TBD

After bumping NCIDS versions, unit tests failed but the pipeline passed. After some quick digging, we found that we do not run our frontend tests during the build process. This should:

- run `npm run lint`
- run `npm run test`
- preferably be one of the first scripts that run so the application fails more quickly

bryanpnih:
Closes #3511

### DIFF
--- a/acquia-pipelines.yml
+++ b/acquia-pipelines.yml
@@ -30,6 +30,10 @@ events:
             script:
               ## Use our override
               - source ${SOURCE_DIR}/scripts/pipelines/setup_app
+        - test-front-end:
+            type: script
+            script:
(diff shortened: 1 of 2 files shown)


bryanpizzillo (approved):

blairlearn:
Running the tests locally, there is a test for `usa-site-alert`, `does not blow up if component does not exist`

The test outputs an error message, but is marked as passing. Lacking a comment in the test code, this seems wrong.

PR #3531, opened 2022-09-14
(#3529) Add PR Check for Commit Naming Standards

### LINKED ISSUE #3529: Implement OCPL Configuration Management Standards Check for pull requests
We now have a GitHub reusable workflow named OCPL Configuration Management Standards Check that will ensure our commit messages match `(#1234) Short description`. This should be added as a workflow to the platform for all pull requests.

bryanpizzillo:
Closes #3529

### DIFF
--- a/.github/workflows/ocpl_cm_standards_check.yml
+++ b/.github/workflows/ocpl_cm_standards_check.yml
@@ -1,0 +1,9 @@
+name: OCPL Configuration Management Standards Check
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, ready_for_review]
+
+jobs:
+  commitlint_remote:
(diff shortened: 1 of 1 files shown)


blairlearn:
It looks good (even if that path is confusing).  The one problem I see is that there is no `workflows` directory under `nciocpl/.github/.github/`.

blairlearn (approved):
That path is even more confusing than I thought!

bryanpizzillo:
> That path is even more confusing than I thought!

Yeah. It is mainly because you can only have workflow templates living in the .github workflow. Yet at the same time ticket templates live in the same repo with different processes. So we chose to just have workflow branches for our versioning and development.

Issue #3225, opened 2021-10-25
Enabler: Create an EntityReferenceSelection plugin


bryanpizzillo:
# Create an EntityReferenceSelection plugin
> **ESTIMATE** TBD

The `DefaultSelector` for defining an entity selector queries the configured entities in the system to allow for selections. This list of entities is then validated when the configuration is imported. This can lead to circular dependencies, as well as cause issues referencing optional modules. For the most part we always set the `target_bundles` to `null`, which means all entities of that type. In some cases, and specifically for this case, `cgov_media_link` should only link to `cgov_file`, `cgov_infographic` and `cgov_video` because they are pages on the site. The issue is that each of those types also has the related pages list on those types, which causes a circular dependency. 

This is something in the JSON API module REALLY cares about this, even though this circular dependency has been working for 2 years under normal circumstances. There is a small possibility this is not an issue with D9.2, but that would be very small.

The task is to create a EntityReferenceSelection plugin, based on `Drupal\cgov_core\Plugin\EntityReferenceSelection\CgovUnpublishedSelection` which is just a sub-class of the DefaultSelection plugin from core. This allows users to add unpublished content to lists and whatnot. Our module should:
1. Replace the entity selection with a text box
   * the user will enter the entities allowed separated by a new line. (the back end will store this as a list, but it is always a text box with one on each line for the admin UI)
   * an empty box will be treated as null, and allow everything.
      * we should never have entity references that allow nothing -- that is dumb.
2. Each entry should be a valid machine name. (all lower case, underscores, etc). The editing interface should validate this.
    * The system will not validate these to ensure that the entities actually exist. 
3. When preparing the entity query, (DefaultSelection::buildEntityQuery) we must take our list and build the entity query

So this may be only inheriting from the DefaultSelection and overriding the validation and editing methods. Technically if the target_bundles are a list then the underlying buildEntityQuery method will work correctly. (I do not think it will throw an exception if you filter the bundles based on some that do not exist)

This should be implemented in a new `cgov_tools` module stored in `docroot/modules/custom/cgov_tools`. This will allow for Kernel and Functional tests to use the minimal profile and run much faster.

bryanpizzillo:
So in Drupal 9.4 this is not an issue. I was able to install with jsonapi and serialization enabled; run unit tests; and use the json API, including looking up media links.

bryanpizzillo:
This may be a nice to have for the future when we do get into circular dependency issues, but for now I am going to close this.

PR #3555, opened 2022-09-27
(#3225) Create an EntityReferenceSelection plugin

ramu-nih:
- Created new Entity reference selection plugin
- Displayed all the entities if the target bundles value is null
- Storing allowed bundles values as a list in backend
- Added Kernel test scripts
- Added Functional test scripts.

See https://github.com/NCIOCPL/cgov-digital-platform/pull/3343 for previous notes


santwuma:
**Next steps:**
1. Bryan to test backend 
2. Bryan to connect with Smita on next dev tasks

Issue #3104, opened 2021-07-02
Story: Remove image deletion for image managers when an image is published


sarinapadilla:
## As a user with image manager permissions, I want to remove the ability to delete published images, so that breaking pages due to reference of deleted images is not possible.

Remove image deletion for image managers for both Image (cgov_image) and Contextual Image (cgov_contextual_image) once an image has been published. If the image has not yet been published (and therefore cannot have been referenced by any other content), the image manager should be able to delete the image.

> **ESTIMATE** TBD

### Acceptance Criteria

```gherkin
Scenario: Image manager tries to delete a new Image before publishing
  Given the user has image manager permissions
    And the user creates a new cgov_image
    And the new cgov_image has not been published (is in an unpublished state)
  When the user tries to delete the cgov_image
  Then the image is deleted
```
```gherkin
Scenario: Image manager tries to delete a new Contextual Image before publishing
  Given the user has image manager permissions
    And the user creates a new cgov_contextual_image
    And the new cgov_contextual_image has not been published (is in an unpublished state)
  When the user tries to delete the cgov_contextual_image
  Then the image is deleted
```

```gherkin
Scenario: Image manager tries to delete a new Image after publishing
  Given the user has image manager permissions
    And the user creates a new cgov_image
    And the new cgov_image has been published (is in a published state)
  When the user tries to delete the cgov_image
  Then they do not have the option to delete the cgov_image
```

```gherkin
Scenario: Image manager tries to delete a new Contextual Image after publishing
  Given the user has image manager permissions
    And the user creates a new cgov_contextual_image
    And the new cgov_contextual_image has been published (is in a published state)
  When the user tries to delete the cgov_contextual_image
  Then they do not have the option to delete the cgov_image
```

### Resources:
* Feature: https://github.com/NCIOCPL/cgov-digital-platform/issues/839

### Notes

---

### Prerequisites
- *required dependency for identifying the problem, scope and then solution for this issue*
- [ ] Short Spike Description - Issue #9999

### Tasks
- [ ] Short Task Description - Issue #9999
- [ ] Short Task Description - Issue #9999

### Technical Notes
- More notes, or technical nuances/direction

adriancofie:
@sarinapadilla @VictoriaSunNIH  Just double checking the req's on this one. Previously there was a stipulation that if the image had EVER been published it would not be deletable by image manager--regardless of it's current publish status. eg. A currently unpublished image that had previously been published. 

[...]

PR #3556, opened 2022-09-27
(#3104) Remove deletion for image managers when an image is published

ramu-nih:
- Restrict image manager to delete published cgov_image and cgov_contextual_image images
  - Added Kernel test script
  - Prevent Image manager role to delete media translations

See https://github.com/NCIOCPL/cgov-digital-platform/pull/3502 for previous notes


santwuma:
Needs requirements - Sarina and Lindsay to connect

PR #3577, opened 2022-10-05
(#3573) Move from Chrome to Chromium

### LINKED ISSUE #3573: Move from Chrome to Chromium for behat testing.
There is no Apple Silicon (ARM64) support for Chrome on Linux. Right now behat relies on Chrome (technically [chromedriver](https://chromedriver.chromium.org/downloads)). Chromium is the open source version (ish) of Chrome, and supports ARM64. The chromedriver can also use Chromium. It actually looks like the chromedriver can use, or is built with(?), chromium. There is a [ubuntu package](https://launchpad.net/ubuntu/bionic/arm64/chromium-chromedriver) for the chromium-based chrome driver.

bryanpizzillo:
Closes #3573


bryanpizzillo:
For the record, this was closed because it did not trigger a CI/CD in Actions.

PR #3599, opened 2022-10-10
Apple Silicon Support Tickets

### LINKED ISSUE #3571: Upgrade to svg-sprite 2.0 (and remove PhantomJS stuff)
A new versions of [SVG-Sprite](https://github.com/svg-sprite/svg-sprite) is here and it no longer has PhantomJS. 🎊🎊🎊🎊  So this ticket is to bump the version and cleanup PhantomJS stuff. This does have a few new package dependencies that are not pure JS,  like [resvg-js](https://www.npmjs.com/package/@resvg/resvg-js) and possibly [playwright-chromium](https://www.npmjs.com/package/playwright-chromium). So in 5 years these things will become problems as they are no longer supported. 😁

### LINKED ISSUE #3573: Move from Chrome to Chromium for behat testing.

### LINKED ISSUE #3579: mozjpeg, called from gulp-imagemin, throws an error.
This one is pretty fun. So the version of imagemin-mozjpeg we use must be built from source to support ARM64. The decoded error message thrown by node is `qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory\n`. The thing is, if there is no build tools it tries to run the intel version, which dies. After installing the tools into the docker container, the current version we have builds and works ok.

So we actually need to install a number of things in the docker container in order to be able to build mozjpeg:
- autoconf
- libtool
- libpng-dev
- m4
- pkg-config
- nasm

### LINKED ISSUE #3580: Remove the AMD64 platform from the web container once ARM64 native support is added.

bryanpizzillo:
This also removes the requirement for PhantomJS.

Closes #3571 
Closes #3573 
Closes #3579 
Closes #3580


KateMashkinaNIH:
Automated regression is passed on ode509. Poked around visually - nothing to note

arcepaul (approved):

blairlearn (approved):

PR #3607, opened 2022-10-18
(#3604) NCIDS Support for Section Nav

### LINKED ISSUE #3604: Mobile Menu needs more information from section nav
The section-nav API implemented in #3243 does return information about the section navigation. However, there is a challenge determining what the "parent section navigation" is for a given section. In the digital platform, the mobile navigation only shows up to 4-levels deep. Many of the sections of the web site fall below those 4 levels. The mobile navigation allows a user to navigate all the way up to the primary menu from the page they are on. Given the mobile navigation is only 4 levels deep, and many (most?), section navigations do not appear in the mobile nav, we must fetch multiple navs from the API. (e.g., the section navigation for the page, any other section navigations, and the mobile navigation)

This change request is to also return 2 extra pieces of information:
* The next navigation to fetch when the user has "moved" out of the current section tree. This should include:
   * The site section id of the next navigation root.
   * The menu type of of that navigation:
      * `section-nav` - for a site section.
      * `mobile-nav` - for the mobile nav.
* The ID of the site section within that parent menu where the returned section navigation should be placed.
   * This should follow all the display rules as defined in ticket #3243. So there may be gaps if multiple parent site sections do not have valid landing pages, or are marked as hiding in navigation. That is ok. Find the closest. 
   * This ID will sometimes be the same ID as the section nav root id.

bryanpizzillo:
- This also adds in Spanish test cases.

Closes #3604

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/src/Controller/CgovNavTreeController.php
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/src/Controller/CgovNavTreeController.php
@@ -4,9 +4,8 @@ namespace Drupal\cgov_site_section\Controller;
 
 use Drupal\Core\Controller\ControllerBase;
 use Drupal\taxonomy\TermInterface;
-use Drupal\path_alias\AliasManagerInterface;
+use Drupal\Core\Render\RendererInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
-use Drupal\Core\Entity\EntityTypeManager;
 use Drupal\Core\Cache\CacheableJsonResponse;
 use Drupal\Core\Cache\CacheableMetadata;
 use Symfony\Component\HttpFoundation\JsonResponse;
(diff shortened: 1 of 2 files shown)


KateMashkinaNIH (approved):

docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/src/Controller/CgovNavTreeController.php
```diff
@@ -17,294 +16,469 @@
...
-      $container->get('path_alias.manager'),
-      $container->get('entity_type.manager'),
+      $container->get('renderer'),
     );
   }
 
   /**
    * Return section nav items.
    */
   public function getSectionNav(TermInterface $taxonomy_term = NULL) {
```
smitahantziosps, line 49: Instead of two functions "getSectionNav" and "getMobileNav" , We can create a single function "getNav" and pass the $menu_type as parameter "section_nav" and "mobile_nav".

bryanpizzillo: If you look at the 2 functions you will see that the logic is more than just section_nav or mobile_nav. So to avoid a conditional in a `getNav` function, they have been split up.

smitahantziosps: Ty! I see now why you have split it up. Thanks all looks good

smitahantziosps (approved):

PR #3666, opened 2022-11-30
(#3665) Remove h4 footer headers

### LINKED ISSUE #3665: The Footer link group headings are appearing as accessibility issues
This is actually "just" a content issue. 

The footer markup we say to use got updated a while ago because we should not use heading elements (e.g. `<h4>`) for items that are not an actual page heading. This is mostly due to the fact that if you are going to use a heading, you need to make sure that no levels are skipped. In the case of the footer html on dev, the "About", "Resources" and "Policies" sections are `<h4>` elements. On many pages there are only `<h1>` and `<h2>` elements. This means that we are skipping `<h3>`.

The correct code can be seen at https://designsystem-dev.cancer.gov/develop/components/footer#nci-big. 

NOTE: I don't know if there are any other elements that are not correct, I just found this issue in a lighthouse report.

olitharp-nci:
closes #3665


olitharp-nci:
#3679

PR #3670, opened 2022-11-30
(#3667) Fix autosuggest analtyics

### LINKED ISSUE #3667: Analytics value for autosuggest on global search bar not working as expected.
The autoSuggestUsage, numCharacters, charactersTyped and numSuggestsSelected values are not working as expected

Per Joe: 

These components are getting values - but not the correct values. so the autoSuggestUsage is sending "true" instead of the values specified. The numCharacters, charactersTyped and numSuggestsSelected are passing numbers (as expected), but not the correct numbers based on the requirements

The requirements for these components are captured in: https://github.com/NCIOCPL/cgov-digital-platform/issues/3279

olitharp-nci:
Closes #3667
WIP


olitharp-nci:
#3673

Issue #3266, opened 2021-12-15
Enabler: Update PHP to 8.0


bryanpizzillo:
# Update PHP to 8.0
> **ESTIMATE** TBD

Acquia is retiring PHP 7.4 10/2022 in lieu of 8.0. Technically PHP 8.1 is out. So if ACSF and ACE support 8.1 when this ticket is being worked on then it would make sense to go to 8.1. **8.1 has issues with our code, so this ticket is just for 8.0.**

> **_NOTE:_** This should probably be done around the time that Acquia starts defaulting ODEs to 8.0.

### Tasks
- [x] Update Docker stack
   * This includes figuring out which PHP module .ini files have to be changed out. (e.g. xdebug.ini changed a bunch from 7.2->7.3)
- [x] Update acquia-pipelines.yaml
- [ ] Change ACE and ACSF web server settings through the UI.
- [ ] Anything else missed from above...

### Previous upgrade tickets
* #3105
* #2523

PR #3736, opened 2023-01-16
(#3266) Upgrade to PHP 8.0

jfrank-nih:
- Updates Acquia Pipelines
- Updates Docker
- Removes json PHP module which is now core
- Updated php-dom-wrapper to work with PHP 8.0
- Fixed an error in our YAML Content subscriber that was incorrectly trying to access a key of an array without checking first.
- Update press release content to add values key to object that is required to be there. The underlying module code expects the key to be there. This was most likely just quietly not working before.
- Bad YAML content failed to set field_search_engine_restrictions correctly. It is unknown why cgov_metatag.module worked with the 0 value with PHP 7 vs PHP 8 failing. This would most likely be something with the Field loading or entity interface that is behaving differently with PHP 8.
- Updates phpunit.xml schema

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,10 +78,10 @@ jobs:
       - name: Setup PHP with tools
         uses: shivammathur/setup-php@v2
         with:
-          php-version: '7.4'
+          php-version: '8.0'
           ## This list of extensions matches those we install with our docker file, then I removed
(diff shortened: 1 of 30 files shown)


KateMashkinaNIH:
passed automated regression tests - and much quicker too!

KateMashkinaNIH:
re-ran regression on ODE529 - pass

bryanpizzillo (approved):

PR #3759, opened 2023-02-04
(#1700) Enable automated logouts

### LINKED ISSUE #1700: Configure inactivity timeout to 120 minutes

### LINKED ISSUE #2061: Admin toolbar User Logout button does not work with SSO.
The logout button in the Drupal menu under `username` > Log out does not currently work. You get immediately logged back in. In reality you are being logged out, detected as not logged in, routed to Okta (where your session is usually still live), routed back to Drupal, and logged back in automatically.

Solution is to point the logout button to the Okta logout URL so that you are logged out of Okta and break the loop.

**Testing**

- Log in to CMS
- Click logout button
- Expected: You land on the Okta log in page; logging back in again takes you to the Okta dashboard where you can click on the CMS name to return to the CMS

@bryanpizzillo's original:

> It keeps logging me back in.

jfrank-nih:
Testing scenarios on tickets.

Closes #1700 
Closes #2061

### DIFF
--- a/blt/blt.yml
+++ b/blt/blt.yml
@@ -26,13 +26,13 @@ drush:
 modules:
   local:
     enable: [dblog, devel, views_ui, cgov_yaml_content]
-    uninstall: [syslog, acquia_connector, shield, acsf, simplesamlphp_auth, cgov_saml_auth_config, seckit, cgov_https_config, cgov_caching_nocdn, cgov_caching_cdn, acquia_purge, akamai]
+    uninstall: [syslog, acquia_connector, shield, acsf, simplesamlphp_auth, autologout, cgov_saml_auth_config, seckit, cgov_https_config, cgov_caching_nocdn, cgov_caching_cdn, acquia_purge, akamai]
   ci:
     enable: {  }
-    uninstall: [acquia_connector, shield, simplesamlphp_auth, cgov_saml_auth_config, cgov_https_config, akamai]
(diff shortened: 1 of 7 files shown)


bryanpizzillo (approved):
Changes look good to me as long as they pass testing.

Issue #3770, opened 2023-02-09
Remove cgov_migration module and other migrate modules.


bryanpizzillo:
### Issue description
> **ESTIMATE** TBD

We still have all the migration data from the migration 4 years ago. We should just remove the cgov_migration module as any of the code was really for Percussion migrations and we don't have Percussion anymore. (Also it is a lot of data that we don't need on our development machines or web servers) There is one gotcha - we still use migrate to load the site navigation for the YAML content.

## Tasks
1. Find a way to load ODE site sections.
    * This is currently done via `blt cgov:install:site-sections` which is called by `blt custom:install_cgov_yaml_content_by_module`.
    * `blt ccgov:install:site-sections` uses migrate, and our old source xml of percussion navigation items, to load the sections into the CMS.
    * It is good to have a large amount of site sections when testing our custom taxonomy manager, so we should keep this close to a www navigation tree.
    * Technically this loads raw HTML, which might just be for the OLD mega menu contents. So we should not need to care about that.  
    * One part of this is to delete the `blt/src/Blt/Plugin/Commands/SiteSectionsCommands.php` (i.e. `blt cgov:install:site-sections`) command.  
1. Update `hooks/common/post-code-deploy/post-code-deploy.sh` and `hooks/common/post-code-update/post-code-update.sh` to remove migration things.
2. Remove any migrate modules from composer.json (and the lock obviously), as well as from our cgov_site.info.yml profile  config. If there are any other dependencies in module.info.yml files, they should be removed from there.
3. Remove the cgov_migration module completely
4. Ensure no production site has any of the modules enabled. (Otherwise we need to turn them off before deploying)

PR #3773, opened 2023-02-13
(#3770) Remove Migrate Modules

bryanpizzillo:
This removes migrate modules and instead uses yaml_content for navigation structure.


adriancofie (commented):
Tested locally and clicked around, looks good to me

adriancofie (approved):

KateMashkinaNIH:
Both manual and automated regression passed

bryanpizzillo:
@blairlearn approved it via Slack.

Issue #3187, opened 2021-10-01
Upgrade to BLT 13


bryanpizzillo:
### Issue description
> **ESTIMATE** TBD

I am entering this ticket as we are doing the Drupal 9 + BLT 12 update. BLT 13 has made some changes that are things that we  will have issues with, for example the features configuration strategy has been removed. This ticket is really to get this out there that it will have to be done sooner rather than wait for the next major upgrade.

bryanpizzillo:
NOTEs: 
* `toggle:modules` is removed in BLT 13
   * Basically they think config-split should handle it, and it does if you are using config split.
   * There is also a fixed issue with how it reads the environment that might not work with  
* https://github.com/acquia/blt-launcher exists, and should work like drush-launcher and not require the shell function anymore
* Behat is now delivered through blt-behat plugin. Acquia also no longer provides support for Behat and BLT. In fact, Acquia recommends using PHPUnit instead of behat. See https://github.com/acquia/blt-behat.
* **And they removed support for features.** 
   * There is some claim that it does not work in Drupal 9, but the Features community fixed the issues.
   * 
   * So this means we need to set the configuration strategy to none, and update all of our scripts to use features, or rebuild it back in.
   * I do not have the appetite to move to config-split, especially when there are so many known issues with handling feature, site, profile, environment splits. See https://support.acquia.com/hc/en-us/articles/360024009393-Managing-Configuration-with-Config-Split for more "here are things that don't work correctly".

bryanpizzillo:
## Attempt 1
1. Install 13.0.1 to minimize the changes needed to get from 12->13. Failed.
2. `composer require acquia/blt:^13.7.0 drush/drush:^11.4.0 --with-all-dependencies` -- Drush 11 is a requirement of Blt 13.?

bryanpizzillo:
## BLT 13.7.0 + Drush 11.4.0 Upgrade log
```
❯ composer require acquia/blt:^13.7.0 drush/drush:^11.4.0 --with-all-dependencies
Info from https://repo.packagist.org: #StandWithUkraine
./composer.json has been updated
Running composer update acquia/blt drush/drush --with-all-dependencies
Gathering patches for root package.
> DrupalProject\composer\ScriptHandler::checkComposerVersion
Loading composer repositories with package information
Info from https://repo.packagist.org: #StandWithUkraine
Updating dependencies
Lock file operations: 3 installs, 36 updates, 2 removals
[...]

PR #3775, opened 2023-02-15
(#3187) Upgrade to BLT 13

bryanpizzillo:
- Backported changes to our drupal:config:import replacement.
- Rebuilt toggle-modules as BLT 13 removed the command. Also, Drush 11 changed how it handles an empty list for disabling, so we had to make additional modifications to that.


bryanpizzillo:
@kate-mashkina passed this on 2/27. The only issues encountered with testing were related to issues on the ACSF environment itself, where clamav was consuming too much memory, as the server seem over taxed with mysql and everything else that was running.

blairlearn (approved):
Looks good to me.  It doesn't blow up my local machine.  If it's good with @KateMashkinaNIH, it's good with me.

PR #3783, opened 2023-02-28
(#3513) Add CTHP card title field

### LINKED ISSUE #3513: Store browser title field separate from page title field
### Issue description
The CDR users would like to be able to edit the browser title for Cancer Information Summary (CIS) pages in the CDR.

> **ESTIMATE** TBD

- Drupal accepts from the CDR and stores a possibly different value for the browser title than the value used for the title displayed at the top of the page, and actually uses this new value as the browser title (sent to the browser as the `<title>` element of the `<head>` block of the HTML document).

There are currently two fields for storing titles in CIS nodes.

1. A field labeled "**Title**" on the node-editing form in Drupal, populated from the `SummaryTitle` field in the CDR document. This is used both for the browser title (`html/head/title`, displayed in the browser tab and tooltip, possibly truncated, with the site identification appended) and the title displayed at the top of the page.
2. A field labeled "**Browser Title**" on the Drupal node-editing form, populated from the CDR `AltTitle` element with the type element equal to "**Short**," and used on the _Cancer Type Home Page_. Despite the label for this field (as well as its description, which says that it is displayed in the browser tab), the value stored in this field is **not** used as the browser title.

#### Titles on Drupal node form
image
#### Title element in the head block of the HTML document
image

### Related issues
https://tracker.nci.nih.gov/browse/OCECDR-5104

bkline:
This will allow the field_browser_title value, populated from the CDR, to be used as for the browser title, as was originally intended.

Tasks include:
- create new field_ctrp_card_title field, 100 characters max
- modify the RESTful services to handle the new field
- modify the tests to include the new field
- adjust the YAML sample content to include the new field
- add the new field to the CTHP: Guide Card List view
- modify the cthp guide card list item template to use new field
- use the field_browser_title value for the head/title element

Deployment will be coupled with a full load of the CIS summaries, to populate the new field with appropriate values, and to adjust the values in the browser title field so that they are unique.

Closes #3513.

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_yaml_content/content/pdq_cis_hp_acupuncture.content.yml
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_yaml_content/content/pdq_cis_hp_acupuncture.content.yml
@@ -21,6 +21,8 @@
   field_date_updated:
     value: '2021-06-23'
   field_browser_title:
+    value: 'Acupuncture (PDQ®)'
+  field_cthp_card_title:
     value: 'Acupuncture'
   field_page_description:
     value: 'Acupuncture is a complementary therapy used by cancer patients to manage symptoms related to cancer and its treatment. Get detailed information about acupuncture and its use as a complementary therapy in cancer in this summary for clinicians.'
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_yaml_content/content/pdq_cis_hp_adult_treatment.content.yml
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_yaml_content/content/pdq_cis_hp_adult_treatment.content.yml
@@ -21,6 +21,8 @@
   field_date_updated:
     value: '2018-11-02'
   field_browser_title:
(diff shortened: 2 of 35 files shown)


bryanpizzillo (approved):
Adding my review, but this will need to wait for the merge of the release branch tonight and then get rebased.

blairlearn:
This works with a clean install of the standard content, but with existing data (i.e. after updating), embedded CTHP card titles are blank.

Steps to reproduce locally

1. Exit from the container
2. `git checkout develop`
3. `composer cgov-clean && composer install`
4. Enter the container
5. `blt cgov:reinstall --no-interaction`
6. Browse to http://www.devbox/types/breast. The embedded summary link titles are present.
7. Exit, but do not shut down, the container.
8. `git checkout pdq-title-fields`
9. `composer cgov-clean && composer install`
10. Re-enter the container
11. `blt drupal:update --no-interaction`
12. `drush cr`
13. Browse to http://www.devbox/types/breast. The embedded summary link titles are blank.

image

blairlearn:
Per discussion in Slack with @KateMashkinaNIH, QA has verified this fix causes no regressions.

Issue #3768, opened 2023-02-09
Enabler: Upgrade Drupal to 9.5


bryanpizzillo:
# Upgrade Drupal to 9.5
> **ESTIMATE** TBD

The title says it all. Go from 9.4 to 9.5.

PR #3821, opened 2023-04-12
(#3768) Add translation to workaround issue

bryanpizzillo:
- We updated Drupal translations and this causes issues for our
  translations being overwritten unless we have a newer file.


bryanpizzillo:
This works on an ODE.

PR #3832, opened 2023-04-25
(#3484) Update Paragraphs and Friends

### LINKED ISSUE #3484: Enabler: Update paragraphs and paragraphs_asymmetric_translation_widgets modules
# Update the paragraphs and paragraphs_asymmetric_translation_widgets modules to the current version.

As part the Drupal 9.4 upgrade, it was discovered that behavior of the paragraphs and paragraphs_asymmetric_translation_widgets had changed. Additionally, the asymmetric translation widget is now considered "stable" instead of "experimental".

The changes don't seem to have impacted the user experience, but there are questions of whether there are impacts to the CDR publishing scripts and the cleanup of removed summary sections.

Things to be done:
- Switch back to "minimum version" constraints
  - `compose require drupal/paragraphs:^1.12 drupal/paragraphs_asymmetric_translation_widgets:^1.0`
- Test CDR publishing
  - Add/update/delete
  - Summary section cleanup

> **ESTIMATE** TBD

bryanpizzillo:
This updates the paragraphs, paragraphs_asymmetric_translation_widgets and entity_reference_revisions module. This ALSO updates from using the classic paragraph's widget to the "experimental" widget. The "experimental" widget is no longer "experimental". The experimental widget changes how paragraphs manages the actual paragraph entities and could break CDR Summary cleanup code.

Closes #3484


bkline:
Nothing much to review in the composer files. Unless the updates actually modify table or column names (which seems far-fetched) I don't see how they could break the cleanup code, but we can always (a) do a before/after table structure comparison; and (b) test with a publishing job or two once this gets stood up. Even if the cleanup breaks, the "damage" would be leaving around more useless rows than we want. If that actually happens (again, very unlikely) we'll just enter another ticket to fix it.

bryanpizzillo:
> Nothing much to review in the composer files. Unless the updates actually modify table or column names (which seems far-fetched) I don't see how they could break the cleanup code, but we can always (a) do a before/after table structure comparison; and (b) test with a publishing job or two once this gets stood up. Even if the cleanup breaks, the "damage" would be leaving around more useless rows than we want. If that actually happens (again, very unlikely) we'll just enter another ticket to fix it.

This won't be a table change issue, but it could be an issue with how it deals with the underlying entity reference revision storage.

bkline:
@bryanpizzillo I've spent a number of hours wrestling with this. As I have documented [elsewhere](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Drupal-Storage-for-PDQ-Summaries), this rabbit hole is very complicated and it makes my brain hurt.

**TL;DR** I think we'll be OK. If you get any late Friday night panic phone calls, feel free to share the pain with me. 😂

As far as I can tell, we don't lose any revisions we want to keep.

> When you call translate on a node all the paragraph fields will have been copied in, instead of being empty.

I think that's OK, as the API code always hands Drupal the complete set of summary sections, rather than try and manipulate an existing list of `paragraph_item` entities.

> When you delete a translation, paragraphs asymmetric widget will delete all the paragraph entities.

I'm not sure what rôle a "widget" would play in this scenario. I think of "widget" as a user interface component, and this is all happening programmatically, bypassing the user interface. In any case, when I delete a Spanish summary, its paragraph entities do seem to be gone.

On the other hand, I'm seeing odd behavior in an edge case. When I delete a Spanish summary without doing anything with the original English summary, the pass to purge older revisions gets the error message

> Default revision can not be deleted

The purge logic gets all of the existing revisions for a node and sorts them so the newest revisions are at the front of the list. Then it walks through revisions, keeping track of how many have been identified for each language as "keep this revision" until we have three to keep for each language or we run out of revisions. The criteria for whether we keep a revision for a given language are:
1. Drupal reports that the language is affected by the revision (`$translation->isRevisionTranslationAffected()` returns `TRUE`); and
2. Drupal reports that the revision is published for this language (`$translation->isPublished()` also returns `TRUE`)

It appears that Drupal considers that the most recent revision is the default revision for the English summary even if the English "translation" for the node was not "affected" by that revision (we had only kept it because it was one of the revisions to be preserved for the Spanish summary, and now we want to get rid of that revision). The failure to prune that revision doesn't cause the publish job to fail, and having extra revisions around isn't a fatal problem. The next time we save a revision for the English summary pruning will again work correctly for that node. I don't know for certain whether this behavior is new with recent changes to Drupal core or the `paragraphs` module, or if it's been that way all along. I do know that we've never hit this edge case, because I went through all of the archived publishing logs back to 2019 (well before we were doing any pruning) and the log message the code would have recorded never appears.

I don't know how much more you want me to keep digging, but I need to take a break to let me head stop throbbing and to deal with two major releases I've got in flight (not counting the PDQ browser title stuff).

Dealing with very complicated APIs which interact with each other in seriously under-documented and conflicting ways is NOT my favorite pastime. 😩

bryanpizzillo:
@bkline 
> On the other hand, I'm seeing odd behavior in an edge case. When I delete a Spanish summary without doing anything with the original English summary, the pass to purge older revisions gets the error message
What is the behavior in the develop branch? To answer your comment about "I don't know how much more you want me to keep digging," if a new error appears because of the update, then yes, please see what is happening there.

bkline:
OK, will keep digging.

KateMashkinaNIH (approved):

PR #3891, opened 2023-06-01
(#3890) Fix Blogs RSS

### LINKED ISSUE #3890: Blog RSS fix does not apply when Updating Existing Site
The fix for #3640 does not apply to existing sites, including ACSF. The fix was to swap the `nid` and `field_date_posted` under the `sorts` key. While the order of `sorts` matters, Features is not detecting as a change that needs to be imported. A workaround is to update the admin label for the sort field, which *will* trigger it as a change.

### Related Tickets
* #3640

bryanpizzillo:
- Needed to update admin_label to force features into thinking that it
  needs to update the configuration state.

Closes #3890


KateMashkinaNIH:
RssFeed issue is fixed - verified earlier posted date is in correct order

PR #3908, opened 2023-06-07
(#3902) Wire up raw html block analytics

### LINKED ISSUE #3902: Implement analytics support for NCIDS Raw HTML Blocks
NCIDS Raw HTML Blocks do not have analytics per se, but there will be links in the content that would be good to track. This exact details will vary based on what HTML has to be made. This ticket is to create a generic way of handling link click tracking within a Raw HTML block.

### Proposed Solution
Currently we track "row" (`pageRows`, `pageRowItems`) analytics based on `usa-section` elements with a `data-eddl-landing-row` data attribute. This is the same logic across all the NCIDS Landing Page Components. An example can be seen below. 
```
<div class="usa-section usa-section--light" data-eddl-landing-row>
   ...CONTENTS...
</div>
```
This should be done for all raw HTML blocks as this is how we know the total rows on a NCIDS Landing Page. (We cannot force this, but will ask Drupal editors to always include this)

This is just one part of the puzzle. The other analytics required for a link on the other landing components are:
 `rowItems`, `rowItemIndex`, `componentType`, `componentTheme`,`componentVariant`, `title`, `linkType`, `linkText`, `linkArea`, `totalLinks`, `linkPosition`.

We can use a feature of JavaScript for handling data attributes, `HtmlElement.dataset`, which will turn dashed attribute names (e.g., `data-foo-bar`), into a "camelCased" name (e.g., `fooBar`). So if we have a series of data attributes that follow a pattern we can have an element like:
```
<a 
     href="/foo"
      data-eddl-landing-rawhtml
      data-eddl-landing-rawhtml-title="{{innertext}}"
      data-eddl-landing-rawhtml-component-variant="Raw HTML"
      ...
      >This is the link text</a>
``` 

Then we can have Javascript that can turn those elements into a data object for the analytics.
```
const trackElem = (elem) => {
  const clickData = Object.entries(elem.dataset)
    .reduce(
      (ac, [key, val]) => {
        if (!key.match(/^eddlLandingRawhtml.+/)) {
          return ac;
        }
        const cleanName = key.replace(/^eddlLandingRawhtml/,'');
        return {
          ...ac,
          [cleanName[0].toLowerCase() + cleanName.slice(1)]: val,
        };
      }, {});
  // Get row info    
  trackOther(
     `LP:RawHTML:LinkClick`,
     `LP:RawHTML:LinkClick`,
     {
        location: 'Body',
        (other data points),
         ...clickData
     }
  );
}

```

Some values should never change, these are: 
* location: 'Body'
* componentType: 'Raw HTML'
* pageRows: total number of rows on the page overall
* pageRowIndex: the index of the selected row from the link click

Other values will have a default, but can be overridden:
* componentTheme
* componentVariant
* title
* linkArea
* linkText
* linkType
* rowItems: 1
* rowItemIndex: 1
* totalLinks: number of links in the row
* linkPosition: the index of the selected link in the row

### What's the updated acceptance criteria?
```gherkin
Scenario: Analytics event for guide card clicks
Given a user is on any page of cancer.gov with guide cards
When the user clicks on a link within the guide card
Then there should be an NCIDataLayer Other event 
  And the event has the name “LP:RawHTML:LinkClick”
  And the event has the the linkname “LP:RawHTML:LinkClick”
  And this event should have the following data:
    | key | value|
    | location| Body|
    | pageRows| This should return the number of rows on the page overall |
    | pageRowIndex | This should return the index of the row number on the page that was clicked |
    | rowItems| Pass as a data attribute. Default is number of components in the row that was clicked.  |
    | rowItemIndex | Pass as a data attribute. Default is the index of the component in the rows that was clicked. |
    | componentType| Raw HTML |
    | componentTheme | Pass as a data attribute. Default is `N/A` but `Light` or `Dark` should be used when available |
    | componentVariant | Pass as a data attribute. Default is `N/A`  |
    | title | Pass as a data attribute. Undefined if no title is passed.  |
    | linkType | Pass as a data attribute. Default is `N/A` but `Internal`, `External`, or `Media` should be used when available |
    | linkText | Pass as a data attribute. Defaults to the text content of the link clicked.   |
    | linkArea | Pass as a data attribute. Defaults to `N/A`   |
    | totalLinks | Pass as a data attribute. Defaults to total number of links in the row. |
    | linkPosition | Pass as a data attribute. Defaults to the index of the link in the row that was clicked. |
```

- Screenshot

### Related Tickets
* https://github.com/NCIOCPL/cgov-digital-platform/issues/3848
* Original Story Ticket #9999

olitharp-nci:
Closes #3902

https://ncigovcdode575.prod.acquia-sites.com/ncids-raw-html-test


bennettcc:
Analytics has approved this PR as of 6/8

bennettcc (approved):

bryanpizzillo (approved):

PR #3916, opened 2023-06-12
Replacing behat with cypress, adding parallel

### LINKED ISSUE #3695: Enabler: Replace Behat with Cypress
# Replace Behat with Cypress
> **ESTIMATE** TBD

Behat is a behavioral testing language that uses Gherkin for test features, and PHP to handle step definitions. Within the BLT stack, Behat is used to do automation against an empty Drupal/cgov_site installation. You write your Gherkin scenarios, and it automates a browser, or directly inserts data into Drupal, to test the codebase. While our stack has on a small amount (3) of Behat tests, we really do not use it and instead use the Drupal Test framework.

We do have the need to create a new behat test as we have a test case that is ill-suited for the Drupal Test framework. Instead of struggling through a Behat test, we should just replace it with cypress for the following reasons:
1. The team does not have a lot of Behat experience. (It is not common outside of Drupal, and even within the Drupal community, the vast number of devs don't do Behat.
2. JS is the common language across the team and Cypress (technically cypress-cucumber-preprocessor) is the behavioral test stack across our projects 
3. Behat seems to be loosing ground within the PHP community. Even Acquia [recommends](https://github.com/acquia/blt-behat#acquia-blt-behat) to use something different.
4. Behat has a number of dependencies that actually caused us issues earlier in 2022. See issue 21 on https://github.com/acquia/blt-behat.
5. While the Drupal Test framework does handle our needs, it is actually quite slow. Moving some browser tests to cypress might decrease the amount of time (~16min) it takes for our CI tests to run. 

This enabler consists of:
1. Implementing cypress in `<repo_root>/tests/cypress`
   * This will require updating blt and ci scripts for setting up environments. (e.g., `npm ci` in the cypress folder)
   * A test command for blt should be created -- `blt test:cypress`. This should also get included when the `blt test` command is run.
      * This command should start the webserver. (it is an attribute when you create a test function -- go look at blt)
   * Cypress should be setup such that [Cypress Github Action](https://docs.cypress.io/guides/continuous-integration/github-actions#Cypress-GitHub-Action) can be used.
   * This means that `blt test:cypress` is used locally, but github action in the CI pipeline.
      * The github action allows for parallelization
2. Porting of the existing Behat tests to cypress
3. Removal of Behat and its friends. Look at acquia/blt-behat for an idea of the dependencies that can be removed.

psdnimishe:
Closes #3695


blairlearn:
Replaced by #4256

PR #3928, opened 2023-06-17
(#3926) Bump Docker image to Ubuntu 20.04

### LINKED ISSUE #3926: Enabler: Bump Docker image to Ubuntu 20.04 (PHP is currently broken)
# Bump Docker image to Ubuntu 20.04
> **ESTIMATE** TBD

The Apt repository for PHP that we use is `ppa:ondrej/php`. According to a reddit post, ondrej removes old package versions for Ubuntus that pass their EOL date. Bionic Beaver, the Ubuntu we have been using as it is closest to Acquia's, went EOL 5/31/2023. As such, the packages were removed. So we can no longer build the local Docker images because PHP cannot be found.

So this ticket is for bumping the image to use 20.04, which is what we use in Github Actions (it did not support Bionic Beaver either). The thinking is that if our CI/CD is using this and those artifacts work on production, then we should be able to do it locally. This will give us until 05/2025.

## Additional Notes
* `chromium-browser` now requires Ubuntu's snap program, which is their app storefront. This is not supported within Docker. I found `ppa:phd/chromium-browser` which seems to be kept up to date.

bryanpizzillo:
- Also removed ClamAV
- Had to mess around with how chromium-browser is installed.

Closes #3926

### DIFF
--- a/docker/web/Dockerfile
+++ b/docker/web/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:bionic
+FROM ubuntu:focal
 
 ################################################################################
 #                                                                              #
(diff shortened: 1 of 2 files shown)


bryanpizzillo:
STEPS:
1. Changed Dockerfile from `ubuntu:bionic` to `ubuntu:focal`
2. Changes `mysql-client-5.7` to `mysql-client-8.0` in the package installation.
    * There is no pre-build mysql client for 5.7 available for 20.04
4. Tested the following (on ARM64 and AMD64):
     * `blt setup --no-interaction`
     * `blt tests:behat --environment=ci --define tests.run-server=true -D behat.web-driver=chrome`
     * `blt custom:paratest`
     * `blt tests:frontend:run --define tests.run-server=false`
     * `blt cgov:reinstall --no-interaction`

blairlearn (approved):

Issue #3964, opened 2023-07-14
META: Fix PHPStan issues


santwuma:
Update Drupal Core in PHPStan

### Modules to be updated
- [x] app_module
- [x] json_data_field
- [x] cgov_application_page
- [x] cgov_blog
- [ ] cgov_event
- [ ] cgov_core
- [ ] cgov_home_landing
- [ ] cgov_image
- [ ] cgov_infographic
- [ ] cgov_list
- [ ] cgov_mail
- [ ] cgov_schema_org
- [ ] cgov_site_section
- [x] cgov_vocab_manager
- [x] pdq_cancer_information_summary
- [x] pdq_core
- [x] pdq_drug_information_summary
- [x] pdq_glossifier

blairlearn:
## phpstan errors as of May 28, 2025

```
 ------ ------------------------------------------------------------------------------------------- 
  Line   modules/custom/json_data_field/src/Plugin/Field/FieldFormatter/JsonDataFieldFormatter.php  
 ------ ------------------------------------------------------------------------------------------- 
  53     Access to an undefined property Drupal\Core\Field\FieldItemInterface::$format.             
         💡 Learn more: https://phpstan.org/blog/solving-phpstan-access-to-undefined-property       
 ------ ------------------------------------------------------------------------------------------- 

 ------ ------------------------------------------------------------------------------------------------------------- 
  Line   profiles/custom/cgov_site/modules/custom/cgov_application_page/cgov_application_page.module                  
 ------ ------------------------------------------------------------------------------------------------------------- 
  28     Access to an undefined property Drupal\Core\Entity\EntityInterface::$field_application_module.               
         💡 Learn more: https://phpstan.org/blog/solving-phpstan-access-to-undefined-property                         
  30     Access to an undefined property Drupal\Core\Entity\EntityInterface::$field_application_module.               
         💡 Learn more: https://phpstan.org/blog/solving-phpstan-access-to-undefined-property                         
  59     Access to an undefined property Drupal\Core\Entity\EntityInterface::$field_application_module.               
         💡 Learn more: https://phpstan.org/blog/solving-phpstan-access-to-undefined-property                         
  61     Access to an undefined property Drupal\Core\Entity\EntityInterface::$field_application_module.               
         💡 Learn more: https://phpstan.org/blog/solving-phpstan-access-to-undefined-property                         
[...]

PR #3951, opened 2023-07-10
(#3767) Alternate "drupal-check" PR

bryanpizzillo:

PR #3967, opened 2023-07-18
(#3964) Fixed PHP Stan issues for cgov_core module

navsunka:


docroot/profiles/custom/cgov_site/modules/custom/cgov_core/tests/src/Kernel/FieldStorage/CGovFieldStorageTest.php
```diff
@@ -142,6 +142,7 @@ public function testFields() {
     // Test the field persistance when it has been removed from all nodes.
     foreach ($this->fieldsToTest as $fieldToTest) {
       // Remove from our content type.
+      /** @var \Drupal\Core\Entity\EntityInterface $field */
```
navsunka, line 145: I have tried removing this and checking $field as instance of a EntityInterface but not sure why the error not gone away.

@bryanpizzillo

docroot/profiles/custom/cgov_site/modules/custom/cgov_core/tests/src/Kernel/FieldStorage/CGovFieldStorageTest.php
```diff
@@ -143,6 +143,7 @@ public function testFields() {
     foreach ($this->fieldsToTest as $fieldToTest) {
       // Remove from our content type.
       $field = FieldConfig::loadByName('node', 'ponies', $fieldToTest["name"]);
+      /** @var \Drupal\Core\Entity\EntityInterface $field */
```
navsunka, line 146: this delete method is an instance of EntityInterface but I tried setting the instance of, still the error exists. so i have not removed @var as of now.

@bryanpizzillo

docroot/profiles/custom/cgov_site/modules/custom/cgov_image/cgov_image.module
```diff
@@ -78,15 +78,18 @@ function cgov_image_media_view(array &$build, EntityInterface $entity, EntityVie
...
  *   The entity object.
- * @param Drupal\Core\Entity\Display\EntityViewDisplayInterface $display
+ * @param \Drupal\Core\Entity\Display\EntityViewDisplayInterface $display
  *   The entity view display holding the display options configured for
  *   the entity components.
  * @param string $view_mode
  *   The view mode the entity is rendered in.
  */
 function _cgov_image_add_enlarge_link(array &$build, EntityInterface $entity, EntityViewDisplayInterface $display, $view_mode) {
+  if(!($entity instanceof MediaInterface)){
```
bryanpizzillo, line 90: You should change the signature of the function to be a MediaInterface; then in the calling function do the check. This is because this is not a hook that is forced to use the most generic of interfaces.

docroot/profiles/custom/cgov_site/modules/custom/cgov_image/cgov_image.module
```diff
@@ -131,7 +134,9 @@ function _cgov_image_filter_view_image(
   if (in_array($img_type, ['article', 'enlarged'])) {
     return;
   }
-
+  if(!($entity instanceof MediaInterface)) {
```
bryanpizzillo, line 137: Same as above. Change the signature to require MediaInterface.

docroot/profiles/custom/cgov_site/modules/custom/cgov_image/cgov_image.module
```diff
@@ -27,7 +28,6 @@ function cgov_image_media_view(array &$build, EntityInterface $entity, EntityVie
   if ($entity->bundle() != 'cgov_image' && $entity->bundle() != 'cgov_contextual_image') {
```
bryanpizzillo, line 28: This method should check if $entity is instanceof MediaInterface. Then all the internal calls can just require MediaInterface in the signature.

docroot/profiles/custom/cgov_site/modules/custom/cgov_image/tests/src/Kernel/CgovImageFieldStorageTest.php
```diff
@@ -126,6 +126,7 @@ public function testFields() {
     // Test the field persistance when it has been removed from all nodes.
     foreach ($this->fieldsToTest as $fieldToTest) {
       // Remove from our content type.
+      /** @var \Drupal\Core\Entity\EntityInterface $field */
```
bryanpizzillo, line 129: See my comment about patching core.

docroot/profiles/custom/cgov_site/modules/custom/cgov_core/tests/src/Kernel/FieldStorage/CGovFieldStorageTest.php
bryanpizzillo: Let's roll a local .patch to the core that fixes this type issue. (And any others we find.)

This way if the specific type changes in a 10.x branch we don't need to go fix it.

docroot/profiles/custom/cgov_site/modules/custom/cgov_core/src/Plugin/Block/DisqusComments.php
```diff
@@ -118,12 +118,11 @@ private function getCurrEntity() {
   /**
    * Create a new node storage instance.
    *
-   * @return Drupal\Core\Entity\EntityStorageInterface
+   * @return \Drupal\Core\Entity\EntityStorageInterface
```
bryanpizzillo, line 121: this should have an `|null` at the end if it can be null. Otherwise update the comment.

However, getStorage will not return a null. So please update `The node storage or NULL.` to just be `The node storage.`

docroot/profiles/custom/cgov_site/modules/custom/cgov_core/src/Plugin/Block/SectionNav.php
```diff
@@ -81,14 +81,15 @@ protected function blockAccess(AccountInterface $account) {
    *   Multidimensional section nav.
    */
   public function getSectionNav() {
+    $renderTree = [];
```
bryanpizzillo, line 84: This logic change will change the behavior of the following calling code. This should be setting $renderTree to NULL and updating the `@return` statement to match.

```
    $navTree = $this->getSectionNav();
    if ($navTree) {
      $build = [
        '#type' => 'block',
        '#nav_tree' => $navTree,
      ];
    }
    return $build;
```

bryanpizzillo (changes requested):
Here a few items both big and small to work on tomorrow until I can get to the other files.

bryanpizzillo (changes requested):
Here a few items both big and small to work on tomorrow until I can get to the other files.

docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.module
```diff
@@ -256,6 +267,9 @@ function cgov_core_entity_view_alter(array &$build, EntityInterface $entity, Ent
  */
 function cgov_core_views_query_alter(ViewExecutable $view, QueryPluginBase $query) {
   $view_id = $view->id();
+  if(!($query instanceof Sql)){
```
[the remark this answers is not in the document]
navsunka: `  if ($view_id === 'moderated_media') {
    foreach ($query->where as &$condition_group) {
      foreach ($condition_group['conditions'] as &$condition) {
        if ($condition['field'] === '.bundle') {
**// This if condition works in exposed filter of that view for bundle type exposed filter**
          $condition['field'] = 'media.bundle';
        }
      }
    }
    foreach ($query->orderby as &$orderby) {
      if ($orderby['field'] === '.bundle') {
**// this condition works when u do sort on bundle type in a media view**
        $orderby['field'] = 'media.bundle';
      }
    }
  }`


**Even after removing this hook query alter those operations works fine in my local.**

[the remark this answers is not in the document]
navsunka: I have removed this hook.

bryanpizzillo (changes requested):
I pushed up some changes I wanted to tweak without back and forth, including one item in this review. I did want to note the change.

The biggest thing is that you should not be modifying phpstan.neon. I would like to be the only one to modify it after discussing possible work arounds. Most should be simple based on the previous discussions. (Like checking types and such)

bryanpizzillo:
Closing this as we are approaching these a different way. Initial group of work was done in https://github.com/NCIOCPL/cgov-digital-platform/pull/3951.

Issue #3981, opened 2023-07-28
Fix field_ncids_theme to be used for other components


dev-rana-publicis:
### Issue description
The current `field_ncids_theme` storage config is setting `Dark` to `nci-promo-block--dark` which will work for NCIDS Promo Block but not NCIDS Hero with Dark Theme. This ticket is to make the field storage work for both components.

SupriyaSaxena0605:
@bryanpizzillo @santwuma Verified this ticket on ODE 614 for NCIDS Promo Block and NCIDS Hero with Dark Theme. Both seems to be working fine.
NCIDS Promo block with Dark theme
![image](https://github.com/NCIOCPL/cgov-digital-platform/assets/31008800/6a93d92a-3add-48d7-9c81-0ecf1d19a9cc)

NCIDS Hero with Dark theme
![image](https://github.com/NCIOCPL/cgov-digital-platform/assets/31008800/05343820-4559-4f9e-9eeb-7625f0069c65)

alinai:
Looks good on dev-acsf. The "Theme" dropdown under NCIDS Promo Block Internal and External and NCIDS Hero shows a new option "DON'T USE" along with "Dark" and "Light" options. Confirmed with @bryanpizzillo that it is expected. cc: @SupriyaSaxena0605

Issue #3949, opened 2023-07-06
Update Link Button Text field to allow 100 characters


sarinapadilla:
### Issue description
The (Override) Link Button Title field on Internal and External Link Buttons currently has a 50 character limit. This should be updated to allow for 100 characters.

### What's the expected change?
- The (Override) Link Button Title field allows 100 characters.

### What's the current functionality?
- The (Override) Link Button Title field allows 50 characters.

### What's the updated acceptance criteria?
```gherkin
Scenario: A content editor adds an override link button title to a link button
Given the content editor has added a component with a link button (e.g. Guide Card with Image and Description)
  And the content editor has added a link button to the Guide Card with Image and Description
  And the content editor is overriding the link button title
Then the content editor is able to add an override title with 100 characters
```

alinai:
ON ODE614, under NCIDS Guide Card 2 and 3, we noticed that the "Override Link Button Title" for NCIDS Link Button Internal link and "Link Button Title" for External Link both allow 100 characters with spaces. 
@sarinapadilla , @bryanpizzillo. @dev-rana-publicis , please confirm that the 100 character limit is expected for both Internal and External Link fields. cc: @SupriyaSaxena0605 , @santwuma

alinai:
@bryanpizzillo , confirmed with @dev-rana-publicis , that all of the following works as expected on ODE614 for #3978 
@dev-rana-publicis  confirmed that NCIDS Link Button Internal and External only work with the following paragraphs:
Guide Card
Guide Card with Image and Description
Call to Action Strip
Hero (w/ button, w/ CTA Strip)

Verified that #3978 works for all of the above.

alinai:
Looks good on dev-acsf.

PR #4007, opened 2023-08-16
(#4005) Fix Config Issues

### LINKED ISSUE #4005: Changed Configs in NCIDS Homepage Items Break drupal:update
There are 2 issues (so far):
1. The change to `field_ncids_theme.yml` from #3981 cannot be applied because `A list field 'field_ncids_theme' with existing data cannot have its keys changed.` This means that we need to add an entry back in with `value: nci-promo-block--dark`.
2. More troubling, the change in #3949 fails because `The SQL storage cannot change the schema for an existing field (field_button_text in paragraph entity) with data.` This will need an update hook, https://evolvingweb.com/blog/resizing-fields-drupal-without-losing-data has some good steps.

bryanpizzillo:
- We cannot remove any list field options that exist in data. (field_ncids_theme)
- Fixes the issue with the schema change for field_button_text.

ODE: https://ncigovcdode614.prod.acquia-sites.com/
Closes #4005

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_home_landing/cgov_home_landing.install
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_home_landing/cgov_home_landing.install
@@ -1,5 +1,7 @@
 <?php
 
+use Drupal\field\Entity\FieldStorageConfig;
+
 /**
  * @file
  * Contains cgov_home_landing.install.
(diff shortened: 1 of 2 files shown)


alinai:
@bryanpizzillo , both https://github.com/NCIOCPL/cgov-digital-platform/issues/3981 and https://github.com/NCIOCPL/cgov-digital-platform/issues/3949 looks good on the ODE614.

PR #4034, opened 2023-08-31
(#4033) Do not enable memcache on ODEs

### LINKED ISSUE #4033: Enabler: Disable memcache on ODEs
# Do not enable the memcache module when installing on an ODE.

ODEs become slow after some usage.  This _may_ be due to an issue with Acquia's memcache configuration.

We MUST have memcache on ACSF, so this ticket is about disabling it only when installing on an ODE.

**Approach**

We're going to patch the `memcache.settings.php` file (**Important:** According to `composer.lock`, we're currently using v1.1.0.).  

Near the top of the file, the script checks whether the memcache module is present and if so, it loads it (optionally setting it to use the memcached daemon).

The patch will add a condition, checking that the `AH_SITE_ENVIRONMENT` doesn't reflect an ODE.

> **ESTIMATE** TBD

### Resources:

- [memcache.settings.php](https://github.com/acquia/memcache-settings/blob/v1.1.0/memcache.settings.php)

### Notes

- The `memcache.settings.php` file comes in when the `composer install` step runs, so the patch will be applied by composer.
- The patch goes in `~/patches/acquia`, and then, the name should likely be `memcache.settings.php.patch`
- How to detect an ODE
From `docroot/sites/settings/cgov_caching.settings.php`

    ```php
      $env = $_ENV['AH_SITE_ENVIRONMENT'];
      if (preg_match('/^ode\d*$/', $env)) {
        $env = 'ode';
      }
    ```

### Testing Notes

There should be no changes in functionality.  The ODE may run slower, but it should no longer become completely unusable.

blairlearn:
Closes #4033

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
         "acquia/blt-drupal-test": "^1.0",
         "acquia/blt-phpcs": "^1.0",
         "acquia/blt-simplesamlphp": "^1.0",
-        "acquia/memcache-settings": "^1.0",
+        "acquia/memcache-settings": "^1.2",
         "brianium/paratest": "^6.3",
         "commerceguys/addressing": "^1.4",
         "composer/installers": "^1.9.0",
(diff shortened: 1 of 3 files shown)


SupriyaSaxena0605:
@Blair @Sandra T. All automated regression tests passed on ODE 627. We could create and verify Article successfully without slowness or ODE crashing. Thanks @Blair for fixing ODEs.

KateMashkinaNIH:
Hit ODE with two full regression runs and environment is still functioning

SupriyaSaxena0605:
@blairlearn Full regression passed on ODE 627 without breaking it.

navsunka (approved):

PR #4130, opened 2023-10-25
(#4129) Activate existing blocked users on import

### LINKED ISSUE #4129: Standard test user accounts blocked on lower tiers
With the changes in #4120 we blocked all of the test user accounts in PROD before we update SAML authentication. But now they are being copied down to lower tiers in blocked status. The bulk user load needs to unblock them if they exist.

Testing steps: You can log in as the standard user accounts on ACSF lower tiers after a refresh.

jfrank-nih:
Closes #4129


blairlearn (approved):

jfrank-nih:
Putting in to #4132 per request from @bryanpizzillo.

PR #4132, opened 2023-10-26
(#4131) Enforce SAML authentication

### LINKED ISSUE #4129: Standard test user accounts blocked on lower tiers

### LINKED ISSUE #4131: Enforce samlauth entry for all user accounts
## Requirement

**Key:** No one can log into production without SSO, unless there are exceptional circumstances, in which case the admin user, and only the admin, can be unblocked and given a good password, and they can login

**Devired:**

- All new users should have an entry added to the `samlauth` upon creation
    - This includes users created via form
    - This includes user created via the bulk user import

This will effectively force users to have to use SAML, based on the `samlauth` requirements.

## Testing

- Create a new user account, giving it a Drupal password
- Attempt to log in on the Drupal login form (not SSO) with that username/password
- Expected: Can't log in
- Actual: Can log in

jfrank-nih:
- Adds new user accounts to authmap table
- One time update to all users to ensure they are in authmap
- Unblocks user accounts on copy-down

Closes #4131
Closes #4129

### DIFF
--- a/cgov-drupal-users.yml
+++ b/cgov-drupal-users.yml
@@ -7,12 +7,14 @@ admin:
 additional_users:
   - username: 'author'
     password: 'password123'
+    saml: false
     roles:
       - 'content_author'
       - 'image_manager'
       - 'admin_ui'
   - username: 'editor'
     password: 'password123'
+    saml: false
     roles:
       - 'content_author'
       - 'content_editor'
@@ -20,6 +22,7 @@ additional_users:
       - 'admin_ui'
   - username: 'adveditor'
     password: 'password123'
+    saml: false
     roles:
       - 'content_author'
       - 'content_editor'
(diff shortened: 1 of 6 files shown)


bryanpizzillo (approved):
Code looks good from a style and nothing weird jumping out. So if it passes QA, it is good to merge.

SupriyaSaxena0605:
@jfrank-nih @santwuma Full regression as well visual regression passed on test-ac. Verified #4131 and #4129 while comparing user creation between test-ac and int-ac.

PR #4145, opened 2023-10-31
(#4144) Fixes redirect manager test random failures

### LINKED ISSUE #4144: Redirect test fails when temporary filename has special characters
### Issue description
If the random string generated in the test for the filename has special characters it can cause the save to fail, and thus the test.

````
1) Drupal\Tests\cgov_redirect_manager\Functional\CgovImporterServiceTest::testRedirectManager
Drupal\Core\File\Exception\DirectoryNotReadyException: The specified file 'temporary://fileNPcLRx' could not be copied because the destination directory 'temporary://:*' is not properly configured. This may be caused by a problem with file or directory permissions.
````

> Drupal\Tests\RandomGeneratorTrait::randomString
> Generates a pseudo-random string of ASCII characters of codes 32 to 126.
> Do not use this method when special characters are not possible (e.g., in machine or file names that have already been validated); instead, use \Drupal\simpletest\TestBase::randomMachineName(). If $length is greater than 3 the random string will include at least one ampersand (‘&’) and at least one greater than (‘>’) character to ensure coverage for special characters and avoid the introduction of random test failures.
> @param int $length — Length of random string to generate.
> @return string
> Pseudo-randomly generated unique string including special characters.

Fix:

jfrank-nih:
Closes #4144

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_redirect_manager/tests/src/Functional/CgovImporterServiceTest.php
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_redirect_manager/tests/src/Functional/CgovImporterServiceTest.php
@@ -67,7 +67,7 @@ class CgovImporterServiceTest extends BrowserTestBase {
 CSV;
 
     // Get random name for csv file and create it.
-    $file_name = $this->randomString() . '.csv';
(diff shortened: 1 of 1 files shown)


bryanpizzillo (approved):
Thanks for figuring this one out quickly.

PR #4149, opened 2023-10-31
(#4148) Suppress undefined variable samlDir on install

### LINKED ISSUE #4148: Undefined variable during local installation
During local installation there is a warning that `$samlDir` is undefined in `cgov_saml_auth_config.settings.php`. This happens if your Docker container has not been recreated to have the `/var/saml` directory. Code should be cleaner to not look for it.

jfrank-nih:
Closes #4148


jfrank-nih:
Putting into #4132

PR #4177, opened 2023-11-13
(#4171, #4176, #4182) Remove drush & blt launchers, remove drupal console.

### LINKED ISSUE #4171: Drush 12 does not support drush launcher.
#### Container

In our container we use [drush-launcher](https://github.com/drush-ops/drush-launcher), which basically tries to find the closest drush to your repo root. So if you are buried 10 levels deep in the front-end theme, a `drush cr` is going to work. With drush 12 this no longer works. 

Since our container only has 1 codebase, we can probably just add `/var/www/vendor/bin` to the `$PATH` and call it a day. This will also put BLT on the path too... This however will not fix the problem elsewhere, but this ticket is not for that.

#### Cloud hooks (`hooks/common/**` structure)

The cloud hooks (e.g. hooks/common/post-code-deploy/post-code-deploy.sh) modify the path to include vendor/bin. No modification is needed.

#### Factory hooks (`factory-hooks/**` structure)

The factory hooks (e.g. post-site-install.sh) invoke blt  directly from the vendor directory in which they were installed. (i.e. `blt="/mnt/www/html/$sitegroup.$env/vendor/acquia/blt/bin/blt"`).

However, the scripts in `vendor/bin` are nothing more than proxies which invoke the scripts from their installation directories. (i.e. the blt script is a PHP file which executes return include __DIR__ . '/..'.'/acquia/blt/bin/blt';

#### CI workfklow

In the CI workflow, it is necessary to add the vendor/bin directory to the path.

In the "Test Ability to Deploy" workflow, it is necessary to add the vendor/bin directory to the path. Additionally, it is also necessary to modify the "Does this workflow need to run?" logic to check whether the workflow itself has changed.

### LINKED ISSUE #4176: Remove drupal/console from Docker build
### Issue description

Remove the drupal console from the Docker build. As noted in #4140, we don't really use it.

Additionally, as of this date, the download for the tool's installer returns a status 500 error.

### LINKED ISSUE #4182: Allow test-deployability workflow to run when it changes
### Issue description
The "deployability test" workflow (`.github/workflows/test-deployability.yml`) doesn't execute when it's changed.  This prevents it from being tested.

Change the workflow run criteria to include the workflow itself.  This means changing the paths list for:
- The `on.push` event.
- The `on.pull_request` event.
- The hash file list (`install_files_hash` step  in the `check_for_past_run` job)

blairlearn:
Closes #4171 
Closes #4176 
Closes #4182 

Standard regression tests only.
[ODE 658](http://ncigovcdode658.prod.acquia-sites.com/)

### DIFF
--- a/docker/web/Dockerfile
+++ b/docker/web/Dockerfile
@@ -127,15 +127,12 @@ RUN npm install -g npm@8.1.0
 # Install composer
 RUN curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/local/bin --filename=composer --2
 
-## Install Drush Launcher (runs local drush instances)
-RUN curl -sL -o /usr/local/bin/drush https://github.com/drush-ops/drush-launcher/releases/download/0.10.1/drush.phar \
-  && chmod +x /usr/local/bin/drush
+# Verify php and composer installs.
+RUN php --version; composer --version;
 
-RUN php --version; composer --version; drush --drush-launcher-version
-
-## Enable BLT function
-RUN curl -sL -o /usr/local/bin/blt https://github.com/acquia/blt-launcher/releases/download/v1.1.0/blt.phar \
-  && chmod +x /usr/local/bin/blt
+# Make vendor/bin versions of drush, blt, etc. available.
+# NOTE: Putting this on the end in order to match the CI workflow.
+ENV PATH $PATH:/var/www/vendor/bin
 
 ## Turn off xdebug as it uses resources that could be best used
 ## elsewhere. Once the container is up you can use
--- a/docker/web/Dockerfile
+++ b/docker/web/Dockerfile
@@ -131,11 +131,7 @@ RUN curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/local
 RUN curl -sL -o /usr/local/bin/drush https://github.com/drush-ops/drush-launcher/releases/download/0.10.1/drush.phar \
   && chmod +x /usr/local/bin/drush
 
-## Install Drupal console
-RUN curl https://drupalconsole.com/installer -L -o /usr/local/bin/drupal \
-  && chmod +x /usr/local/bin/drupal
-
-RUN php --version; composer --version; drupal --version; drush --drush-launcher-version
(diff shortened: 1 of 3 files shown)


docker/web/Dockerfile
```diff
@@ -127,19 +127,14 @@ RUN npm install -g npm@8.1.0
...
-RUN curl -sL -o /usr/local/bin/drush https://github.com/drush-ops/drush-launcher/releases/download/0.10.1/drush.phar \
-  && chmod +x /usr/local/bin/drush
+# Verify php and composer installs.
+RUN php --version; composer --version;
 
-## Install Drupal console
-RUN curl https://drupalconsole.com/installer -L -o /usr/local/bin/drupal \
-  && chmod +x /usr/local/bin/drupal
+# Make vendor/bin versions of drush, blt, etc. available.
+ENV PATH /var/www/vendor/bin/:$PATH
```
bryanpizzillo, line 134: Can you just add this to the end of the path so it matches the CI workflow?

docker/web/Dockerfile
```diff
@@ -127,19 +127,14 @@ RUN npm install -g npm@8.1.0
...
 
-RUN php --version; composer --version; drupal --version; drush --drush-launcher-version
-
-## Enable BLT function
-RUN curl -sL -o /usr/local/bin/blt https://github.com/acquia/blt-launcher/releases/download/v1.1.0/blt.phar \
-  && chmod +x /usr/local/bin/blt
+# Favor the downloaded composer over the one in /vendor/bin.
+ENV PATH /usr/local/bin:$PATH
```
bryanpizzillo, line 137: Get rid of this after changing the above.

bryanpizzillo (changes requested):
Got 1 small thing.

docker/web/Dockerfile
blairlearn: Done

docker/web/Dockerfile
blairlearn: Bonus, not needing to add a path entry that already exists.

bryanpizzillo (approved):

PR #4179, opened 2023-11-15
(#4170) New PHPCS issues with D10

### LINKED ISSUE #4170: New PHPCS issues with D10
These all look to be the same and could be fixed prior to d10. We should see if this is just a rule (`SlevomatCodingStandard.Namespaces.AlphabeticallySortedUses.IncorrectlyOrderedUses`) we can turn on so we are no just hoping this will be fixed. 

The errors are too long for github.

navsunka:
Closes #4170


bryanpizzillo:
This was committed as part of #4181

PR #4222, opened 2024-01-10
(#3802) Updated NCIDS to not use local

### LINKED ISSUE #3802: Enabler: Stop using local @nciocpl packages
# Stop using local @nciocpl packages
> **ESTIMATE** TBD

Now that we are using Github actions, we now have npm tokens available to us in the workflow. This means we do not have to download the packages locally. This will make it a bit easier to update ncids and other packages. We already have instructions to pass in your own .npmrc that contains the local token, so there should be no process changes there. (Just a lot more questions across the team on why rebuild-feq is not working for them because they never set it up.)

R4R is not being updated as it is going to move to a generic app module at some point and will not be in our code base at all.

## Testing
- Regression testing only

bryanpizzillo:
ODE: http://ncigovcdode673.prod.acquia-sites.com/AH_VIEW

Closes #3802


bryanpizzillo:
Just adding the note here. Updating our nciocpl eslint packages caused 1336 issues. So some dependency of our packages got bumped and now trigger linting issues.

KateMashkinaNIH:
Visual regression and FE tests passed

arcepaul (approved):

PR #4225, opened 2024-01-16
(#4209, #4208, #4210) Refactor blt cgov:reinstall

### LINKED ISSUE #4208: Enabler: Split BLT reinstall command in two
Splitting the current `blt cgov:reinstall` into two, and changing it into a combined command.

* `blt cgov:install-drupal`
* `blt cgov:install-content`
* `blt cgov:reinstall` 

Here is the current command:

https://github.com/NCIOCPL/cgov-digital-platform/blob/0867ed128618e3a78298e7d87fbac5c897fe99f3/blt/src/Blt/Plugin/Commands/SlevenCommands.php#L13-L32

The `install-drupal` command should have:

* Drupal install
* Load users
* Translate locales

The `install-content` content command should have the remaining commands:

* Install content
* Load front-end globals
* Load glossifier

The `reinstall` should just call the other two in order to achieve a site installed with content.

## Testing

* Regressions 
* For devs: run `cgov:reinstall` locally and verify you get the same content

## Notes

### LINKED ISSUE #4209: Enabler: Update cgov:reinstall to use artifact:install:drupal
## Change

Update `blt cgov:reinstall` to use `artifact:install:drupal` instead of `drupal:install`

https://github.com/NCIOCPL/cgov-digital-platform/blob/0867ed128618e3a78298e7d87fbac5c897fe99f3/blt/src/Blt/Plugin/Commands/SlevenCommands.php#L21

Make sure that reinstalling on local dev works as expected.

## Testing

- Standard regression tests

## Background

> ODEs use artifact:install:drupal and local uses drupal:install. They both are, for all intents and purpose, doing the same thing. The only difference is if we enable the multisite flag on an ODE. If an ODE uses multisite, then the local environment should as well.
> When might we use multisite on an ODE and local? If we had to have 2 profiles or themes. Back when we were working on the redesign we had a ticket for figuring out to to have 2 installed sites, one with ncids_trans and the other with cgov_common . However, we changed the scope such that only DCEG Connect was on cgov_common and dropped the ticket. So it is a thing.
> Isn't ACSF multisite? Not for the purposes of BLT.
> TODO:
> I want to say change cgov:reinstall to call artifact:install:drupal so that if we have multisite in the future we don't have to change the command then. (more so, so we don't have to wonder why it does not work when we change the blt.yml)

### LINKED ISSUE #4210: Enabler: Update post code deploy hooks to use cgov:reinstall
## Change

In the [post-code-deploy](https://github.com/NCIOCPL/cgov-digital-platform/blob/develop/hooks/common/post-code-deploy/post-code-deploy.sh) and [post-code-update](https://github.com/NCIOCPL/cgov-digital-platform/blob/develop/hooks/common/post-code-update/post-code-update.sh) scripts, we need to remove the multiple commands for installing content and instead update with just `blt cgov:reinstall`.

These are the lines:

https://github.com/NCIOCPL/cgov-digital-platform/blob/4977e65d837f5e807376b7001e1a08104b142d51/hooks/common/post-code-update/post-code-update.sh#L61-L77

All of the commands being done there should be found within:

https://github.com/NCIOCPL/cgov-digital-platform/blob/0867ed128618e3a78298e7d87fbac5c897fe99f3/blt/src/Blt/Plugin/Commands/SlevenCommands.php#L18-L29

## Sequencing

After: #4209 

## Testing

- Standard regression tests on ODE

blairlearn:
Closes #4208 
Closes #4209 
Closes #4210

### DIFF
--- a/blt/src/Blt/Plugin/Commands/SlevenCommands.php
+++ b/blt/src/Blt/Plugin/Commands/SlevenCommands.php
@@ -18,7 +18,7 @@ class SlevenCommands extends BltTasks {
   public function reinstall() {
 
     $commands = [
-      'drupal:install' => [],
+      'artifact:install:drupal' => [],
       'cgov:user:load-all' => [],
       'cgov:locales:translate' => [],
       'custom:install_cgov_yaml_content_by_module' => [
--- a/blt/src/Blt/Plugin/Commands/SlevenCommands.php
+++ b/blt/src/Blt/Plugin/Commands/SlevenCommands.php
@@ -18,9 +18,37 @@ class SlevenCommands extends BltTasks {
   public function reinstall() {
 
     $commands = [
+      'cgov:install-drupal' => [],
+      'cgov:install-content' => [],
+    ];
+
+    $this->invokeCommands($commands);
+  }
+
+  /**
+   * Install the digital platform, without content.
+   *
+   * @command cgov:install-drupal
+   */
+  public function installDrupal() {
+
+    $commands = [
       'artifact:install:drupal' => [],
       'cgov:user:load-all' => [],
       'cgov:locales:translate' => [],
+    ];
+
+    $this->invokeCommands($commands);
+  }
+
+  /**
+   * Install the standard test content.
+   *
+   * @command cgov:install-content
+   */
+  public function installContent() {
+
+    $commands = [
       'custom:install_cgov_yaml_content_by_module' => [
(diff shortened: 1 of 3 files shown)


jfrank-nih (approved):

jfrank-nih:
Code review is good to go.

SupriyaSaxena0605:
@blairlearn @kate-mashkina Regression Tests passed on ODE675

PR #4230, opened 2024-01-22
(#4211) Enabler: Block Admin in cgov:user:load-all

### LINKED ISSUE #4211: Enabler: Block Admin in cgov:user:load-all
## Change

Update the BLT [`cgov:user:load-all` function](https://github.com/NCIOCPL/cgov-digital-platform/blob/4977e65d837f5e807376b7001e1a08104b142d51/blt/src/Blt/Plugin/Commands/UsersCommands.php#L19-L23) to block the administrator user by default. The command [`cgov:block-admin-user`](https://github.com/NCIOCPL/cgov-digital-platform/blob/4977e65d837f5e807376b7001e1a08104b142d51/drush/Commands/cgov/CgovUserCommands.php#L62) can be used to block the admin.

It should block the administrator user if:

- There is no `admin` section in the file
- There is an `admin` section in the file but the key `enabled` is not set to true

Items that will need to be changed:

1. The [loadUsers function](https://github.com/NCIOCPL/cgov-digital-platform/blob/4977e65d837f5e807376b7001e1a08104b142d51/blt/src/Blt/Plugin/Commands/UsersCommands.php#L38C31-L38C31) should block the admin if the `admin` array key does not exist. The message can be updated to indicate that admin has been blocked.

2. The [isAdminBlockValid check](https://github.com/NCIOCPL/cgov-digital-platform/blob/4977e65d837f5e807376b7001e1a08104b142d51/blt/src/Blt/Plugin/Commands/UsersCommands.php#L67) should be updated to look for the `enabled` key as a boolean. _**Edit 1/23:**_ We want to check that the key is a boolean _if it exists_. Not providing the key is valid but in that case the value of the "enabled" key should default to `false`.

3. The [drush command `cgov:change-admin-user`](https://github.com/NCIOCPL/cgov-digital-platform/blob/4977e65d837f5e807376b7001e1a08104b142d51/drush/Commands/cgov/CgovUserCommands.php#L31) should also be updated to block the admin user if the `enabled` key is `false`.

4. The [isAdminValid check](https://github.com/NCIOCPL/cgov-digital-platform/blob/4977e65d837f5e807376b7001e1a08104b142d51/drush/Commands/cgov/CgovUserCommands.php#L95) for drush should also be updated to expect the key as a boolean. _**Edit 1/23:**_ Like above, no key provided is valid and would be assumed to be `false`.

The admin section should look like this in all files in the future, except on PROD:

```
admin:
    username: 'admin'
    password: 'my-password'
    enabled: true
```

_**Edit 1/23**_: But this is also valid:

```
admin:
    username: 'admin'
    password: 'my-password'
```

And the above would be the same as this:

```
admin:
    username: 'admin'
    password: 'my-password'
    enabled: false
```

## Post-Update

Post update we have to update the YAML file on _all_ lower tier environments if we want the admin user to be available on them.

### One
- Without updating an ODE server's YAML file, create a new ODE
- Deploy new code to the ODE
- Verify that the admin user account is blocked and can't be logged into
- 
### Two
- After ODE creation but before deploying new code, update the ODE server's users YAML file to have no admin section
- Deploy new code to the ODE
- Verify that the admin user account is blocked and can't be logged into

### Three
- After ODE creation but before deployment, update the ODE server's users YAML file to have the admin section with `enabled: true`
- Create a new site in that tier
- Verify that the admin user account is unblocked and can be logged into
- Run standard regressions

navsunka:
- block the admin user using key in loadUsers function.
- updated the post-install hook as we are handling block-admin in loadUsers function.

Closes #4211

### DIFF
--- a/blt/src/Blt/Plugin/Commands/UsersCommands.php
+++ b/blt/src/Blt/Plugin/Commands/UsersCommands.php
@@ -34,16 +34,29 @@ class UsersCommands extends BltTasks {
     }
 
     $commands = [];
-
+    $blocked_admin = [
+      'cgov:acsf:block-admin' => [],
+    ];
     if (array_key_exists('admin', $userConfig)) {
       if ($this->isAdminBlockValid($userConfig['admin'])) {
-        $commands['cgov:user:reset-admin'] = [
-          'users_file_path' => $users_file_path,
-        ];
+        if ($userConfig['admin']['enabled'] === TRUE) {
+          $commands['cgov:user:reset-admin'] = [
+            'users_file_path' => $users_file_path,
+          ];
+        }
+        else {
+          $commands = $blocked_admin;
+          $this->logger->info("Admin user is not explicitly enabled and has been blocked.");
+        }
       }
       else {
-        $this->logger->warning("Skipping admin user. You must provide both the username and the password.");
+        $commands = $blocked_admin;
+        $this->logger->warning("Skipping admin user. You must provide both the username and password and set enable to either true or false.");
       }
+    }
+    else {
+      $commands = $blocked_admin;
+      $this->logger->info("No admin user provided so admin user has been blocked.");
     }
 
     if (array_key_exists("additional_users", $userConfig)) {
@@ -66,12 +79,26 @@ class UsersCommands extends BltTasks {
    */
   protected function isAdminBlockValid($admin) {
     // @todo This should be more robust and check bad chars.
-    return (
-      array_key_exists('username', $admin) &&
-      $admin['username'] != '' &&
-      array_key_exists('password', $admin) &&
-      $admin['password'] != ''
-    );
+    // Validate username and password.
+    $userAndPasswordValid = FALSE;
+    if (array_key_exists('username', $admin) &&
+        $admin['username'] != '' &&
+        array_key_exists('password', $admin) &&
+        $admin['password'] != '') {
+      $userAndPasswordValid = TRUE;
+    }
+
+    // If present, the enabled key must be boolean.
+    // It is also valid if the key is absent.
+    $enabledSectionValid = FALSE;
+    if (array_key_exists('enabled', $admin) && is_bool($admin['enabled'])) {
+      $enabledSectionValid = TRUE;
+    }
+    elseif (!array_key_exists('enabled', $admin)) {
+      $enabledSectionValid = TRUE;
+    }
+
+    return $userAndPasswordValid && $enabledSectionValid;
   }
 
   /**
--- a/drush/Commands/cgov/CgovUserCommands.php
+++ b/drush/Commands/cgov/CgovUserCommands.php
@@ -2,8 +2,8 @@
 
 namespace Drush\Commands\cgov;
 
-use Drupal\user\Entity\User;
 use Drupal\user\Entity\Role;
+use Drupal\user\Entity\User;
 use Drush\Commands\DrushCommands;
 use Symfony\Component\Yaml\Yaml;
 
@@ -35,7 +35,7 @@ class CgovUserCommands extends DrushCommands {
    * @bootstrap full
    */
   public function changeAdminUserInfo($usersFile) {
-    // TODO: Relative file paths do not work here. Make them work.
+    // @todo Relative file paths do not work here. Make them work.
     if (!file_exists($usersFile)) {
       throw new \Exception(dt('Unable to load file !file.', ['!file' => $usersFile]));
     }
@@ -49,13 +49,20 @@ class CgovUserCommands extends DrushCommands {
     }
 
     if (!$this->isAdminValid($userConfig['admin'])) {
+      $this->blockAdminUser();
+      $this->logger->debug('Admin has been blocked.');
       throw new \Exception(dt('Invalid config !file.', ['!file' => $usersFile]));
     }
 
     $user = $userConfig['admin']['username'];
     $pass = $userConfig['admin']['password'];
-
-    $this->resetAdmin($user, $pass);
+    if ($userConfig['admin']['enabled'] === TRUE) {
+      $this->resetAdmin($user, $pass);
+    }
+    else {
+      $this->blockAdminUser();
+      $this->logger->debug('Admin has been blocked.');
+    }
   }
 
   /**
@@ -93,12 +100,25 @@ class CgovUserCommands extends DrushCommands {
    *   TRUE if valid, FALSE if not.
    */
   protected function isAdminValid($admin) {
-    return (
-      array_key_exists('username', $admin) &&
-      $admin['username'] != '' &&
-      array_key_exists('password', $admin) &&
-      $admin['password'] != ''
-    );
+    // Validate username and password.
+    $userAndPasswordValid = FALSE;
+    if (array_key_exists('username', $admin) &&
+        $admin['username'] != '' &&
+        array_key_exists('password', $admin) &&
+        $admin['password'] != '') {
+      $userAndPasswordValid = TRUE;
+    }
+    // If present, the enabled key must be boolean.
+    // It is also valid if the key is absent.
+    $enabledSectionValid = FALSE;
+    if (array_key_exists('enabled', $admin) && is_bool($admin['enabled'])) {
+      $enabledSectionValid = TRUE;
+    }
+    elseif (!array_key_exists('enabled', $admin)) {
+      $enabledSectionValid = TRUE;
+    }
+
+    return $userAndPasswordValid && $enabledSectionValid;
   }
 
   /**
@@ -157,7 +177,7 @@ class CgovUserCommands extends DrushCommands {
    */
   public function bulkLoadUsers($usersFile) {
 
-    // TODO: Relative file paths do not work here. Make them work.
+    // @todo Relative file paths do not work here. Make them work.
     if (!file_exists($usersFile)) {
       throw new \Exception(dt('Unable to load file !file.', ['!file' => $usersFile]));
     }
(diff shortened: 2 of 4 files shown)


blt/src/Blt/Plugin/Commands/UsersCommands.php
```diff
@@ -70,7 +76,8 @@ protected function isAdminBlockValid($admin) {
       array_key_exists('username', $admin) &&
       $admin['username'] != '' &&
       array_key_exists('password', $admin) &&
-      $admin['password'] != ''
+      $admin['password'] != '' &&
+      array_key_exists('enabled', $admin)
```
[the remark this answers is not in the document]
blairlearn: I've also chatted with @jfrank-nih about the definition of "valid".  He's going to update the ticket for clarity, but this block (the old structure) should also be considered valid.

```yaml
admin:
  username: 'admin'
  password: 'password123'
```

In this case, the `enabled` key is missing and should be treated as though it was set to `false`.

blt/src/Blt/Plugin/Commands/UsersCommands.php
```diff
@@ -42,9 +42,15 @@ public function loadUsers() {
...
-        $this->logger->warning("Skipping admin user. You must provide both the username and the password.");
+        $this->logger->warning("Skipping admin user. You must provide both the username, password & enable key value.");
       }
     }
+    else {
+      $commands = [
+        'cgov:acsf:block-admin' => [],
+      ];
+      $this->logger->warning("Admin has been blocked.");
+    }
```
blairlearn, line 53: There is no check for whether the `admin` user should actually be enabled. At present, this code will enable the user as long as YAML data structure is set, regardless of the value set in the `enabled` key.

The pseudocode should be

```
if the admin key exists {
  if the admin block is valid {
    if the enabled key is present AND set to true {
      enable the admin user
    }
    else {
      disable the admin user.
    }
  }
  else {
    disable the admin user.
    log a warning.
  }
}
else {
  disable the admin user.
}
```

For the messages....

On the warning message, let's tweak the text to say 
"Skipping admin user. You must provide both the username and password and set enable to either true or false."

The other cases are all normal operation and at most should be logged as `debug` messages, but **only** if they won't normally be output.  (Our BLT output is already quite noisy.)

blairlearn (changes requested):

blt/src/Blt/Plugin/Commands/UsersCommands.php
```diff
@@ -70,7 +87,9 @@ protected function isAdminBlockValid($admin) {
       array_key_exists('username', $admin) &&
       $admin['username'] != '' &&
       array_key_exists('password', $admin) &&
-      $admin['password'] != ''
+      $admin['password'] != '' &&
+      array_key_exists('enabled', $admin) &&
+      is_bool($admin['enabled'])
```
blairlearn, line 92: This condition isn't quite right yet.

The `username` and `password` keys are both required.

If the `enabled` key is **NOT** present, the block is still valid.

If the `enabled` key **IS** present, the block is only valid if the value of `enabled` is a boolean.

Break the checks into multiple steps. That should make it clearer what needs to happen.

drush/Commands/cgov/CgovUserCommands.php
```diff
@@ -49,13 +49,20 @@ public function changeAdminUserInfo($usersFile) {
...
     $user = $userConfig['admin']['username'];
     $pass = $userConfig['admin']['password'];
-
-    $this->resetAdmin($user, $pass);
+    if ($userConfig['admin']['enabled'] === TRUE) {
+      $this->resetAdmin($user, $pass);
+    }
+    else {
+      $this->blockAdminUser();
+      $this->logger->info('Admin has been blocked.');
```
blairlearn, line 64: Please change this to `debug` instead of `info`.

blairlearn (changes requested):

blt/src/Blt/Plugin/Commands/UsersCommands.php
```diff
@@ -66,12 +83,17 @@ public function loadUsers() {
...
+    if (array_key_exists('username', $admin) &&
+        $admin['username'] != '' &&
+        array_key_exists('password', $admin) &&
+        $admin['password'] != '') {
+      $flag = TRUE;
+    }
+    if (array_key_exists('enabled', $admin) && !is_bool($admin['enabled'])) {
+      $flag = FALSE;
+    }
+    return $flag;
```
blairlearn, line 96: Let's make this a bit simpler to read...

```php
  protected function isAdminBlockValid($admin) {
    // @todo This should be more robust and check bad chars.
    // Validate username and password.
    $userAndPasswordValid = FALSE;
    if (array_key_exists('username', $admin) &&
        $admin['username'] != '' {
      $userAndPasswordValid = TRUE;
    }

    // If present, the enabled key must be boolean.
    // It is also valid if the key is absent.
    $enabledSectionValid = FALSE;
    if (array_key_exists('enabled', $admin) && is_bool($admin['enabled'])) {
      $enabledSectionValid = TRUE;
    }
    else if (!array_key_exists('enabled', $admin)) {
      $enabledSectionValid = TRUE;
    }
    return $userAndPasswordValid && enabledSectionValid;
  }
```

drush/Commands/cgov/CgovUserCommands.php
```diff
@@ -49,13 +49,20 @@ public function changeAdminUserInfo($usersFile) {
     }
 
     if (!$this->isAdminValid($userConfig['admin'])) {
+      $this->blockAdminUser();
+      $this->logger->info('Admin has been blocked.');
```
blairlearn, line 53: This one should also be `debug`.  Sorry for missing that on the last round.

blairlearn (changes requested):

KateMashkinaNIH:
Round One:
Unable to login as admin ( Unrecognized username or password is shown)
Successfully logged in as siteadmin, author and editor and did a sanity check on article workflow in cms

blairlearn (approved):

blt/src/Blt/Plugin/Commands/UsersCommands.php
```diff
@@ -37,14 +37,31 @@ public function loadUsers() {
...
+        if ($userConfig['admin']['enabled'] === TRUE) {
+          $commands['cgov:user:reset-admin'] = [
+            'users_file_path' => $users_file_path,
+          ];
+        }
+        else {
+          $commands = [
+            'cgov:acsf:block-admin' => [],
+          ];
+          $this->logger->info("Admin has been blocked.");
```
jfrank-nih, line 49: I don't love adding `cgov:acsf:block-admin` to the `$commands` variable three separate times. It feels a bit redundant. That said, each instance _is_ for a separate reason _so at least_ the logged messages should reflect it.

1. "Admin user is not explicitly enabled and has been blocked
2. (This ones's good) "Skipping admin user. You must provide both the username and password and set enable to either true or false."
3. "No admin user provided so admin user has been blocked"

blt/src/Blt/Plugin/Commands/UsersCommands.php
```diff
@@ -66,12 +83,25 @@ public function loadUsers() {
...
+    if (array_key_exists('username', $admin) &&
+        $admin['username'] != '' &&
+        array_key_exists('password', $admin) &&
+        $admin['password'] != '') {
+      $userAndPasswordValid = TRUE;
+    }
+    // If present, the enabled key must be boolean.
+    // It is also valid if the key is absent.
+    $enabledSectionValid = FALSE;
+    if (array_key_exists('enabled', $admin) && is_bool($admin['enabled'])) {
```
jfrank-nih, line 97: I'm not quite sure what's going on and why `is_bool` is not operating as I expect it to, because this seems valid, but this block is passing the admin check:

```yaml
admin:
  username: 'admin'
  password: 'password123'
  enabled: 'this is not a valid boolean'
```

jfrank-nih (changes requested):

blt/src/Blt/Plugin/Commands/UsersCommands.php
navsunka: I have tested this scenario and is_bool is returning false. Hence it is blocking the admin user.

blt/src/Blt/Plugin/Commands/UsersCommands.php
```diff
@@ -37,14 +37,31 @@ public function loadUsers() {
...
-        $commands['cgov:user:reset-admin'] = [
-          'users_file_path' => $users_file_path,
-        ];
+        if ($userConfig['admin']['enabled'] === TRUE) {
+          $commands['cgov:user:reset-admin'] = [
+            'users_file_path' => $users_file_path,
+          ];
+        }
+        else {
+          $commands = [
```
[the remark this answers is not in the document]
navsunka: $commands is returning array
`[
  "cgov:acsf:block-admin" => []
  "cgov:user:load-additional" => array:1 [
    "users_file_path" => "/var/www/cgov-drupal-users.yml"
  ]
]`

array2
`[
  "cgov:user:reset-admin" => array:1 [
    "users_file_path" => "/var/www/cgov-drupal-users.yml"
  ]
  "cgov:user:load-additional" => array:1 [
    "users_file_path" => "/var/www/cgov-drupal-users.yml"
  ]
]`

blt/src/Blt/Plugin/Commands/UsersCommands.php
navsunka: I have declared it above and i have changed the log messages as listed.

jfrank-nih (approved):

PR #4241, opened 2024-02-07
(#4240) Remove language set preference for users

### LINKED ISSUE #4240: Remove preferred language selection from  siteadmin, editor and author
*Description of what is to be accomplished.*
We are updating regression tests to move away from admin and use leas privileged users instead for workflows - we need to remove english from preferred admin language from available users so we can continue verifying semi translations of the cms fields when creating content
**Parent Issue:** #9999

### Notes

*Some complementary notes if necessary:*
* > Here goes a quote from an email
* Here goes whatever useful information can exist…
* More notes, or technical nuances/direction

navsunka:
ODE: https://ncigovcdode691.prod.acquia-sites.com/

Closes #4240

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.module
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.module
@@ -211,8 +211,13 @@ function __cgov_core_text_filter_modify($form_element) {
  * language for the UI pages to English.
  */
 function cgov_core_user_presave(EntityInterface $entity) {
-  if ($entity->isNew()) {
+  if ($entity->isNew() && empty($entity->preferred_admin_langcode->value)) {
     $entity->set('preferred_admin_langcode', 'en');
+  }
+  elseif ($entity->preferred_admin_langcode->value === 'no-preference') {
+    // support the drush command(bulk-load-users), we use no-preference because
+    // aLL new users on pre-save have a value for this field whether or not it is empty.
+    $entity->set('preferred_admin_langcode', '');
   }
 }
 
--- a/drush/Commands/cgov/CgovUserCommands.php
+++ b/drush/Commands/cgov/CgovUserCommands.php
@@ -298,6 +298,11 @@ class CgovUserCommands extends DrushCommands {
     else {
       $pwd = NULL;
     }
+    $preferred_admin_langcode = NULL;
+    if (array_key_exists('preferred_admin_langcode', $user) && !empty(trim($user['preferred_admin_langcode']))) {
+      $preferred_admin_langcode = trim($user['preferred_admin_langcode']);
+    }
+
     if (array_key_exists('saml', $user) && is_bool($user['saml'])) {
       $saml = $user['saml'];
     }
@@ -314,6 +319,9 @@ class CgovUserCommands extends DrushCommands {
       // Update the existing account.
       $this->logger->warning(dt('User !user exists already. Updating...', ['!user' => $name]));
       $account->setpassword($pwd);
+      if ($preferred_admin_langcode !== NULL) {
+        $account->set('preferred_admin_langcode', $preferred_admin_langcode);
+      }
       $account->activate();
     }
     else {
@@ -323,6 +331,9 @@ class CgovUserCommands extends DrushCommands {
         'access' => '0',
         'status' => 1,
       ];
+      if ($preferred_admin_langcode !== NULL) {
+        $new_user['preferred_admin_langcode'] = $preferred_admin_langcode;
+      }
       $account = User::create($new_user);
 
       if (!$account) {
--- a/cgov-drupal-users.yml
+++ b/cgov-drupal-users.yml
@@ -9,6 +9,7 @@ additional_users:
   - username: 'author'
     password: 'password123'
     saml: false
+    preferred_admin_langcode: 'no-preference'
     roles:
       - 'content_author'
       - 'image_manager'
@@ -16,6 +17,7 @@ additional_users:
   - username: 'editor'
     password: 'password123'
     saml: false
+    preferred_admin_langcode: 'no-preference'
     roles:
       - 'content_author'
       - 'content_editor'
@@ -33,6 +35,7 @@ additional_users:
   - username: 'siteadmin'
     password: 'password123'
     saml: false
+    preferred_admin_langcode: 'no-preference'
     roles:
       - 'content_author'
       - 'content_editor'


bryanpizzillo (changes requested):
@navsunka can you add a field for `preferred_admin_langcode` somewhere around https://github.com/NCIOCPL/cgov-digital-platform/blob/develop/drush/Commands/cgov/CgovUserCommands.php#L337.
(When the setting gets set may indicate if it must come before or after the save)

Implementation details:
1. In cgov_core.module, if a new user is being created:
    1.  If the key `preferred_admin_langcode` is set in the entity, then don't change it to `'en'` in presave.
    2. If the key is not set in the entity, then set it to `'en'` if this is a new user.
2. In CgovUserCommands: (this is for new AND existing users)
    1. If the key does not exist in the users file, then do not set the field.
        * We do not want to have to set this for every single user.  
    3. If the key exists, then set the value on the user entity for saving.
        * You are gonna need to see what the "i don't care" value is. I don't know if it is undefined or empty or something. If this is empty in the DB ping me to discuss.
 
You should also update the comment above the chain of events starting at https://github.com/NCIOCPL/cgov-digital-platform/blob/develop/drush/Commands/cgov/CgovUserCommands.php#L178.

navsunka:
In user presave, i'm getting preferred_admin_langcode field as empty even if it is not set in user object and because of that I have done some minor changes in the logic as per requirement.
Below are the test scenarios:
New user :- If user key exist(en, es, ''), set the value for that user. But if key does not exist, set 'en' which will get triggered from node presave.
Existing user:- if key exist, update user object with key value and if key doesnot exist, don't set key in user object.

@bryanpizzillo

bryanpizzillo:
> In user presave, i'm getting preferred_admin_langcode field as empty even if it is not set in user object and because of that I have done some minor changes in the logic as per requirement. Below are the test scenarios: New user :- If user key exist(en, es, ''), set the value for that user. But if key does not exist, set 'en' which will get triggered from node presave. Existing user:- if key exist, update user object with key value and if key doesnot exist, don't set key in user object.
> 
> In config YAML preferred_admin_langcode = en, this will store en language preferred_admin_langcode = es, this will store es language preferred_admin_langcode = '', this will store no preference if key doesn't exist, this will store en language
> 
> @bryanpizzillo

We discussed on Friday that you would put "no-preference" in the drupal-users yaml file. If that is not set, then you set to 'en'. So I think you need to flip the logic because we still want users created through the admin interface to get english set as their preferred language.

navsunka:
Scenarios for Creating New user through YAML file:
If preferred_admin_langcode is set to no-preference, then it will set no-preference for the user.
If preferred_admin_langcode key present but value empty, then it will set en for the user.
If preferred_admin_langcode key not present, then it will set en for the user.
If preferred_admin_langcode is set to en, then it will set en for the user.
If preferred_admin_langcode is set to es, still it will set en for the user.

Scenarios for Creating New user through UI:
Always set en for the user.

Scenarios for updating existing user through YAML file:
If preferred_admin_langcode is set to no-preference, then it will set no-preference for the user.
If preferred_admin_langcode key present but value empty, then it will set no-preference for the user.
If preferred_admin_langcode key not present, then it will not update existing language preference for the user.
If preferred_admin_langcode is set to en, then it will set en for the user.
If preferred_admin_langcode is set to es, then it will set es for the user.

drush/Commands/cgov/CgovUserCommands.php
```diff
@@ -298,6 +298,11 @@ protected function loadAdditionalUser($user) {
     else {
       $pwd = NULL;
     }
+    $preferred_admin_langcode = '';
```
bryanpizzillo, line 301: This is not needed here. You can do this down at 322. If the yaml has the key, then you need to be setting it on the user object. full stop.

drush/Commands/cgov/CgovUserCommands.php
```diff
@@ -314,6 +319,10 @@ protected function loadAdditionalUser($user) {
       // Update the existing account.
       $this->logger->warning(dt('User !user exists already. Updating...', ['!user' => $name]));
       $account->setpassword($pwd);
+      if (array_key_exists('preferred_admin_langcode', $user)) {
+        $user['preferred_admin_langcode'] = ($user['preferred_admin_langcode'] === 'no-preference') ? '' : $user['preferred_admin_langcode'];
```
bryanpizzillo, line 323: This should not be needed. 

If the key exists in the users file (`(array_key_exists('preferred_admin_langcode', $user))`), then set it on the account (`$account->set('preferred_admin_langcode', $preferred_admin_langcode);`). 

This way we ensure users in the system match the file we have set. Additionally `$preferred_admin_langcode` matches what you are doing for a new user. Which makes the code above to set `$preferred_admin_langcode` make sense all the way up there.

docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.module
```diff
@@ -211,9 +211,12 @@ function __cgov_core_text_filter_modify($form_element) {
  * language for the UI pages to English.
  */
 function cgov_core_user_presave(EntityInterface $entity) {
-  if ($entity->isNew()) {
+  if ($entity->isNew() && ($entity->preferred_admin_langcode->value !== 'no-preference')) {
     $entity->set('preferred_admin_langcode', 'en');
   }
+  if ($entity->preferred_admin_langcode->value == 'no-preference') {
```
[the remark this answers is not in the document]
bryanpizzillo: You also need to add a comment here that this is support the drush command, and that we use `no-preference` because ALL new users on pre-save have a value for this field whether or not it is empty. So our command needs to be able to differentiate when we intentionally want it empty, vs when we what to have it empty so it defaults to english. (P.S. Make sure to put the command name in here).

bryanpizzillo (changes requested):
The code below is a bit messy, and I can see where it will break when we set specific languages for a new user. So I have made some requests.

navsunka:
Scenarios for Creating New user through YAML file:
If preferred_admin_langcode is set to no-preference, then it will set no-preference for the user.
If preferred_admin_langcode key present but value empty, then it will set en for the user.
If preferred_admin_langcode key not present, then it will set en for the user.
If preferred_admin_langcode is set to en, then it will set en for the user.
If preferred_admin_langcode is set to es, still it will set es for the user.
Scenarios for Creating New user through UI:
Always set en for the user.
Scenarios for updating existing user through YAML file:
If preferred_admin_langcode is set to no-preference, then it will set no-preference for the user.
If preferred_admin_langcode key present but value empty, then it will not update existing language preference for the user.
If preferred_admin_langcode key not present, then it will not update existing language preference for the user.
If preferred_admin_langcode is set to en, then it will set en for the user.
If preferred_admin_langcode is set to es, then it will set es for the user.

drush/Commands/cgov/CgovUserCommands.php
```diff
@@ -298,6 +298,10 @@ protected function loadAdditionalUser($user) {
     else {
       $pwd = NULL;
     }
+    if (array_key_exists('preferred_admin_langcode', $user)) {
```
bryanpizzillo, line 301: You should still initialize `$preferred_admin_langcode`. I would just do right above this `$preferred_admin_langcode = NULL;` to specifically set it to a value you can check for. Then lower on, instead of !empty, just `$preferred_admin_langcode !== NULL`.  (I am surprised PHPStan did not flag this... oh wait, we probably are not linting this file because it is in drush. <insert table flip gif here>)

drush/Commands/cgov/CgovUserCommands.php
```diff
@@ -298,6 +298,10 @@ protected function loadAdditionalUser($user) {
...
+      $preferred_admin_langcode = $user['preferred_admin_langcode'];
```
bryanpizzillo, line 302: While we are initializing the var, out of an abudance of caution, can you trim `$user['preferred_admin_langcode']` to make sure we don't pass in something to Drupal that it might be offended by?

While we are at it too, let's not set $preferred_admin_langcode if the trimmed value ends up being an empty string.

The reason for that logic to go here is that everything past this point can assume if `$preferred_admin_langcode` is not NULL, then it is a properly set value. Which end up being simpler checks.

bryanpizzillo (changes requested):
A few minor things, but it is getting there.

drush/Commands/cgov/CgovUserCommands.php
navsunka: I have initialized $preferred_admin_langcode to NULL

drush/Commands/cgov/CgovUserCommands.php
navsunka: I have set this variable only if array key exist and value is present after trimming it.

drush/Commands/cgov/CgovUserCommands.php
```diff
@@ -314,6 +318,9 @@ protected function loadAdditionalUser($user) {
       // Update the existing account.
       $this->logger->warning(dt('User !user exists already. Updating...', ['!user' => $name]));
       $account->setpassword($pwd);
+      if (array_key_exists('preferred_admin_langcode', $user) && !empty($preferred_admin_langcode)) {
```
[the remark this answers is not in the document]
navsunka: I'm setting this variable in array only if  $preferred_admin_langcode is !NULL.

bryanpizzillo (approved):
@navsunka this looks great. Go ahead and rebase it.

santwuma:
@navsunka to update YAML content on all environments before merge

navsunka:
I have set preferred_admin_langcode to no-preference in Drupal user file for all the environments.
@bryanpizzillo

PR #4256, opened 2024-02-13
(#3695, #4255) Replace Behat with Cypress

### LINKED ISSUE #3695: Enabler: Replace Behat with Cypress

### LINKED ISSUE #4255: Enabler: Update GitHub Actions to use Node 20.
# Update GitHub Actions to a version using Node 20.

We are using several actions which run on node 16; that version is deprecated.

https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

Update:
  - actions/cache
  - actions/checkout
  - actions/setup-node

blairlearn:
- Run tests in  cypress tasks via 'blt tests:cypress`.
  - Tests run in parallel via cypress-parallel.
- Tests run in GH workflow.
- Preserve parallel-weights file for tuning adjustments.
- Bump actions to versions using Node 20
  - actions/cache
  - actions/checkout
  - actions/setup-node

Closes #3695 
Closes #4255


blairlearn:
[ODE 703](https://ncigovcdode703.prod.acquia-sites.com/)

.github/workflows/ci.yml
```diff
@@ -125,10 +125,45 @@ jobs:
       - name: Run Front-end Tests
         run: |
           blt tests:frontend:run --define drush.alias='${drush.aliases.ci}' --define tests.run-server=false --environment=ci --ansi --verbose --no-interaction
+      - name: Install cypress
+        run: |
+              blt tests:cypress:npm-install --define drush.alias='${drush.aliases.ci}' --environment=ci --ansi --verbose --no-interaction
```
bryanpizzillo, line 130: What happens when you run `blt tests:all`? Do these get run everytime? Just wondering if npm-install should be outside of the tests namespace.

.github/workflows/ci.yml
```diff
@@ -125,10 +125,45 @@ jobs:
...
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: cypress-pretest-screenshots
+          path: ./tests/cypress/cypress/screenshots
       - name: Behavioral Tests
         timeout-minutes: 120
         run: |
-          echo "TODO: Add Cypress tests here"
+              blt tests:cypress:runimpl --define drush.alias='${drush.aliases.ci}' --define tests.run-server=false --environment=ci --ansi --verbose --no-interaction
```
bryanpizzillo, line 149: This should be run to follow the other test naming conventions.

blt/blt.yml
```diff
@@ -231,6 +231,11 @@ validate:
       - files.php.tests
 
 tests:
+  cypress:
+    basePath: ${repo.root}/tests/cypress
```
bryanpizzillo, line 235: This should include a set of the tests we want to run as was done in https://github.com/NCIOCPL/cgov-digital-platform/blob/d7671e9d0ff82fe16aebe6b0c6eea2c3b6109904/blt/blt.yml#L237. 

What is being done in that branch is that it is listing a number of folders that contain features. Each of the folders run sequentially, but the features within that folder run in parallel. 

"Pre-test" scripts would come first.

bryanpizzillo, line 235: This should also control the number of threads to use. If you look at https://github.com/NCIOCPL/cgov-digital-platform/blob/d7671e9d0ff82fe16aebe6b0c6eea2c3b6109904/blt/blt.yml#L241, you will see that ci.blt.yml sets the threads to 3. This is because we found that locally 2 was the best, but Github actions could handle 3.

blt/ci.blt.yml
```diff
@@ -3,6 +3,7 @@
 #   sync - Uses `blt drupal:sync` to pull a remote db from drush.aliases.remote.
 #   import - Imports a .sql file from setup.dump-file.
 # setup.strategy: import
+setup.strategy: install
```
bryanpizzillo, line 6: Why is this needing to be set?

blt/src/Blt/Plugin/Commands/Tests/CypressTestCommand.php
```diff
@@ -0,0 +1,178 @@
...
+   * @command tests:cypress:pretest
+   *
+   * @interactGenerateSettingsFiles
+   * @interactInstallDrupal
+   * @validateDrupalIsInstalled
+   * @launchWebServer
+   *
+   * @return void
+   */
+  public function cypressSetuptests() {
```
bryanpizzillo, line 99: So pre-tests should be run as part of tests. See my comments on the blt.yml

blt/src/Blt/Plugin/Commands/Tests/CypressTestCommand.php
```diff
@@ -0,0 +1,178 @@
...
+
+    if($exit_code) {
+      throw new BltException('Cypress pre-test failed.');
+    }
+  }
+
+  /**
+   * Run cypress test directly, bypassing the setup steps.
+   *
+   * @command tests:cypress:runimpl
```
bryanpizzillo, line 119: as noted in the workflow this should be `tests:cypress:run` to match the naming the testing naming convention.

blt/src/Blt/Plugin/Commands/Tests/CypressTestCommand.php
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+  }
+
+  /**
+   * Helper method to handle the details of launching Xvfb.
+   */
+  private function startXvfb() {
+
+    $port = CypressTestCommand::XVFB_PORT;
```
bryanpizzillo, line 159: You know, you could just make these attributes like @launchWebServer that can wrap the running of the command.

blt/src/CgovNpmTrait.php
```diff
@@ -0,0 +1,55 @@
...
+   * Creates a task to run an npm script.
+   *
+   * @param null|string $pathToNpm
+   *   Path to a specific NPM installation.
+   *
+   * @return \Robo\Collection\CollectionBuilder
+   *   CollectionBuilder for executing the npm run-script task.
+   */
+  protected function taskNpmRunScript($pathToNpm = NULL) {
+    return $this->task(RunScript::class, $pathToNpm);
```
bryanpizzillo, line 52: Since I am noting that you should be taking in the tests to run, you should probably allow this to take parameters. You can do `npm run foo -- --bar --baz`. The "empty" `--` option just passes along the parameters to the script it is going to run. See https://docs.npmjs.com/cli/v9/commands/npm-run-script.

This would allso you to vary the threads, etc.

blt/src/Task/Npm/RunScript.php
```diff
@@ -0,0 +1,48 @@
...
+  /**
+   * Set the name of the script to be executed.
+   *
+   * @param mixed $scriptName
+   *   Name of the script to execute.
+   *
+   * @return $this
+   *   Instance of the task
+   */
+  public function scriptName($scriptName) {
```
bryanpizzillo, line 30: See my note above about how npm run can take additional parameters with the `--` option

bryanpizzillo (changes requested):
I have some items. Somewhere along the way the configuration work done in https://github.com/NCIOCPL/cgov-digital-platform/pull/4200 was lost. But we want to set what feature folders should run in the blt.yml, as well as be able to control threads from there. Github last we tested, could handle 3 threads instead of just 2. But local machines would struggle with 3.

.github/workflows/ci.yml
blairlearn: Done.

blt/blt.yml
blairlearn: Done

.github/workflows/ci.yml
blairlearn: The steps have been consolidated to expect "pre-tests" to be the first in the list.

Additionally, the logic has been updated so `npm ci` is only run if the node modules have not been previously installed.

blt/blt.yml
blairlearn: Done

blt/src/Blt/Plugin/Commands/Tests/CypressTestCommand.php
blairlearn: Done

blt/src/Blt/Plugin/Commands/Tests/CypressTestCommand.php
blairlearn: Fixed

blt/src/Blt/Plugin/Commands/Tests/CypressTestCommand.php
blairlearn: That does simplify the logic a bit.  And gave me an idea for something similar with running `npm ci`.

blt/src/CgovNpmTrait.php
blairlearn: Yep, that's what I ended up with.

blt/src/Task/Npm/RunScript.php
blairlearn: 👍

blt/blt.yml
```diff
@@ -231,6 +231,16 @@ validate:
...
+    paths:
+      - specDir: cypress/e2e/pretest/**/*.feature
```
bryanpizzillo, line 237: This gives me slight bad vibes. However, as long as there is just 1 feature this is ok. So maybe make a note here to be REAL careful.

bryanpizzillo (approved):
I have one small thing, just a comment.

blt/blt.yml
blairlearn: Yeah, it gave me bad vibes too. That's why I had originally wanted to do that as a discrete step.  We may have to revisit it one day, for now, I'll add a comment as we discussed.

KateMashkinaNIH (approved):

PR #4278, opened 2024-03-07
(#4233) Allow site name in meta title

### LINKED ISSUE #4233: Allow site name in meta title (`<title>`) to be optional/configurable
### Request Description

There is a need for some of our platform sites to not have explicit NCI branding. Currently the browser title is configured in meta tags as `'[cgov_tokens:cgov-title] - [cgov_tokens:cgov-trans-org-name]'`. The `[cgov_tokens:cgov-trans-org-name]` just returns `t('NCI')` as all pages need to have a translated version of NCI.

What we propose is to add two new basic site settings (a [Drupal post may work for you](https://www.drupal.org/forum/support/post-installation/2019-02-06/add-a-new-custom-field-to-site-infomation-form-in-drupal8)):
1. A multi-select with a label of Browser Title Site Name Display Option[1] for how the site name should display the org name. The options should be:
    * NCI -- This is the default and should do what is done today. Make this dropdown value be `default`
    * No Title -- this should just use `'[cgov_tokens:cgov-title]'` as the title tag (i.e. no ` - [cgov_tokens:cgov-trans-org-name]`)
    * Custom Title - This will use an untranslated (no `t()`) custom title for the `- [cgov_tokens:cgov-trans-org-name]` part, although do not modify the `cgov_tokens:cgov-trans-org-name` to output this custom title -- you will need something else to generate the correct browser title.
2. A required text input for storing the Custom Site Title, that is disabled until the "Custom Title" option is selected.

> **NOTE:**
> This approach does not support having multiple Custom Site Titles for multiple languages. The LOE to do so would be fairly high due to:
> * The config_translation module does not work right for our configuration. Actually, we found it would not translate these types of things anyway
> * The `t()` function is only good if the codebase's translations.po file has the translated strings. Since this is entered as content we would not know all the strings possible to put in the file...
> * One should never hardcode fields for languages. You should always support the enabled languages, while currently this is English and Spanish, it comes up from time to time.
> * So this basically leaves us with some complex map widget of enabled languages to custom strings. (Something like a list with a "add translation" button and you select the language, and the button becomes disabled when there are no more languages -- so basically implementing field translations in configs.

In order to ensure backwards compatibility, if the Browser Title Site Name Display Option has a null value, you should default to acting as if NCI were chosen.

The cgov_site profile's install (or postinstall) hook should set this value. You cannot add the system.site config to the config/install because reasons. (Well reasons being that the name is set by blt/drush install and the front URL is setup when a root site section is created with a landing page. So it is volatile )

You should also add an update hook to set the default to be `default` for any current sites.

[1] Very much open to a shorter name that means something to people.

### Acceptance Criteria
```gherkin
Scenario: A site admin is editing the Main page content on the Site Branding settings
  Given the site admin is logged into the CMS
    And they hover over the Configuration section in the mega menu
    And they hover over the System section
    And they click the Site Branding option
    And the site name in the meta title has been updated to be optional/configurable
  When a user is editing the site branding settings
  Then they will see a collapsable section titled "Site Branding" 
  When the "Site Branding" section is expanded
  Then the first field displayed will be the "Browser Title Site Name Display Option"
    And this field will be required
    And this field will be a dropdown
    And the options will be "NCI", "No Title", and "Custom Title"
    And the "NCI" option will be the default
    And the tooltip text for this field will be "Branding title appended to the end of the browser title for all pages on the site. Default is "- NCI". Select "No Title" for sites that do not want to have NCI appended to their browser titles."
    And the second field displayed will be the "Custom Site Title"
    And this field will be required
    And the field will be a text input
    And the maximum number of characters for the field will be 50
    And the tooltip text for the field will be "Custom title text to be appended to the browser title. Do not include dash (-)."
    And the Custom Site Title field will be disabled
  When the site admin selects the "Custom Title" option from the Browser Title Site Name Display Option
  Then the Custom Site Title field will become enabled
  When the user selects the "Custom Title" option from the Browser Title Site Name Display Option
    And the user enters text into the Custom Site Title field
  Then the user returns to the Browser Title Site Name Display Option
    And the user selects a different options from this field
  Then the Custom Site Title field will return to the disabled state
    And the text entered into the Custom Site Title field will keep the text previously entered
```

```gherkin
Scenario: A site admin chooses "NCI" as the option from the "Browser Title Site Name Display Option" field in the site branding settings on main page content
  Given the site admin is on the site branding settings for a main page content
    And they selected "NCI" as the option for the "Browser Title Site Name Display Option" field
  When the site admin chooses the "NCI" option from the "Browser Title Site Name Display Option" field
    And they view a page with the Browser Title "My Page"
  Then the <title> tag will be "My Page - NCI"
```

```gherkin
Scenario: A site admin chooses "No Title" as the option from the "Browser Title Site Name Display Option" field in the site branding settings on main page content
  Given the site admin is on the site branding settings for a main page content
    And they selected "No Title" as the option for the "Browser Title Site Name Display Option" field
  When the site admin chooses the "No Title" option from the "Browser Title Site Name Display Option" field
    And they view a page with the Browser Title "My Page"
  Then the <title> tag will be "My Page"
```

```gherkin
Scenario: A site admin chooses "Custom Title" as the option from the "Browser Title Site Name Display Option" field in the site branding settings on main page content
  Given the site admin is on the site branding settings for a main page content
    And they selected "Custom Title" as the option for the "Browser Title Site Name Display Option" field
  When the site admin chooses the "Custom Title" option from the "Browser Title Site Name Display Option" field
    And the user views a page with the Browser Title "My Page"
    And they have not entered any text into the "Custom Site Title" field
    And the user tries to save the edits to the page
  Then an error will appear on the "Custom Site Title" field which will state, "Please enter the Custom Site Title text to save this page."
  When the user views a page with the Browser Title "My Page"
    And they enter text, "Custom1" into the "Custom Site Title" field
  Then the <title> tag will be "My Page - Custom1"
```

navsunka:
ODE: https://ncigovcdode712.prod.acquia-sites.com/

Closes #4233

### DIFF
--- a/docroot/profiles/custom/cgov_site/cgov_site.install
+++ b/docroot/profiles/custom/cgov_site/cgov_site.install
@@ -18,5 +18,4 @@ function cgov_site_install() {
 
   // Install proper language negotiation even if English only.
   $siteHelper->installLanguageNegotiation();
-
 }
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_application_page/cgov_application_page.module
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_application_page/cgov_application_page.module
@@ -151,7 +151,7 @@ function cgov_application_page_preprocess_html(&$variables) {
         $token_service = \Drupal::token();
         $title = Html::escape(
           $token_service->replace(
-            '[cgov_tokens:cgov-title] - [cgov_tokens:cgov-trans-org-name]',
+            '[cgov_tokens:cgov-title][cgov_tokens:browser-title-site-name-meta]',
             [
               'node' => $node,
             ]
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_application_page/config/install/metatag.metatag_defaults.node__cgov_application_page.yml
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_application_page/config/install/metatag.metatag_defaults.node__cgov_application_page.yml
@@ -20,5 +20,5 @@ tags:
   og_title: '[node:field_browser_title:value]'
   og_type: Website
   og_url: '[current-page:url]'
-  title: '[cgov_tokens:cgov-title] - [cgov_tokens:cgov-trans-org-name]'
+  title: '[cgov_tokens:cgov-title][cgov_tokens:browser-title-site-name-meta]'
   twitter_cards_type: summary
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_article/config/install/metatag.metatag_defaults.node__cgov_article.yml
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_article/config/install/metatag.metatag_defaults.node__cgov_article.yml
@@ -7,7 +7,7 @@ tags:
   canonical_url: '[node:url]'
   content_language: '[node:langcode]'
   description: '[node:field_page_description:value]'
-  title: '[cgov_tokens:cgov-title] - [cgov_tokens:cgov-trans-org-name]'
+  title: '[cgov_tokens:cgov-title][cgov_tokens:browser-title-site-name-meta]'
   dcterms_coverage: 'nciglobal,ncienterprise'
   dcterms_issued: '[node:field_date_posted:date:short]'
   dcterms_is_part_of: '[cgov_tokens:cgov-analytics-group]'
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_biography/config/install/metatag.metatag_defaults.node__cgov_biography.yml
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_biography/config/install/metatag.metatag_defaults.node__cgov_biography.yml
@@ -7,7 +7,7 @@ tags:
   canonical_url: '[node:url]'
   content_language: '[node:langcode]'
   description: '[node:field_page_description:value]'
-  title: '[cgov_tokens:cgov-title] - [cgov_tokens:cgov-trans-org-name]'
+  title: '[cgov_tokens:cgov-title][cgov_tokens:browser-title-site-name-meta]'
   dcterms_coverage: 'nciglobal,ncienterprise'
   dcterms_issued: '[node:field_date_posted:date:short]'
   dcterms_is_part_of: '[cgov_tokens:cgov-analytics-group]'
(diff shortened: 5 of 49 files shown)


docroot/lib/CgovPlatform/Tests/CgovSchemaExclusions.php
```diff
@@ -39,6 +39,7 @@ class CgovSchemaExclusions {
     // for the tracked issue.
     'adobe_dtm.settings',
     'views.view.samlauth_map',
+    'system.site',
```
bryanpizzillo, line 42: This should not be here. If you tests are complaining about things missing from the schema, you need to fix the schema.

docroot/profiles/custom/cgov_site/modules/custom/cgov_metatag/cgov_metatag.install
```diff
@@ -0,0 +1,15 @@
...
+ * @file
+ * Install, update and uninstall functions for the cgov_metatag.
+ */
+
+/**
+ * Set NCI as default site Branding.
+ */
+function cgov_metatag_update_10001() {
+  $siteHelper = \Drupal::service('cgov_core.tools');
+    // Set Default Site Branding.
```
bryanpizzillo, line 13: Please outdent this so it aligns

docroot/profiles/custom/cgov_site/modules/custom/cgov_metatag/cgov_metatag.module
```diff
@@ -31,3 +31,86 @@ function cgov_metatag_entity_presave($entity) {
...
+    ->getEditable('system.site')
+    ->set('sitebranding', $value)
+    ->save();
+}
+
+/**
+ * Custom validate callback.
+ */
+function cgov_metatag_handler_validate(&$form, $form_state) {
+  // Validate 404 error path.
```
bryanpizzillo, line 100: This comment looks like a copy/paste error. Please remove or update.

bryanpizzillo (changes requested):
One big thing, a couple of small things.

IF you can't extend the schema for the `system.site` config, then we should just make a new config and route for the editing form and make it something separate from the Basic Site Settings.

santwuma:
Needs final code review from @bryanpizzillo

bryanpizzillo:
@andyvanavery31, @sarinapadilla, @lburack does this pass from a product standpoint?

andyvanavery31:
@bryanpizzillo I reviewed it, and as long as you are good with it, and haven't requested any changes that I haven't seen yet, then I'm good with it

bryanpizzillo (approved):
Looks good!

PR #4282, opened 2024-03-11
(#4267) Pin Highcharts to v11.4.0

### LINKED ISSUE #4267: Pin Highcharts to a specific version
### Issue description
*Current implementation of Highcharts uses the most recent stable version. It is not pinned to a specific version. This could cause instability and compatibility issues between our current use/configurations against newer version if there happens to be a breaking change.

> **ESTIMATE** 2

- Add version number to loading of Highcharts library (use most recent version since this is what would currently be loaded)

seyilonge-nci:
* Updated Highcharts and plugins to v11.4.0

Closes #4267


alinai:
Looks good on ODE711. Thanks @seyilonge-nci.
cc: @KateMashkinaNIH , @santwuma

blairlearn (approved):

PR #4290, opened 2024-03-25
(#2260) Add Override Thumbnail Image to Videos

### LINKED ISSUE #2260: Add Promotional Image to Videos
Videos can be added to a list.
Lists can display images.
So videos should have a promotional image (optional; not a required field). 

**This ticket is to add the promotional image field to videos.**

The potential use-case for this is on the [DCEG Videos page](https://dceg-prod-acsf.cancer.gov/news-events/videos). Right now, this is a mini landing page list, without images. But we may want to display images in the future.

navsunka:
ODE: http://ncigovcdode730.prod.acquia-sites.com/
Closes #2260

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_video/cgov_video.module
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_video/cgov_video.module
@@ -6,6 +6,8 @@
  */
 
 use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Entity\Display\EntityViewDisplayInterface;
 
 /**
  * Implements hook_form_alter().
@@ -19,5 +21,17 @@ function cgov_video_form_alter(&$form, FormStateInterface $form_state, $form_id)
     $userCreatedDescription = $form['field_media_oembed_video']['widget']['#description'];
     $form['field_media_oembed_video']['widget'][0]['value']['#description'] = $userCreatedDescription;
   }
+}
 
+/**
+ * Implements hook_ENTITY_TYPE_view().
+ */
+function cgov_video_media_view(array &$build, EntityInterface $entity, EntityViewDisplayInterface $display, $view_mode) {
+  if ($entity->bundle() != 'cgov_video') {
+    return;
+  }
+  if(!empty($entity->field_override_thumbnail_video->target_id) && isset($build['field_override_thumbnail_video'])) {
+    $build['thumbnail'] = $build['field_override_thumbnail_video'];
+    unset($build['field_override_thumbnail_video']);
+  }
 }
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_video/config/install/core.entity_view_display.media.cgov_video.default.yml
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_video/config/install/core.entity_view_display.media.cgov_video.default.yml
@@ -14,6 +14,7 @@ dependencies:
     - field.field.media.cgov_video.field_list_description
     - field.field.media.cgov_video.field_media_oembed_video
     - field.field.media.cgov_video.field_meta_tags
+    - field.field.media.cgov_video.field_override_thumbnail_video
     - field.field.media.cgov_video.field_page_description
     - field.field.media.cgov_video.field_pretty_url
     - field.field.media.cgov_video.field_related_resources
@@ -38,12 +39,12 @@ content:
     label: hidden
     settings: {  }
     third_party_settings: {  }
-    weight: 16
+    weight: 11
     region: content
   content_moderation_control:
     settings: {  }
     third_party_settings: {  }
-    weight: -20
+    weight: 0
     region: content
   created:
     type: timestamp
@@ -62,7 +63,7 @@ content:
         granularity: 2
         refresh: 60
     third_party_settings: {  }
-    weight: 0
+    weight: 2
     region: content
   field_browser_title:
     type: string
@@ -70,14 +71,14 @@ content:
     settings:
       link_to_entity: false
     third_party_settings: {  }
-    weight: 8
+    weight: 5
     region: content
   field_caption:
     type: text_default
     label: hidden
     settings: {  }
     third_party_settings: {  }
-    weight: 13
+    weight: 9
     region: content
   field_card_title:
     type: string
@@ -85,14 +86,14 @@ content:
     settings:
       link_to_entity: false
     third_party_settings: {  }
-    weight: 10
+    weight: 6
     region: content
   field_date_display_mode:
     type: list_default
     label: hidden
     settings: {  }
     third_party_settings: {  }
-    weight: 21
+    weight: 15
     region: content
   field_date_posted:
     type: cgov_date_formatter
@@ -102,7 +103,7 @@ content:
       format_type: cgov_display_date
       display_strategy: latest
     third_party_settings: {  }
-    weight: 17
+    weight: 12
     region: content
   field_date_reviewed:
     type: cgov_date_formatter
@@ -112,7 +113,7 @@ content:
       format_type: cgov_display_date
       display_strategy: latest
     third_party_settings: {  }
-    weight: 19
+    weight: 13
     region: content
   field_date_updated:
     type: cgov_date_formatter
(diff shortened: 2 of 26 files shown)


docroot/profiles/custom/cgov_site/modules/custom/cgov_image/config/install/image.style.video_display_large_medium_small_full_content.yml
```diff
@@ -0,0 +1,25 @@
+langcode: en
+status: true
+dependencies:
+  config:
+    - crop.type.4x3
+  module:
+    - crop
+name: video_display_large_medium_small_full_content
+label: 'Video Display: Large/Medium/Small & Full Content'
```
bryanpizzillo, line 9: What is this style? There should be a style for each:
* Full
* Large
* Medium
* Small

docroot/profiles/custom/cgov_site/modules/custom/cgov_video/config/install/core.entity_view_display.media.cgov_video.default.yml
```diff
@@ -160,6 +161,17 @@ content:
     third_party_settings: {  }
     weight: 25
     region: content
+  field_override_thumbnail_video:
```
bryanpizzillo, line 164: Is default being used somewhere specific? If not, then you should just hide this field.

docroot/profiles/custom/cgov_site/modules/custom/cgov_image/config/install/image.style.video_display_large_medium_small_full_content.yml
bryanpizzillo: Firstly, Full and Large end up being the same size on screen - 994 px. So the "Full" display mode for a video should use the "large" image mode. (Full will go away at some point anyway)

As far as Large, Medium and Small:

At 0-1023 pixels wide we are full-width, which ends up being 991px in non-retina. If we go by 2x then we get 1982, which is not 1988. 
On desktop and up, the image is 994, which gives us 1988, which is what is in requirements.

At the very least there should be a different Large vs Medium/Small where Medium/Small is resized to 1982. 

**HOWEVER**, part of the use of image styles is future responsive images. This is where we render an image with multiple sizes, which multiple max sizes. In this case, is important to separate all 3 image styles. 

So while they have the same max-width, they are 3 different display cases, and if we turned on responsive images, which will happen some day, we would end up having to split out all the sizes.

So for the implementation, let's create 3 separate "Video Display: XXX" image styles for Large, Medium and Small.

bryanpizzillo (changes requested):
Some items to review.

docroot/profiles/custom/cgov_site/modules/custom/cgov_video/config/install/core.entity_view_display.media.cgov_video.default.yml
andyvanavery31: @navsunka I updated the requirements on the ticket so it broke out "Video Display: Full Content", "Video Display: Large", "Video Display: Medium", "Video Display: Small" to account for that part of Bryan's comment

docroot/profiles/custom/cgov_site/modules/custom/cgov_image/config/install/image.style.video_display_large_medium_small_full_content.yml
andyvanavery31: @navsunka I updated the requirements on the ticket so it broke out "Video Display: Full Content", "Video Display: Large", "Video Display: Medium", "Video Display: Small" to account for that part of Bryan's comment

bryanpizzillo: However, as I noted, there does not need to be "Video Display: Full Content"
> Firstly, Full and Large end up being the same size on screen - 994 px. So the "Full" display mode for a video should use the "large" image mode. (Full will go away at some point anyway)

@navsunka  can you please remove the Full image style and just point Full at the large image style.

docroot/profiles/custom/cgov_site/modules/custom/cgov_video/cgov_video.module
```diff
@@ -19,5 +21,43 @@ function cgov_video_form_alter(&$form, FormStateInterface $form_state, $form_id)
...
+ * @param Drupal\Core\Entity\EntityInterface $entity
+ *   The entity object.
+ * @param Drupal\Core\Entity\Display\EntityViewDisplayInterface $display
+ *   The entity view display holding the display options configured for
+ *   the entity components.
+ * @param string $view_mode
+ *   The view mode the entity is rendered in.
+ */
+function _cgov_video_ovverride_thumbnail(array &$build, EntityInterface $entity, EntityViewDisplayInterface $display, $view_mode) {
+  $build['thumbnail'] = $build['field_override_thumbnail_video']['#items']->view([
```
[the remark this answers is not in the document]
bryanpizzillo: @navsunka Why is this resolved when it does not match the code? (or at least adding a comment to this thread as to why it could not be resolved?)

bryanpizzillo (changes requested):
@navsunka and @andyvanavery31 - I noted that the Full display mode should just use the Large image style in my comment. However, I see that was missed and we have a full image style. **We should not have a full image style**.  The Full Display Mode should use  the Large image style.

Also @navsunka sorry I missed this, but can you put these display mode configs in the cgov_video module. They have nothing to do with the cgov_image or cgov_contextual_image content types (which are in cgov_image, and which is why their image styles live there.) The logic is that if we don't want cgov_video on a site, we don't want the thumbnail's display modes either. Each module should hold the config that is related to *that* module.

docroot/profiles/custom/cgov_site/modules/custom/cgov_video/config/install/core.entity_view_display.media.cgov_video.default.yml
bryanpizzillo: @navsunka this comment was resolved, but there was no comment or code change. Did you mark it resolved by accident?

bryanpizzillo (changes requested):
I had longer comments, but github removed them because of all the "resolved" comments that were unresolved. So this is going to be pithier than before.
1. Remove the Full Image style and change the Full display mode to use the Large image style as I noted in my original comment.
2. Move the video image styles to the cgov_video module as the video image styles are only for the cgov_video content type.

docroot/profiles/custom/cgov_site/modules/custom/cgov_image/config/install/image.style.video_display_large_medium_small_full_content.yml
navsunka: Removed the Full image style.

docroot/profiles/custom/cgov_site/modules/custom/cgov_video/config/install/core.entity_view_display.media.cgov_video.default.yml
navsunka: Hidden the override_thumbnail_video field for default.

alinai:
Looks good on ODE 730. @navsunka , confirmed that we are making it 4:3 once uploaded. 
cc: @KateMashkinaNIH

docroot/profiles/custom/cgov_site/modules/custom/cgov_video/config/install/core.entity_view_display.media.cgov_video.video_display_cthp_multimedia.yml
```diff
@@ -51,7 +52,18 @@ content:
...
     third_party_settings: {  }
-    weight: 1
+    weight: 2
+    region: content
+  field_override_thumbnail_video:
+    type: image
+    label: above
+    settings:
+      image_link: ''
+      image_style: cthp_card
```
bryanpizzillo: this should change to `video_display_cthp_card`

bryanpizzillo (changes requested):
Some quick things

alinai:
@navsunka , looks good! cc: @KateMashkinaNIH

bryanpizzillo (approved):

PR #4296, opened 2024-04-01
(#4212) Bump Typescript to 5.2.2

### LINKED ISSUE #4212: Enabler: Bump TypeScript to 5.x in ncids_trans
# Bump TypeScript to 5.x
> **ESTIMATE** TBD

We are using a very? old version of TypeScript. Which is find and all, but this version does not support using package exports, which are in the process of moving to in NCIDS. This will be especially important as we start doing more custom elements for certain template types in order to keep build sizes small.

This ticket will also require jest to be updated as ts-jest is what jest uses to transpile the typescript, and the version we have does not support 5. (Saying jest needs to be updated means all the things Jest needs -- so if you get assigned this ticket, and you do not know what this means, don't try it.)

Hopefully this should not need Node updated to 18.

bryanpizzillo:
This also bumped Jest to 29 and the TS moduleResolution was changes to honor package.json exports.

Closes #4212


arcepaul (approved):

olitharp-nci (approved):

bryanpizzillo:
This passed visual regression tests.

PR #4297, opened 2024-04-02
(#4262) Enabling Admin user if its blocked.

### LINKED ISSUE #4262: In drush resetAdmin unblock the user
In the drush `resetAdmin` command we are setting the admin username and password. But we're not _enabling_ the user if it's blocked. This means that a reset admin that was blocked on a higher tier is still blocked.

https://github.com/NCIOCPL/cgov-digital-platform/blob/2921bb868eaca051fbc6b04a01a9d2b0fb1cca5b/drush/Commands/cgov/CgovUserCommands.php#L132-L147

To fix this we should update this code in changeAdminUserInfo:

https://github.com/NCIOCPL/cgov-digital-platform/blob/2921bb868eaca051fbc6b04a01a9d2b0fb1cca5b/drush/Commands/cgov/CgovUserCommands.php#L57-L65

It should now read:

````php
   $user = $userConfig['admin']['username'];
   $pass = $userConfig['admin']['password'];
   $this->resetAdmin($user, $pass);
    if ($userConfig['admin']['enabled'] === TRUE) {
      $this->unBlockAdminUser();
    }
    else {
      $this->blockAdminUser();
      $this->logger->debug('Admin has been blocked.');
    }
````

Then we should add a new `unBlockAdminUser` that does the opposite of `blockAdminUser`

### PR DESCRIPTION
**Pull request by navsunka**
Closes #4262

### DIFF
--- a/drush/Commands/cgov/CgovUserCommands.php
+++ b/drush/Commands/cgov/CgovUserCommands.php
@@ -56,8 +56,9 @@ class CgovUserCommands extends DrushCommands {
 
     $user = $userConfig['admin']['username'];
     $pass = $userConfig['admin']['password'];
+    $this->resetAdmin($user, $pass);
     if ($userConfig['admin']['enabled'] === TRUE) {
-      $this->resetAdmin($user, $pass);
+      $this->unblockAdminUser();
     }
     else {
       $this->blockAdminUser();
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: drush/Commands/cgov/CgovUserCommands.php [hunk 1]
```diff
@@ -90,6 +91,42 @@ public function blockAdminUser() {
...
+   * @usage drush cgov:unblock-admin-user
+   *   UnBlock the admin user (user 1).
+   * @bootstrap full
+   */
+  public function unblockAdminUser() {
+    /** @var \Drupal\user\Entity\User $account */
+    if ($account = User::load(1)) {
+      $account->activate();
+      $account->save();
+      if (\Drupal::moduleHandler()->moduleExists('samlauth')) {
```
jfrank-nih, line 108: You'll actually want to check to see if `externalauth` exists, not `samlauth`. External Authentication is the module that controls that table.

drush/Commands/cgov/CgovUserCommands.php
```diff
@@ -90,6 +91,42 @@ public function blockAdminUser() {
...
+        // Unlink saml for admin user.
+        $authname = \Drupal::database()->select('authmap', 'm')
```
jfrank-nih, line 110: It doesn't hurt to check first, but there's also no particular need to, is there? We could just issue the delete command and it will delete if anything exists.

jfrank-nih (changes requested):

jfrank-nih:
Just the standard regressions, please.

jfrank-nih (approved):

PR #4304, opened 2024-04-10
(#3062) same site section at root should not get created

### LINKED ISSUE #3062: A new site section should not be able to be created at the root

navsunka:
ODE: https://ncigovcdode744.prod.acquia-sites.com/

Closes #3062

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/cgov_site_section.module
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/cgov_site_section.module
@@ -774,6 +774,12 @@ function cgov_site_section_entity_bundle_field_info_alter(&$fields, EntityTypeIn
     ]
     );
   }
+  if ($entity_type->id() == 'taxonomy_term'  && $bundle === 'cgov_site_sections') {
+    if (isset($fields['parent'])) {
+      // Add custom constraint to this field
+      $fields['parent']->addConstraint('CgovValidateParentConstraint', []);
+    }
+  }
 }
 
 /**
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/src/Plugin/Validation/Constraint/ValidateParentConstraint.php
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/src/Plugin/Validation/Constraint/ValidateParentConstraint.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace Drupal\cgov_site_section\Plugin\Validation\Constraint;
+
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * A composite constraint validating Site Section parent.
+ *
+ * @Constraint(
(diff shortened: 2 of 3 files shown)


bryanpizzillo (commented):
IDK how this was actually tested, but I enabled Greek on the ODE and *was* able to create a new root term. (Which is correct) 

This use case simulates when we would create a new Spanish root, which IIRC, is part of the process of making a site support Spanish.

IDK if @lburack looked at this, but the error message did not make much sense. If I try and create a duplicate, then "The site section cannot be saved. Please set the parent site section under the Relations tab." shows. But I do have a parent set, it just happens to be `<root>`.

We should be noting that you cannot have more than 1 site section in the same language at the root.

lburack:
I did not review the error message. I suggest the following instead: 

> The site section cannot be saved. There is already a site section at the root (the homepage). Please update the parent site section under the Relations tab.

@andyvanavery31 @bryanpizzillo @navsunka

KateMashkinaNIH:
> IDK how this was actually tested, but I enabled Greek on the ODE and was able to create a new root term. (Which is correct)
@bryanpizzillo since we are not supposed to use admin user for testing purposes, I did a workaround of editing Spanish root and moving it one level down in the relations tree, effectively getting rid of existing spanish root and then proceeding with new root creation, same thing with english ( and then simply moving them back to root to restore original site sections relations after deleting test roots) 

ODE functions according to acceptance criteria ( plus I verified that user IS able to select all 3 Navigation Display Options) 
Also Regression tests pass

bryanpizzillo:
> > IDK how this was actually tested, but I enabled Greek on the ODE and was able to create a new root term. (Which is correct)
> > @bryanpizzillo since we are not supposed to use admin user for testing purposes, I did a workaround of editing Spanish root and moving it one level down in the relations tree, effectively getting rid of existing spanish root and then proceeding with new root creation, same thing with english ( and then simply moving them back to root to restore original site sections relations after deleting test roots)
> 
> ODE functions according to acceptance criteria ( plus I verified that user IS able to select all 3 Navigation Display Options) Also Regression tests pass

This was more about how to test that you *could* create a new top level node in another language. But I guess the method you could use worked. I assume you mean by "moving it down" was to move Spanish to under English.

bryanpizzillo (approved):

PR #4306, opened 2024-04-10
(#4305) Unblock css and js assets from login protection

### LINKED ISSUE #4305: Under certain circumstances aggregated CSS and JS files can be redirected to /saml/login on a -cms site.
We don't exactly know how we have not encountered this issue until now, but when going to `https://www-cms-test.cancer.gov/user/login?show_login_fields=true` CSS & JS asset linked to from the HTML may be 302 redirects to /saml/login. This is because of our CMSProtectionSubscriber. It allows access to /user/login, but that is it. Any additional urls, such as aggregate CSS and JS files would be redirected. However, the CMSProtectionSubscriber only fires if the file does not exist. If the file exists, then it can be downloaded.

During v2.4.0 automation testing, a console error saying that jQuery could not be found on the window was causing the automation tests to end abruptly. This was because our jQuery file on the front-end is allowed to be aggregated, and that aggregated route was being blocked. This did not manifest itself on the front-end, just a console error. 

An additional issue in the mix is that ACSF dev has aggregation turned off. So only production and test are the only two environments where this could have occurred, and prod does not have automation tools or people logging in via the form login. This ticket should also make ACSF dev follow the same aggregator settings as ACSF test.

bryanpizzillo:
I enabled CSS aggregation on all environments. The local.settings.php that ships with Drupal disables it, and that is ok. A dev can enable if they need local troubleshooting. However, if www-dev-ac was also www-cms-dev-ac or something and did not turn off aggregation, we would have encountered this issue MUCH sooner. 

Also locally once aggregation was enabled, it was pretty quick to test www-cms.devbox and verify the routes selected were correct. (Because it was reproducible easily with a `drush cr` and refresh the login page)

Closes #4305

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_core/src/EventSubscriber/CmsProtectionSubscriber.php
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_core/src/EventSubscriber/CmsProtectionSubscriber.php
@@ -151,6 +151,8 @@ class CmsProtectionSubscriber implements EventSubscriberInterface {
       'samlauth.saml_controller_acs',
       'system.403',
       'system.401',
+      'system.css_asset',
+      'system.js_asset',
     ];
     return in_array($route_name, $login_routes);
   }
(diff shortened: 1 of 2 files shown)


blairlearn (approved):

bryanpizzillo:
Kate is having problems testing locally, so it looks like we just need to merge and deploy to ACSF. I tested it locally and confirmed I can get the jQuery error, and I can also fix the jQuery error with what I did.

PR #4331, opened 2024-04-29
(#2745) Cache clear when redirect is created

### LINKED ISSUE #2745: Redirects Not Working because 404s are cached
### Issue description
Adding a URL redirect directly in Drupal (i.e., **not** using the bulk import manager) _should_ take effect immediately. But there have been times when the new redirect doesn't work and gives a 404. The issue has something to do with 404s being cached for a long time in Drupal and the cache not getting cleared on an update (meaning, someone tries to request the page before the redirect is in place).
The cache information from Akamai, and Varnish says that there was a “Dynamic Page Cache” hit which means that Drupal cached the original 404, and upon the re-request, the other caches cleared but the dynamic page cache did not.

> **ESTIMATE**

I think the problem occurs with:
1.  View a URL and get a 404
2. Create a redirect for that URL
3.  Try and view the URL and still get the 404

- When a redirect is added, it should work, redirecting the old URL to the new one. It should not matter if someone has tried to request the page before the redirect is in place. The redirect should work once it is added.

### What's the actual result?
- a Dynamic Page Cache hit, meaning someone requested the new URL before the redirect was created in the system. That 404 is then cached in the system, and the redirect, once it's created, gives you a 404 page error.

- [Instructions for adding redirects](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Redirects#adding-a-redirect-directly-in-drupal)(Specifically, the "Adding a Redirect Directly in Drupal" section.)

### Related Tickets

- #4095 Related, or a way to deal with this ticket.
- #1866 Required to clear all the way through Akamai

# Testing

- Go to ACSF environment and log in to CMS
- Pick a page URL that doesn't exist (e.g. `https://www-test-acsf.cancer.gov/my-nonexistant-page`) and go to it; get 404
- Create a redirect (e.g. `my-nonexistant-page`) and point it to anything you want
- Go to the URL after you create the redirect
    - Expected: Redirected
    - Actual: 404
    - Workaround: `drush pins https://www-test-acsf.cancer.gov/my-nonexistant-page` and then go to Akamai control panel and clear cache for that URL; then it will work
    
After the fix is in place, you should get redirected without the workaround. A ~10-30 second pause between the redirect creation and trying to go to it is recommended in order to give the cache clear time to propagate 

For testing purposes, these are the cache clears that each action in the redirects module should incur:

- Creating URL:
    - One cache clear per language for the _from_ URL of the redirect
- Editing URL:
    - One cache clear per language for the new _from_ URL of the redirect
    - One tag cache clear for `redirect-X` where `X` is the Drupal ID number of the redirect (this would clear out the _old_ from URL)
- Deleting URL:
    - One tag cache clear for `redirect-X`

A member of the Ops team can check that these actually went through in the Akamai purge reports.

navsunka:
Closes #2745

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.module
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.module
@@ -22,6 +22,9 @@ use Drupal\views\Plugin\views\query\QueryPluginBase;
 use Drupal\Core\Cache\Cache;
 use Drupal\Core\StringTranslation\TranslatableMarkup;
 use Drupal\Component\Render\FormattableMarkup;
+use Drupal\Core\Site\Settings;
+use Drupal\redirect\Entity\Redirect;
+use Drupal\Core\Language\Language;
 
 /**
  * Implements hook_block_build_alter().
@@ -847,4 +850,48 @@ function cgov_core_taxonomy_term_delete(EntityInterface $entity) {
       }
     }
   }
+}
+
+/**
+ * Implements hook_ENTITY_TYPE_insert().
+ */
+function cgov_core_redirect_insert(Redirect $entity) {
+  $moduleHandler = \Drupal::service('module_handler');
+  if (!$moduleHandler->moduleExists('purge')) {
+    return;
+  }
+    // The value for cgdp.cache_url_host is computed in cgov_core.settings.php.
+    $host = Settings::get('cgdp.cache_url_host');
+    $host = rtrim($host, '/');
+    $sourcePath = '/' . $entity->redirect_source->path;
+    $purgeInvalidationFactory = \Drupal::service('purge.invalidation.factory');
+    $purgeProcessors = \Drupal::service('purge.processors');
+    $purgePurgers = \Drupal::service('purge.purgers');
+    $processor = $purgeProcessors->get('drush_purge_invalidate');
+    if (empty($processor)) {
+      return;
+    }
+    $language = [$entity->language->value];
+    // If no language is specified for the redirect,
+    // purge the url for all languages.
+    if ($entity->language->value === Language::LANGCODE_NOT_SPECIFIED) {
+      $langcodes = \Drupal::languageManager()->getLanguages();
+      $language = array_keys($langcodes);
+    }
+
+    $invalidations = [];
+    // Creating a absolute url for all languages based on $language,
+    // these absolute url will pass to purger to invalidate cache of url.
+    foreach ($language as $value) {
+      $language_prefix = \Drupal::config('language.negotiation')->get("url.prefixes.$value");
+      if (!empty($language_prefix)) {
+        $sourcePath = Url::fromUri("internal:/{$language_prefix}{$sourcePath}")->toString();
+      }
+      $invalidateUrl = $host . $sourcePath;
+      $invalidations[] = $purgeInvalidationFactory->get('url', $invalidateUrl);
+    }
+    if (!empty($invalidations)){
+      $purgePurgers->invalidate($processor, $invalidations);
+    }
+
 }
--- a/docroot/sites/settings/cgov_core.settings.php
+++ b/docroot/sites/settings/cgov_core.settings.php
@@ -37,10 +37,16 @@ if ($is_acsf_env && $acsf_db_name) {
       $domain = $site_name;
     }
   }
+  $settings['cgdp.cache_url_host'] = 'https://' . $domain;
   $config['simple_sitemap.settings']['base_url'] = 'https://' . $domain;
 } else {
   // NOTE: you can override this in your local.settings.php.
   $config['simple_sitemap.settings']['base_url'] = 'https://www.cancer.gov';
+  // Setting this cgdp host as a current host.
+  // This will work for non ACSF environments.
+  // Using $_SERVER variable because drupal services aren't
+  // available in settings.php.
+  $settings['cgdp.cache_url_host'] = 'https://' . $_SERVER['HTTP_HOST'];
 }
 
 // Get our current environment


docroot/profiles/custom/cgov_site/modules/custom/cgov_core/src/EventSubscriber/ResponseCacheSubscriber.php
```diff
@@ -63,6 +64,18 @@ public function setHeaderCacheControl(ResponseEvent $event) {
...
 
+  /**
+   * Clear rendered cache on save of url redirect.
+   */
+  public function redirectCacheResponse(ResponseEvent $event) {
+    $response = $event->getResponse();
+    // Check if the response is a redirect.
+    $route = $event->getRequest()->attributes->get('_route');
+    if ($response instanceof RedirectResponse && ($route === 'entity.redirect.canonical' || $route === 'redirect.add')) {
+      drupal_flush_all_caches();
```
bryanpizzillo, line 75: This should not be used in our code. I saw this for another item. This has a huge performance impact when this happen.

bryanpizzillo (changes requested):
It think this would be better served by adding a hook_update on a redirect and calling the purge module to purge the URL. We can discuss if you would like.

Sorry that the last comment from the ticket was not clear - what it should say is "call the Purge module to purge the URL":
> When a redirect is created, call the purge. Test this on ACSF. Redirects to work after 15 minutes.

bryanpizzillo (changes requested):
The try/catch block is the biggest issue with the code below. This is not to say to just remove it.

docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.module
```diff
@@ -735,3 +739,49 @@ function cgov_addl_lib_default_value_function(EntityInterface $entity, FieldDefi
...
+          throw new \Exception($e->getMessage());
+        }
+        catch (CapacityException $e) {
+          // Capacity exceptions happen when too much was purged during this request.
+          throw new \Exception($e->getMessage());
+        }
+        catch (LockException $e) {
+          // Lock exceptions happen when another code path is currently processing.
+          throw new \Exception($e->getMessage());
+        }
```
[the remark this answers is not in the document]
navsunka: I just referred to the code written in the contributed module but ideally, we are invalidating only one URL which might not throw any exception. so I removed the 'try-catch' as of now.

santwuma:
@navsunka can you please rebase this?

docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.module
```diff
@@ -735,3 +739,36 @@ function cgov_addl_lib_default_value_function(EntityInterface $entity, FieldDefi
     ['value' => 'None'],
   ];
 }
+
+
+function cgov_core_redirect_insert($entity) {
+  $moduleHandler = \Drupal::service('module_handler');
+  if ($moduleHandler->moduleExists('purge')) {
+    // host is getting set from sites/settings/cgov_core.settings.php.
+    $host = Settings::get('cgdp.cache_url_host');
```
blairlearn, line 748: Not to be overly nit-picky, but `$host` is set right there on line 747.  The _value_ for `cgdp.cache_url_host` is computed in `cgov_core.settings.php`.

blairlearn (changes requested):

docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.module
```diff
@@ -848,3 +852,46 @@ function cgov_core_taxonomy_term_delete(EntityInterface $entity) {
...
+    }
+    $language = [$entity->language->value];
+    $sourcePath = '/' . $entity->redirect_source->path;
+    $langcodes = \Drupal::languageManager()->getLanguages();
+    // If und set means redirect is created for all languages
+    // in that case we need to purge url for all languages
+    // installed in system.
+    if ($entity->language->value === 'und'){
+      $language = array_keys($langcodes);
+    }
```
blairlearn, line 882: First, please move the `$sourcePath` assignment up a line.  That's unrelated to the language concerns and shouldn't interrupt them.

The comment is somewhat inaccurate.  The `'und'` value (it's actually `Language::LANGCODE_NOT_SPECIFIED`) means that _no language_ has been specified.  So, something along the lines of "If no language is specified for the redirect, purge the url for all languages."

Instead of `'und'`, this should be `Language::LANGCODE_NOT_SPECIFIED`.
Move the `$langcodes` assignment into the `if` block, since that's the only place it's needed.

docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.module
```diff
@@ -848,3 +852,46 @@ function cgov_core_taxonomy_term_delete(EntityInterface $entity) {
...
+    $host = rtrim($host, '/');
+    $invalidations = [];
+    foreach ($language as $value) {
```
blairlearn, line 885: And again, a comment explaining what this block of code is doing. Not a line-by-line breakdown, just the purpose of this loop.

blairlearn (changes requested):

blairlearn (approved):

alinai:
Regression Test passed on ODE781. Need to verify it manually.

KateMashkinaNIH:
Tested Blair's specific test case - it still results in 404 in the end

https://github.com/NCIOCPL/cgov-digital-platform/issues/2745#issuecomment-2226167633

docroot/sites/settings/cgov_core.settings.php
```diff
@@ -37,6 +37,7 @@
       $domain = $site_name;
     }
   }
+  $settings['cgdp.cache_url_host'] = 'https://' . $domain;
```
bryanpizzillo, line 40: @navsunka (cc: @blairlearn ) This is probably why testing is failing on www-dev-ac. That www-dev/test/int-ac, nor odes, would satisfy the check on line 31, `if ($is_acsf_env && $acsf_db_name) {`.

So cgdp.cache_url_host will not be set when you do the code in the hook:
```
   // The value for cgdp.cache_url_host is computed in cgov_core.settings.php.
    $host = Settings::get('cgdp.cache_url_host');
    if (empty($host)) {
      return;
    }
```

Which means the cache will not be cleared.

navsunka: if $host is empty, we will then fetch the current base URL which will work for lower environments.

docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.module
```diff
@@ -848,3 +851,50 @@ function cgov_core_taxonomy_term_delete(EntityInterface $entity) {
...
+function cgov_core_redirect_insert(Redirect $entity) {
+  $moduleHandler = \Drupal::service('module_handler');
+  if (!$moduleHandler->moduleExists('purge')) {
+    return;
+  }
+    // The value for cgdp.cache_url_host is computed in cgov_core.settings.php.
+    $host = Settings::get('cgdp.cache_url_host');
+    if (empty($host)) {
+      $host = \Drupal::request()->getSchemeAndHttpHost();
+    }
```
blairlearn, line 867: Please remove this if block and set `cgdp.cache_url_host` to the appropriate value in `cgov_core.settings.php`.  (See my other comment.)

blairlearn (changes requested):

blairlearn:
After speaking with @bryanpizzillo about this, we realized changes are needed in order to accommodate drush and cron

First, `cgdp.cache_url_host` should be set in `cgov_caching.settings.php` instead of `cgov_core.settings.php`.

The entirety of `cgov_caching.settings` is the `true` case for an `if` statement. When the condition is `true`, the code is executing in an Acquia environment. (More on that in moment.)

Add an `else` clause where `cgdp.cache_url_host` is set to `'http://www.devbox'`.  That's the only non-Acquia environment we're going to support.

Now, back to the true clause:

Find this block of code (lines 37-40 in the current code):

```php
  // Setup proper .edgerc path for Akamai module
  $ah_group = isset($_ENV['AH_SITE_GROUP']) ? $_ENV['AH_SITE_GROUP'] : NULL;
  $config['akamai.settings']['edgerc_path'] = "/mnt/gfs/home/$ah_group/common/.edgerc";
```

Immediately after this, you're going to add
```php
if ($is_acsf_env && $acsf_db_name) {
   // Look up the ACSF domain.  (INCLUDE THIS LINE AS A COMMENT!!!)
   // You may have to duplicate some logic from cgov_core.settings.php.
} elseif (file_exists('/var/www/site-php') && isset($_ENV['AH_SITE_ENVIRONMENT'])) {
   // Determine non-ACSF Acquia environment.    (INCLUDE THIS LINE AS A COMMENT!!!)
   // See below for the actual logic.
}
```

The logic for the non-ACSF Acquia environments is based on the `$env` variable, which is set at the start of the outer `if` block.

 if `$env` is: | set `cgdp.cache_url_host` to
  ---: | ---
  dev | https://www-dev-ac.cancer.gov
  int | https://www-int-ac.cancer.gov
  test | https://www-test-ac.cancer.gov
  ode | `'https://' . $_ENV['AH_SITE_NAME'] . 'prod.acquia-sites.com'`

jfrank-nih:
Commit has been pulled into PR #4466. Closing this PR.

PR #4368, opened 2024-06-17
(#4367) Change Feature Card Image to 16x9

### LINKED ISSUE #4367: Change Feature Card Image Size on Home and Landing Pages to 16x9
### Issue description
**Request:** Change the image size of the 3 Feature Card Row on Home and Landing Pages to a 16x9 image from 4x3 image. 

**Context:** as part of preparation for the upcoming Redesign Release 3, the team conducted a design phase that included creation of a 3 Feature Card Row to be used on Mini Landing Pages (MLPs). It was determined we will use a 16x9 image on the MLPs for all breakpoints in order to reduce wasted vertical space. It was determined to be advantageous to make this same change on the 3 Feature Card Row for Home and Landing Pages which currently use a 4x3 image, making the component taller and pushing down content on the page which is garnering negative feedback. By making this change, it will also save us time in the future when we create the 3 Feature Card Row for MLPs.

NOTES
- Documentation on the Drupal Guide on MyNCI should be updated accordingly after this change is made.
- Approved by Anna/Alex via email on 06/11/24.

- The 3 Feature Card Row on Home and Landing Pages will use a 16x9 image as opposed to a 4x3. 

- The 3 Feature Card Row on Home and Landing Pages uses a 4x3 image. 

### Updated Acceptance Criteria
**Adding Feature Cards to Home & Landing Page**

```gherkin
Scenario: A content editor adds a Feature Card Row to a Home & Landing Page on the CMS
  Given a content editor is logged into the CMS
    And they are creating or editing a Home & Landing Page
  When they add a Feature Card Row to the Home & Landing Page
  Then there will be no change to the user experience from the current functionality
```

**Viewing a Home & Landing Page with a Feature Card Row**

```gherkin
Scenario: A site visitor views a Home & Landing Page that has a Feature Card Row on it
  Given a site visitor is on a Home & Landing Page
    And the Home & Landing Page they are visiting has a Feature Card Row
    And the Feature Card Row has an Internal Feature Card, and External Feature Card, and a Multimedia Card to test all use cases
  When the feature cards have Override Images added
  Then the override image will appear on the feature card at an aspect ratio of 16:9 for all breakpoints above mobile large (breakpoint > 639px)
  When the internal feature card and multimedia feature card have no override image
  Then the internal feature card's and multimedia feature card's respective promotional images will appear on the the respective cards at an aspect ratio of 16:9 for all breakpoints above mobile large (breakpoint > 639px)
  When the breakpoint is less than tablet (640px > breakpoint)
  Then no image will be displayed on the feature card
  When the breakpoint is greater than tablet (breakpoint > 879px) 
  Then the cards will be horizontally aligned with image aspect ratios of 16:9 each
  When the breakpoint is less than tablet large (880px > breakpoint)
  Then the cards will be vertically stacked with image aspect ratios of 16:9 each
```

dev-rana-publicis:
Closes #4367
ODE: https://ncigovcdode771.prod.acquia-sites.com/


andyvanavery31:
Could we add the external feature card and the multimedia card so we can confirm if they are working properly? Also for the internal feature cards and multimedia cards, we need to check that in the case of them using the override image or the promotional image both work properly.

andyvanavery31 (approved):
Passes Product review.

laurelthrash:
Currently in review by product (approved above) and QA (forthcoming). After approved by product, will go through code review.

bryanpizzillo (changes requested):
Two main things:
### 1. One cannot simply remove a config 
You can't only delete `docroot/profiles/custom/cgov_site/modules/custom/cgov_image/config/install/image.style.ncids_featured_4x3.yml`. Removing a config from config/install makes it so new installations do not have that file - existing sites will continue to have it.

So you need to make a update hook in cgov_image.install that does 2 things:
1. Import the features so no config is referencing the ncids_featured_4x3 image style.
2. Delete the ncids_featured_4x3 image style.

### 2. ~~The picture elements should not exist anymore.~~
So since we just have a single `` element, we should not use `<picture>` anymore. So you will want to add a little CSS to fix this up, which of course needs to be backported to NCIDS when the 2 images are removed there.

So the easiest thing to do for now is to:
1. move the `` tag to be a direct child of the `<a>` tag.
2. add the `nci-card__image` class to the ``
    * Remember it must hide on mobile, and nci-card__image makes that happen 
4. Add the following css to `.nci-card__image` selector as a hack in the cgov feature card scss
    ```
    width: 100%;
    height: auto;
    // This should match the transform for scale, but that is applied to img in a card.
    // Don't exactly know why the hover zoom does not work on tablet...
    transition: transform .5s ease;
    ```
 5. Add the hacks you did to make it work to the NCIDS ticket that is for removing the picture element. If it does not exist, yell at someone to make it.

In NCIDS Storybook, Just moving the img outside of the picture element will break the images from 571-879 because the image does not get width:100%. This is most likely due to the fact that the 16x9 images now have to be larger to fill the larger space. I would think since tablet-tablet-lg is the only place we had 16x9 images, then cgdp should be ok with image sizes.

laurelthrash:
Moved from "Ready for Testing" to "In Progress." Dev to review/address feedback from Bryan above before reviewed is conducted by QA.

bryanpizzillo (changes requested):
@dev-rana-publicis - can you please add 
> nci-card has been changed to not force a width for cards. So the 3 feature and guide card rows need to be updated to add grid-col-4 to the cards/li elements. IDK if the guide cards with image and description are affected.

To this ticket?

bryanpizzillo:
This ticket was cherry-picked into the `feature/redesign-r3.1` feature branch instead of getting pulled into `develop` at that time. It was released in https://github.com/NCIOCPL/cgov-digital-platform/releases/tag/v3.7.0. So if you are looking for PR review comments, this is the place to look.

Issue #175, opened 2018-12-12
Validation for Pretty URL and Site Structure


nicolecrom:
This story is to create code to validate pretty URLS entered by users in both common fields and site sections. Related to issue #174 and #176 .  We would like to use the same validator for both functions if possible.

[Validation - for Common Field Use](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Common-Fields#pretty-url)
- No capital letters
- Only allows 0-9, a-z, and - (dash)
- When preceded by the parent section's pretty URL path, the full alias must be unique within the site
  - This includes validating against any microsite (future requirement)

[Validation for Site Section use](https://github.com/NCIOCPL/cgov-digital-platform/wiki/Vocabularies-Site-Sections#pretty-url):
* No capital letters
* Only allows 0-9, a-z, and - (dash)
* When preceded by the parent section's pretty URL path, the full alias must be unique within the site
   * This includes validating against any microsite (future requirement)
   * Note: will need to check if a page on the site already exists with this same pretty URL and lives within this parent section
   * Further refinements will need to be made to previous rule

PR #4378, opened 2024-07-08
(#3170) sitesection must have unique url.

### LINKED ISSUE #3170: Site section under a parent site section must have an unique pretty url
### Issue description
*Description of the issue*
When preceded by the parent section's pretty URL path, site section should not be allowed to have duplicated Purls
> **ESTIMATE** TBD

1.  Go to Structure -> taxonomy -> site sections in cms
2. select home's children
2.  modify one of the children to have same Purl as the other chid under same parent 
3.  save and - system allows it 

- the system should error that only unique Purls are allowed and not let the section to be saved
- The error message should be, "The section is unable to be saved. Please enter a unique pretty url."

### What's the actual result?
- it saves with no problem which results in duplicate site section urls

- Screenshot
Screen Shot 2021-09-16 at 5 24 59 PM

### Related Tickets
* Original ticket with requirements where this bug was missed https://github.com/NCIOCPL/cgov-digital-platform/issues/175
*  PR that had a comment mentioning that this issue was noted and will need a bug ticket to fix it later https://github.com/NCIOCPL/cgov-digital-platform/pull/589

navsunka:
Closes #3170

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/cgov_site_section.module
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/cgov_site_section.module
@@ -776,8 +776,13 @@ function cgov_site_section_entity_bundle_field_info_alter(&$fields, EntityTypeIn
   }
   if ($entity_type->id() == 'taxonomy_term'  && $bundle === 'cgov_site_sections') {
     if (isset($fields['parent'])) {
-      // Add custom constraint to this field
+      // Add custom constraint to validate parent term combinations.
       $fields['parent']->addConstraint('CgovValidateParentConstraint', []);
+    }
+    if (isset($fields['field_pretty_url'])) {
+      // Add custom constraint for pretty url to check the uniqueness of url
+      // based on computed path.
+      $fields['field_pretty_url']->addConstraint('UniqueCgovUrl');
     }
   }
 }
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/src/Plugin/Validation/Constraint/ValidateParentConstraintValidator.php
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/src/Plugin/Validation/Constraint/ValidateParentConstraintValidator.php
@@ -135,14 +135,14 @@ class ValidateParentConstraintValidator extends ConstraintValidator implements C
    *   Boolean.
    */
   public function checkChildUniqueName($formData, $parentItems) {
-    $name = $formData['name'][0]['value'];
+    $name = strtolower($formData['name'][0]['value']);
     $langcode = $formData['langcode'][0]['value'];
     $tid = !empty($formData['tid']) ? $formData['tid'][0]['value'] : '';
     if (count($parentItems) === 1) {
       $vid = 'cgov_site_sections';
       $terms = $this->termStorage->loadTree($vid, $formData['parent'][0]['target_id'], 1);
       foreach ($terms as $term) {
-        if ($term->name === $name && $term->tid !== $tid && $langcode === $term->langcode) {
+        if (strtolower($term->name) === $name && $term->tid !== $tid && $langcode === $term->langcode) {
           return FALSE;
         }
       }
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/src/Plugin/Validation/Constraint/UniqueCgovUrlConstraint.php
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/src/Plugin/Validation/Constraint/UniqueCgovUrlConstraint.php
@@ -26,6 +26,13 @@ class UniqueCgovUrlConstraint extends CompositeConstraintBase {
   public $sectionInUse = 'This section already has a landing page.';
 
   /**
+   * Unique pretty url.
+   *
+   * @var string
+   */
+  public $uniquePrettyUrl = 'The section is unable to be saved. Please enter a unique pretty url.';
+
+  /**
    * {@inheritdoc}
    */
   public function coversFields() {
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/src/Plugin/Validation/Constraint/UniqueCgovUrlConstraintValidator.php
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/src/Plugin/Validation/Constraint/UniqueCgovUrlConstraintValidator.php
@@ -3,10 +3,12 @@
 namespace Drupal\cgov_site_section\Plugin\Validation\Constraint;
 
 use Drupal\Core\DependencyInjection\ContainerInjectionInterface;
+use Drupal\Core\Entity\ContentEntityInterface;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Field\EntityReferenceFieldItemListInterface;
 use Drupal\Core\Field\FieldItemListInterface;
+use Drupal\taxonomy\Entity\Term;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\Validator\Constraint;
 use Symfony\Component\Validator\ConstraintValidator;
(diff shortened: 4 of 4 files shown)


docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/src/Plugin/Validation/Constraint/ValidatePrettyUrlConstraint.php
```diff
@@ -0,0 +1,24 @@
...
+ * )
+ */
+class ValidatePrettyUrlConstraint extends Constraint {
+
+  /**
+   * Unique pretty url.
+   *
+   * @var string
+   */
+  public $uniquePrettyUrl = 'The site section cannot be saved. Site sections must have unique pretty url.';
```
blairlearn, line 22: @andyvanavery31 and @sarinapadilla - would one of you please chime in with text for the error message?  [The original ticket](https://github.com/NCIOCPL/cgov-digital-platform/issues/175) only specifies that they must be unique.

docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/src/Plugin/Validation/Constraint/ValidatePrettyUrlConstraintValidator.php
```diff
@@ -0,0 +1,72 @@
...
+      // Exclude the current term if it exists.
+      if (!$entity->isNew()) {
+        $query->condition('tid', $entity->id(), '!=');
+      }
+      $result = $query->execute();
+      if (!empty($result)) {
+        $this->context->addViolation($constraint->uniquePrettyUrl);
+      }
+    }
+  }
```
blairlearn, line 70: Should this be returning something?  Line 51 above returns `NULL` if `$constraint` isn't the right type.

blairlearn (changes requested):

docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/src/Plugin/Validation/Constraint/ValidatePrettyUrlConstraint.php
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * A composite constraint validating Site Section pretty url.
+ *
+ * @Constraint(
+ *   id = "CgovValidatePrettyUrlConstraint",
+ *   label = @Translation("ValidatePrettyUrlConstraint", context="Validation"),
+ * )
+ */
+class ValidatePrettyUrlConstraint extends Constraint {
```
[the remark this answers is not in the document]
navsunka: I have changed the name to UniquePrettyUrl and this UniqueCgovUrlConstraint is checking for node type not for taxonomy.

docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/src/Plugin/Validation/Constraint/ValidatePrettyUrlConstraintValidator.php
```diff
@@ -0,0 +1,72 @@
...
+use Drupal\Core\DependencyInjection\ContainerInjectionInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidator;
+
+/**
+ * Validates the ValidatePrettyUrlConstraintValidator constraint.
+ */
+class ValidatePrettyUrlConstraintValidator extends ConstraintValidator implements ContainerInjectionInterface {
```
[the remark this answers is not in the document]
navsunka: I have changed the name to UniquePrettyUrlValidator and this UniqueCgovUrlConstraintValidator is checking for node type not for taxonomy.

docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/src/Plugin/Validation/Constraint/ValidatePrettyUrlConstraintValidator.php
```diff
@@ -0,0 +1,72 @@
...
+   */
+  public function validate($field, Constraint $constraint) {
+
+    if (!($constraint instanceof ValidatePrettyUrlConstraint)) {
+      return NULL;
+    }
+
+    $entity = $field->getEntity();
+    $prettyUrl = $entity->field_pretty_url->value;
+    if (!empty($prettyUrl)) {
```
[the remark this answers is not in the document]
navsunka: Fixed.

docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/src/Plugin/Validation/Constraint/ValidatePrettyUrlConstraintValidator.php
navsunka: It should not written anything, I have written same as in drupal core constraint validators.

docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/src/Plugin/Validation/Constraint/ValidatePrettyUrlConstraint.php
andyvanavery31: @navsunka @blairlearn updated the ticket #3170 with text for the error message in the Expected Results section.

docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/src/Plugin/Validation/Constraint/ValidatePrettyUrlConstraintValidator.php
blairlearn: You are correct, it should not be returning anything.  Looking at how the `validate` method is declared in `Symfony\Component\Validator\ConstraintValidatorInterface`
([link](https://github.com/symfony/symfony/blob/v6.3.12/src/Symfony/Component/Validator/ConstraintValidatorInterface.php#L32)), the method is void and arguably it's the return NULL which should really just be `return`.

docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/cgov_site_section.module
```diff
@@ -776,9 +776,13 @@ function cgov_site_section_entity_bundle_field_info_alter(&$fields, EntityTypeIn
   }
   if ($entity_type->id() == 'taxonomy_term'  && $bundle === 'cgov_site_sections') {
     if (isset($fields['parent'])) {
-      // Add custom constraint to this field
+      // Add custom constraint to this field.
```
blairlearn, line 779: Please replace this comment with something more descriptive. A good comment should explain the _intent_ of the code.  In this case, what rule is being validated? That a parent exists? That there's only one?

docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/cgov_site_section.module
```diff
@@ -776,9 +776,13 @@ function cgov_site_section_entity_bundle_field_info_alter(&$fields, EntityTypeIn
...
       $fields['parent']->addConstraint('CgovValidateParentConstraint', []);
     }
+    if (isset($fields['field_pretty_url'])) {
+      // Add custom constraint to pretty url.
+      $fields['field_pretty_url']->addConstraint('CgovValidatePrettyUrlConstraint', []);
```
blairlearn, line 784: Same thing as above.

blairlearn (changes requested):
The new constraint classes duplicate the functionality of the existing `UniqueCgovUrlConstraint` and `UniqueCgovUrlConstraintValidator` classes.

Technically, yes, the existing classes validate for `node` instead of `taxonomy_term`, however they are enforcing the same rule, just for different entity types.  This should be done with the existing classes.

The new classes also introduce a bug where only a single site section can have a given pretty url value, regardless of where it exists in the hierarchy.  (e.g. You cannot create a site section with a pretty url value of "funding" under "news' if another site section with the value of "funding" already exists under "research".)

The validation logic should be along the lines of

1. Fetch the parent `cgov_site_section`
2. Get the parent's `computed_path`
3. Append the proposed `field_pretty_url` value
4. Check for the existence of any other `cgov_site_section` with that same `computed_path`.

You'll also need to make certain the new `computed_path` doesn't collide with the `computed_path` for anything else.

A specific scenario to check:

1. I make `/foo` site section
2. I make `/foo/bar` node, where the node has `bar` as its `field_pretty_url`
3. I then go make `/foo/bar` site section where `bar` is `field_pretty_url`

KateMashkinaNIH:
this scenario does not pass - it allowed me to create a site section 
Scenario: System does not allow the user to create a site section where the site section pretty URL matches another node item's site section path + pretty URL
  Given the Research site section with path /research exists 
    And a node item within Research site section with pretty URL set to "areas" exists
    And the user is creating a new site section within Research site section
  When the user sets the site section pretty URL to "areas" and tries to save
  Then the system will not allow the user to save the site section

KateMashkinaNIH:
one more thing - the site section name should be case insensitive, e.g. About Cancer and About cancer should be treated the same and system should not allow those two section to be created with the same purl

blairlearn (changes requested):

docroot/profiles/custom/cgov_site/modules/custom/cgov_site_section/src/Plugin/Validation/Constraint/ValidateParentConstraintValidator.php
```diff
@@ -135,14 +135,14 @@ public function getLanguageRoots() {
    *   Boolean.
    */
   public function checkChildUniqueName($formData, $parentItems) {
-    $name = $formData['name'][0]['value'];
+    $name = strtolower($formData['name'][0]['value']);
```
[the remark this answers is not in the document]
blairlearn: From [Slack conversation](https://ncisape.slack.com/archives/D051JFA5LB1/p1729013864347149?thread_ts=1729010782.098829&cid=D051JFA5LB1), this change is related to the requirement that pretty url uniqueness should be case-insensitive (e.g. `chicken` and `Chicken` are treated as the same pretty  url).

blairlearn (approved):

PR #4387, opened 2024-07-16
(#4386) Update landing page analytics to not be only for LP

### LINKED ISSUE #4386: Update Landing Page analytics functions to support better reuse across page types
> [!NOTE]
> This issue was found on the feature/redesign-3.1 (i.e. Mini-landing Page release) branch while testing a ticket for content blocks. RawHTML is affected as well, and unfortunately, was discovered after it was merged. 

The landing page analytics helper, `landingClickTracker` is setup to handle the analytics tracking information for landing page items. With the introduction of mini-landing pages, we need to change how this works because mini-landing pages requirement a different set of events for the same feature. As there are multiple components that are shared between home/landing and mini-landing pages, we need to update the landingClickTracker so that we do not have to add logic to all the components to figure out what page they are on. 

Add the following data elements:
* `pageType` - comes from the `dcterms.type` `<meta>` tag.
* `pageTemplate` - comes from the `cgdp.template` `<meta>` tag.

Additionally, the `LP:` at the beginning of the EDDL event name needs to be defined based on the `pageType` parameter. Currently only Home/Landing and Mini-landing pages can use a component that would use these analytics. So we only need to configure these two types for now. The mapping is:
* when `pageType` is `cgvHomeLanding` the beginning will be `LP:`
* when `pageType` is `cgvMiniLanding` the beginning will be `MLP:`
* if `pageType` is neither of the above the beginning will be `UNKNOWN:`

> [!NOTE]
> **Technical Note:** please handle the beginning check in such a way as to us only needing to add a type, as in a switch statement, instead of something else where if we need to enable some new page type, we have to waste time redoing and fixing the check.

Right now the `pageType` data is not sent and the event names are `LP:XXX` even though they should be `MLP:XXX` when on an MLP. (As the current requirements state)

### What's the updated acceptance criteria?
```gherkin
```

None

### Related Tickets
* #3786
   * This added the helper
* #4013
   * This updated the helper to better support home/landing components by adding columns.

bryanpizzillo:
MLP can be tested at https://ncigovcdode784.prod.acquia-sites.com/test/raw-html-sample-page. Home/landing live at the normal locations.

ODE: https://ncigovcdode784.prod.acquia-sites.com/
Closes #4386

### DIFF
--- a/docroot/profiles/custom/cgov_site/themes/custom/ncids_trans/front-end/src/entrypoints/addl-cancer-types-landing-page/components/cgdp-cancer-type-az-list/cgdp-cancer-type-az-list.ts
+++ b/docroot/profiles/custom/cgov_site/themes/custom/ncids_trans/front-end/src/entrypoints/addl-cancer-types-landing-page/components/cgdp-cancer-type-az-list/cgdp-cancer-type-az-list.ts
@@ -1,4 +1,8 @@
-import { getLandingRowsAndColsInfo } from '../../../../lib/core/analytics/landing-page-contents-helper';
+import {
+	getLandingRowsAndColsInfo,
+	getPageInfo,
+	getEventNameBeginning,
+} from '../../../../lib/core/analytics/landing-page-contents-helper';
 import { trackOther } from '../../../../lib/core/analytics/eddl-util';
 import { USAAccordion } from '@nciocpl/ncids-js/usa-accordion';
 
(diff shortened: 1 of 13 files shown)


olitharp-nci (approved):

welshja:
Tested on rawHTML block on sample page as well as rawHTML block on News & Events page on the ODE. Looks to be working as expected:
1. EDDL linknames were "MLP" for MLP and "LP" for LP.
2. Page Type and Page Template captured proper values for each.

PR #4438, opened 2024-09-10
(#4437) Exclude additional files from built artifact

### LINKED ISSUE #4437: Exclude additional unwanted files from built artifacts
We want to exclude additional unwanted files from the final built artifacts. These are mostly things added by Drupal at build time, though may also include some testing artifacts.

```
docroot/README.md
docroot/example.gitignore
```

Possibly exclude all `docroot/**/README.md` and `README.txt`.

jfrank-nih:
Closes #4437


bryanpizzillo (approved):
This looks good but it needs to be tested on ACSF

PR #4424, opened 2024-08-19
(#4412) Header Skips Event Listing Fix

karlikpj:
Closes #4412

PR #4466, opened 2024-10-08
(#1866, #2745) Akamai cache clearing per URL and redirect purges

### LINKED ISSUE #1866: Determine what to do about Akamai module's basepath setting
The base path can be used with a common hostname for purging. Unfortunately it is a setting that will vary across environments. Additionally, some environment will serve up content under 2 names. (Microsite URL under cancer.gov and admin interface.)

We really should not cache based on hostname and URLs, and if we need to we can use the Akamai interface.

## Testing

- Needs to be tested on AC and ACSF
- Run standard regressions
- Test #2745
- `/sitemap.xml` or `/sitemaps/pageinstructions.xml` (depending on environment) should have URLs with the correct domain on AC now
- Verify that the meta tag `cgdp.domain` is correct (should reflect the domain and, if on ACSF, site)
- Make sure SSO still works
- Make sure WYSIWYG editor still shows up (i.e. license file is found)

### LINKED ISSUE #2745: Redirects Not Working because 404s are cached

jfrank-nih:
A couple of tickets that require messing with the settings files:

* Enables cache clearing by URL in Akamai
* Purges redirect URLs at redirect creation

Closes #1866
Closes #2745

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -220,7 +220,10 @@
                 "2834130 : Updates Features Diff to Include Alters": "https://www.drupal.org/files/issues/2019-06-09/features-detect-overrides-update-2834130-4-D8.patch"
             },
             "drupal/purge": {
-                "3094343 : Garbled purge tag IDs": "https://www.drupal.org/files/issues/2023-09-07/purge-3094343-15.patch"
+                "3094343 : Garbled purge tag IDs": "https://www.drupal.org/files/issues/2023-09-07/purge-3094343-15.patch",
+                "3460094 : Running drush p:queue-add fails (https://www.drupal.org/files/issues/2024-07-25/3460094-5-there-is-no-purger-supporting-array.patch)":  "patches/drupal/purge/3460094.diff",
+                "3460094 : Purge defines drush commands twice (https://www.drupal.org/files/issues/2024-10-18/3460094-remove_drush_services_yml.patch)":  "patches/drupal/purge/3460094-2.diff"
+
             },
             "drupal/responsive_favicons": {
                 "3376766 : Incorrect path if Drupal is installed in subfolder": "https://git.drupalcode.org/project/responsive_favicons/-/merge_requests/6.patch",
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.module
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.module
@@ -23,6 +23,9 @@ use Drupal\Core\Cache\Cache;
 use Drupal\Core\StringTranslation\TranslatableMarkup;
 use Drupal\Component\Render\FormattableMarkup;
 use Drupal\Core\Cache\RefinableCacheableDependencyInterface;
+use Drupal\Core\Site\Settings;
+use Drupal\redirect\Entity\Redirect;
+use Drupal\Core\Language\Language;
 
 /**
  * Implements hook_block_build_alter().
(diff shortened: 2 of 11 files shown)


docroot/sites/settings/cgov_caching.settings.php
```diff
@@ -7,14 +7,15 @@
  * @see https://docs.acquia.com/site-factory/tiers/paas/workflow/hooks
  */
 
+ if (!function_exists('settings_get_environment')) {
```
jfrank-nih, line 10: Since we know that `settings_get_environment` is only set in `cgov_settings_utilities.inc`, is the `function_exists` check really necessary? The `require_once` should serve to prevent it being loaded twice.

docroot/sites/settings/cgov_settings_utilities.inc
```diff
@@ -0,0 +1,69 @@
...
+    $domain = array_keys($domains)[0];
+    foreach ($domains as $site_name => $site_info) {
+      if (!empty($site_info['flags']['preferred_domain'])) {
+        $domain = $site_name;
+      }
+    }
+  // Then check if this is a different Acquia Cloud site.
+  } else {
+    $env = settings_get_environment();
+    // On local we just want to use the default.
```
jfrank-nih, line 54: Do we want to use the default (prod URL) or `www.devbox` on local? Nowhere else in our code do we hardcode `www.devbox`. It's only in the docker and Apache configuration.

docroot/sites/settings/cgov_caching.settings.php
blairlearn: That would be a safe change.

docroot/sites/settings/cgov_settings_utilities.inc
blairlearn: This is something @bryanpizzillo and I talked about a bit.  What it came down to is that the only url we're going to support for local development is `www.devbox`.

composer.json
```diff
@@ -224,7 +224,10 @@
                 "3417771 : Some styling no longer applies due to form-actions -> field-actions html change (https://git.drupalcode.org/project/paragraphs/-/merge_requests/93.diff)": "patches/drupal/paragraphs/3417771.diff"
             },
             "drupal/purge": {
-                "3094343 : Garbled purge tag IDs": "https://www.drupal.org/files/issues/2023-09-07/purge-3094343-15.patch"
+                "3094343 : Garbled purge tag IDs": "https://www.drupal.org/files/issues/2023-09-07/purge-3094343-15.patch",
+                "3460094 : There is no purger supporting array":  "patches/drupal/purge/3460094.diff",
+                "3460094 : Remove drupal.services.yml":  "patches/drupal/purge/3460094-2.diff"
```
blairlearn, line 229: This needs a link to where the patch came from.  See #4477 (issue #4424) for examples.

docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.module
```diff
@@ -848,3 +851,71 @@ function cgov_core_taxonomy_term_delete(EntityInterface $entity) {
     }
   }
 }
+
+/**
+ * Method to clear source paths for redirects when inserted or updated.
+ */
+function clearRedirectSourceCache(Redirect $entity) {
```
blairlearn, line 858: This is in the global namespace.  The convention is that helper functions such as this should have names starting with an underscore and the module name, and the rest of the name should use snake_case.  In this case, please rename the function to `_cgov_core_clear_redirect_source_cache`

docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.module
```diff
@@ -848,3 +851,71 @@ function cgov_core_taxonomy_term_delete(EntityInterface $entity) {
...
+    return;
+  }
+  $purgeQueue = \Drupal::service('purge.queue');
+  $purgeQueue->add($queuer, $invalidations);
+}
+
+/**
+ * Implements hook_ENTITY_TYPE_insert().
+ */
+function cgov_core_redirect_insert(Redirect $entity) {
```
blairlearn, line 910: Please explain the reason for implementing this hook.

docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.module
```diff
@@ -848,3 +851,71 @@ function cgov_core_taxonomy_term_delete(EntityInterface $entity) {
...
+  clearRedirectSourceCache($entity);
+}
+
+/**
+ * Implements hook_ENTITY_TYPE_update().
+ */
+function cgov_core_redirect_update(Redirect $entity) {
```
blairlearn, line 917: Same.

blairlearn (changes requested):

blairlearn (approved):

PR #4476, opened 2024-10-25
(#4475) Patch for intermittent configuration "disappearances".

### LINKED ISSUE #4475: We have had intermittent issues in Production where sites become unavailable
An example of an error we see is:
```
Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException: The "paragraph" entity type did not specify a translation handler. in Drupal\Core\Entity\EntityTypeManager->getHandler() (line 262 of /mnt/www/html/ncigov01live/docroot/core/lib/Drupal/Core/Entity/EntityTypeManager.php)
```

We have many different "gremlins" that seem to occur where the sites just "lose" configuration information. (e.g., fields disappear from content types) 

In tracking down the error above we found https://www.drupal.org/project/paragraphs/issues/3031598. The paragraphs team noted this is a core issue, which took us to https://www.drupal.org/project/drupal/issues/3207813#comment-15778426. 

It sounds like a true fix for Drupal would be handled in https://www.drupal.org/project/drupal/issues/1905334, but that is a ways away. (Such comments as "we should get rid of .module files" mean it is going to take a lot longer)

### Proposed Solution
Let's implement the fix in https://www.drupal.org/project/drupal/issues/3207813#comment-15778426 and try it out and make sure it does not break anything. Hopefully this will resolve the random issues that crop up from time to time in production.

blairlearn:
This addresses an intermittent issue with configurations being "lost" despite no code changes or deployments taking place.
See https://www.drupal.org/project/drupal/issues/3207813#comment-15778426.

Closes #4475

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -213,7 +213,8 @@
                 "3150294 : (Unofficial Patch) New translations should use initial workflow state" : "patches/drupal/workflow-hack-for-3150294.patch",
                 "1308152 : Add stream wrappers to access extension files": "https://www.drupal.org/files/issues/2021-12-15/1308152-395.patch",
                 "2971390 : Make exposure of translation meta fields conditional" : "https://www.drupal.org/files/issues/2023-08-25/2971390-75.patch",
-                "3413508 : Admin page access denied even when access is given to child items": "https://www.drupal.org/files/issues/2024-01-11/3413508-22.patch"
+                "3413508 : Admin page access denied even when access is given to child items": "https://www.drupal.org/files/issues/2024-01-11/3413508-22.patch",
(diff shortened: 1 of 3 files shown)


alinai:
When I have a published page ( front end - https://sbir-test-acsf.cancer.gov/small-business-funding/grants/test-article123) opened in a tab and I update the same content  ( back-end - https://sbir-cms-test.cancer.gov/small-business-funding/grants/test-article123)  in another tab and publish, then if I refresh the page I had opened in my first tab (front end page) I get a Page Not Found page. I have to refresh it multiple times until the published page (which was already published before the update) appears with the update.
What I understand is, I may not see the update at the front end right away on an already published page and several refresh may load the update but the already published page should continue displaying and I should not get Page Not Found as that page was already published. I got the following console error when I got Page Not Found:

image


I got the following console error when the page was loaded at the front end:

image

On incognito window:
Back end:
image

Front end:
image

blairlearn:
The various JavaScript errors are now #4492 and are unrelated.  The page not found however is a problem.

alinai:
Passed on ODE 868

PR #4480, opened 2024-10-30
(#4355) Update to Next NCIDS

karlikpj:
Closes #4355

PR #4491, opened 2024-11-05
(#4485) Wide Guide Card Line Height Fix

### LINKED ISSUE #4485: Fix Line-Height of WGC Heading
### Issue Description
The line height spacing on the Wide Guide Card (WGC) is incorrect. You can see this when the heading wraps one more than one line (see screenshot below).

- The line-height of the Wide Guide Card Title should be token 2.

- The line-height of the Wide Guide Card Title is not token 2.

- Image

### Related Tickets
* Original Story Ticket #4433

karlikpj:
Closes #4485

### DIFF
--- a/docroot/profiles/custom/cgov_site/themes/custom/ncids_trans/front-end/src/lib/components/cgdp-wide-guide-card/index.scss
+++ b/docroot/profiles/custom/cgov_site/themes/custom/ncids_trans/front-end/src/lib/components/cgdp-wide-guide-card/index.scss
@@ -28,7 +28,11 @@
 		@include u-bg('transparent');
 	}
 	.nci-guide-card__title {
-		@include typeset("heading", "xl");
+		@include typeset(
+			'heading',
+			$theme-h2-font-size,
+			$theme-heading-line-height
+		);
 		@include u-text-align('left');
 	}
 	.nci-guide-card__body {
@@ -53,11 +57,11 @@
 	}
 	.nci-guide-card__description {
 		@include u-margin-bottom(3);
-		@include typeset("body", "xs", $theme-body-line-height);
+		@include typeset('body', 'xs', $theme-body-line-height);
 		max-width: measure($theme-text-measure);
 
 		& > p {
-		 margin-block-start: 0;
+			margin-block-start: 0;
 		}
 	}
 }


bennettcc:
PR should be on hold for review until the NCIDS release is merged: https://github.com/NCIOCPL/cgov-digital-platform/pull/4480

docroot/profiles/custom/cgov_site/themes/custom/ncids_trans/front-end/src/lib/components/cgdp-wide-guide-card/index.scss
```diff
@@ -29,6 +29,7 @@
 	}
 	.nci-guide-card__title {
 		@include typeset("heading", "xl");
+		@include u-line-height('heading', 2);
```
bryanpizzillo, line 32: The line-height should go into `typeset()` as the third option. 

Additionally, I wonder if the token for the line-height should be used (`$theme-heading-line-height`). The use of the token would be for the case in which the line-height of this heading should match the line height of prose. If that were the case, then this should probably use one of the heading tokens too (`$theme-h2-font-size`).

@andyvanavery31 - Is the intent that the font size of the Wide Guide Card should be the same as a visual/semantic H2 in prose? (Whereby if we globally changed the font-size of body content, this should change as well)

bryanpizzillo (changes requested):
This is going to need to be rebased and probably fixed up after the NCIDS update/spacing changes. However, there is a change requested.

docroot/profiles/custom/cgov_site/themes/custom/ncids_trans/front-end/src/lib/components/cgdp-wide-guide-card/index.scss
andyvanavery31: @bryanpizzillo @karlikpj the intent is that the line-height for the WGC title is consistent with other headings of the same size, which is the same as nci-heading-h2 (font size: 24.96px and line-height: 29.952px). Yes it would be good to use a token for the font so that in the future if we made a change to global size it would apply to this.

olitharp-nci (approved):

andyvanavery31 (approved):
Passes Product Review.

bryanpizzillo (approved):

PR #4513, opened 2024-12-04
(#4479) Remove double-escaping of application_page titles.

### LINKED ISSUE #4479: Apostrophe's in application page titles show as encoded
### Issue description
In application pages only (from what I've seen), if you have an apostrophe in the browser title it will show up as encoded. For example:

`Search Results - President&amp;amp;#039;s Cancer Panel`

It appears to be encoding the apostrophe correctly to `&#039;` and then encoding _that_ (twice more?) to `&amp;amp;#039;`

On localhost I'm only seeing two levels of encoding--`&amp;#039;` instead of `&amp;amp;#039;`. But that's possibly because on the PCP search page where this surfaced the apostrophe is not in the browser title itself but in the site title and is getting appended to the page browser title to create the title tag.

> **ESTIMATE** TBD

1.  Fire up local development
2. Create a new Application page
3. In the browser title include an apostrophe (`James's Test Page`)

- Browser title renders apostrophy correctly

### What's the actual result?
- `<title>James&amp;#039;s Test Page - NCI</title>`

- Screenshot 2024-10-29 at 12 39 22 PM
- Screenshot 2024-10-29 at 12 39 39 PM

This also occurs in the page title field, and with the characters `<`, `>`, `"`, and `&`.

In the App Module Instance Settings, `drupalConfig.removeHeadElements` specifies a list of element names (not the actual tag names) to remove from the `<head>` block.  e.g. `title_tag`, `canonical_url`, and so on.

For Sitewide Search, this list is empty, and thus the `<title>` tag is rendered.  For the dictionaries, dynamic listing pages, and manual listing pages, `title_tag` is included in the list and the tag suppressed.

R4R is its own special case and doesn't have the ability to suppress the `title` tag.

blairlearn:
Closes #4479

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_application_page/cgov_application_page.module
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_application_page/cgov_application_page.module
@@ -150,7 +150,7 @@ function cgov_application_page_preprocess_html(&$variables) {
         // of the metatag module's title tag.
         $token_service = \Drupal::token();
         $title = Html::escape(
-          $token_service->replace(
+          $token_service->replacePlain(
             '[cgov_tokens:cgov-title][cgov_tokens:browser-title-site-name-meta]',
             [
               'node' => $node,


bryanpizzillo (approved):
Looks good.

PR #4542, opened 2025-01-09
(#2288) Updated metadata dcterms issue to capture only date

### LINKED ISSUE #2288: Meta data dcterms.issued should just capture date
On every page dcterms.issued is set and shows the date and time like this:
meta name="dcterms.issued" content="12/25/1980 - 07:00"
(Ignore the actual date. Just an example).

A related issue is that on Spanish pages this value is presented in this format:
meta name="dcterms.issued" content="25/12/1980 - 07:00" />

Please update the the formatting of this value to be just the date and consistent between English and Spanish pages, like so:
meta name="dcterms.issued" content="1980-12-25" /> so that the format is YYYY-MM-DD.

IMPACT: This value is then passed into Adobe Analytics as the release date of a piece of content. We utilize this information in reporting on blogs and other timely content, but this is breaking reports given that the current format isn't recognized as a valid date/time. Please update this value to only show the date value.

navsunka:
**@NCIOCPL/qa-team** Please let @welshja know when this is on an ODE.

Closes #2288


blairlearn:
In addition to QA, this will also require review by the analytics team.

blairlearn (approved):

PR #4559, opened 2025-01-27
(#4503) Remove views-element-container from News & Events

### LINKED ISSUE #4503: CR: News & Events Remove views-element-container from the Left Column
### Request Description

This change request is to take the `views-element-container` out of the markup because it should not be in there. This is on the [News & Events page](https://www.cancer.gov/news-events) on the left column of two-column container.

- The `views-element-container` is around the left column of the two column container and it is to be removed.

- The `views-element-container` is around the left column of the two-column container

image

### Related Tickets
* Original Story Ticket #4013

dev-rana-publicis:
Closes #4503


andyvanavery31 (approved):

bryanpizzillo (approved):
Its fine. I do wonder what this might break in the future, but looks ok.

PR #4573, opened 2025-01-30
(#4533) Add severity level to syslog output.

### LINKED ISSUE #4533: Add severity to Drupal syslog output
The Drupal syslog module outputs a whole bunch of stuff except for the severity. During the implementation of Data Dog we noticed that all the `drupal-watchdog.log` entries were missing any severity level, and thus datadog lists them all as informational. This task is to add severity to the syslog settings configuration.

### Technical Approach
I think the best way is to make a `cgov_syslog` module and put the desired config into `config/install`. Then change toggle modules to enable/disable `cgov_syslog` instead of `syslog` in `blt.yml`. This way we can have features keep track of the `syslog.settings` config when the module is enabled. One does not normally enable `syslog` with `watchdog`. (`watchdog` being the database logger.)

blairlearn:
Closes #4533


docroot/profiles/custom/cgov_site/modules/custom/cgov_syslog/cgov_syslog.module
```diff
@@ -0,0 +1,33 @@
...
+      $syslog_config->set($key, $value);
+    }
+    $syslog_config->save();
+  }
+}
+
+/**
+ * Update hook to replace syslog.settings with ours.
+ */
+function cgov_syslog_update_9001(): void {
```
bryanpizzillo, line 31: update hooks go into .install files.

Also why do we need the override config here? Why not set the syslog config from the .install for new sites? and set it in an update hook?

blt/blt.yml
```diff
@@ -36,7 +36,7 @@ modules:
       ]
     uninstall:
       [
-        syslog,
+        cgov_syslog,
```
bryanpizzillo, line 39: you will still need to uninstall both `syslog` and the `cgov_syslog`.

blt/blt.yml
```diff
@@ -146,13 +146,13 @@ modules:
...
         cgov_caching_nocdn
       ]
   01live:
     enable:
       [
-        syslog,
+        cgov_syslog,
```
bryanpizzillo, line 155: Install is fine here because it would trigger the install of all the dependencies of the module, e.g., syslog.

bryanpizzillo (changes requested):
I agree we should have a `cgov_syslog` to hold our settings. I don't know if it needs its own settings.

docroot/profiles/custom/cgov_site/modules/custom/cgov_syslog/cgov_syslog.module
blairlearn: The difficulty I ran into is that putting a copy of syslog.settings into the config/install results in a message because the settings have already been loaded by syslog and our settings (the new format string) doesn't get used.

It's annoying, but I didn't see a good alternative.

docroot/profiles/custom/cgov_site/modules/custom/cgov_syslog/cgov_syslog.install
```diff
@@ -0,0 +1,29 @@
...
+  $syslog_config = \Drupal::configFactory()->getEditable('syslog.settings');
+  $syslog_config->set('format', '!severity|!base_url|!timestamp|!type|!ip|!request_uri|!referer|!uid|!link|!message');
+  $syslog_config->save();
+    \Drupal::logger('cgov_syslog')->info('Override syslog.settings with cgov_syslog.settings');
+  }
+
+/**
+ * Update hook to update syslog.settings to our desired configuration.
+ */
+function cgov_syslog_update_9001(): void {
```
jfrank-nih, line 27: I am not great with Drupal update hooks, but should this be `10001`?

blairlearn: Should be, and will be.

jfrank-nih (approved):

PR #4635, opened 2025-03-18
(#4634) Remove cgov_common from front-end build

### LINKED ISSUE #4634: Remove cgov_common theme building from scripts
> [!WARNING]
> There may be libraries references to cgov_common in the ncids_trans theme. There should not, but also, I can't guarantee anything. This may even appy to the ck-editor css.
> 
> There are a number of Functional tests that are using cgov_common, as well as references to it in modules.

Now that we no longer are using cgov_common and waiting for ticket #4648 to remove the theme, we should disable the cgov_common build from the blt front-end source + build. This will save time from the CI/CD builds, probably a few minutes, but that adds up over time.

bryanpizzillo:
Closes #4634

### DIFF
--- a/scripts/blt/custom/frontend-assets.sh
+++ b/scripts/blt/custom/frontend-assets.sh
@@ -8,8 +8,3 @@ set -e
 pushd ncids_trans/front-end
 npm run build:prod
 popd
-
-## Install cgov_common packages
-pushd cgov/cgov_common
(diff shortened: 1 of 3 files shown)


bryanpizzillo:
Bah, this is not trivial. A number of tests use the cgov_common theme for some reason. Closing.

PR #4672, opened 2025-04-17
(#4614) Update CSS Comments for 16x9 Aspect Ratio

### LINKED ISSUE #4614: Legacy CSS for Inline Video requires updates to remove percent values.
### Issue description
The existing legacy css for the Inline Video component is using percent values for padding. While this works, its not consistent with uswds token values and needs to be updated.

> **ESTIMATE** TBD

1.  View `docroot/profiles/custom/cgov_site/themes/custom/ncids_trans/front-end/src/lib/components/cgdp-inline-video/index.scss`
2.  line number 11 & 26 contains `padding-bottom: 56.25%;`
3. 

-

### What's the actual result?
-

- Screenshot

### Related Tickets
* Issue #9999
* Issue #9999

karlikpj:
Closes #4614

### DIFF
--- a/docroot/profiles/custom/cgov_site/themes/custom/cgov/src/styles/_brp.scss
+++ b/docroot/profiles/custom/cgov_site/themes/custom/cgov/src/styles/_brp.scss
@@ -30,6 +30,7 @@ $gray: #9e9e9e;
     position: relative;
     overflow: hidden;
     height: 0;
+    /* 16:9 Aspect Ratio (divide 9 by 16 = 0.5625) x width 100% */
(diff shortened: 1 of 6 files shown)


andyvanavery31 (approved):

bryanpizzillo:
We should investigate [add-aspect](https://designsystem.cancer.gov/foundations/spacing/height-width#aspect-ratio) which adds in the css [aspect-ratio](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/aspect-ratio) style to constrain the video instead of using magic numbers.

PR #4714, opened 2025-05-06
(#4685) Glossification breaking with italics

### LINKED ISSUE #4685: Glossification Modal Does not Work with Italicized Terms
### Issue description
This issue was raised via Becky Chasan on 04/22 via the WCMS User Help email distribution list. When a term is both Glossified and italicized the Glossify modal does not launch in the WYSIWYG. Additionally, a 403 error is thrown with different styling and the content editor cannot select the checkbox to glossify the term. The issue appears to be in the DOM as an aspect of the JS.

#### Scenario 1: HTML inside the glossification link

1. Navigate to any body section WYSIWYG that contains terms that can be Glossified
2. Click the Glossification icon
3. Select at least one term to Glossify
4. Click "Save"
5. Italicize the word that was Glossified 
6. Click the Glossification icon
7. You will receive an error message

- The glossification modal will launch when there are terms that are both glossified and italicized. The glossification modal will launch when there are extraneous style tags.

### What's the actual result?
- The Glossification modal will not launch when at least one term is both glossified and italicized

### Additional details
* The editor is opinionated about the order of `<a>` and `<em>` tags. If we tried moving it to `<em><a>...</a></em>`, as soon as we left source mode it switched it back to `<a><em>...</em></a>`. We'd have to test to verify, but I wonder if CKEditor 4 was either not opinionated or opinionated the opposite direction. If that was the case, maybe the problem with glossification and HTML has been there for a while but the switch to CKEditor 5 is just exposing it. Attempting to fool it by doing `<em> <a>...</a></em>` did not work.

### Related Tickets
* N/A

erikscottmeyer:
Closes #4685

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/pdq_glossifier/js/ckeditor5_plugins/glossifier/src/glossifyactioncommand.js
+++ b/docroot/profiles/custom/cgov_site/modules/custom/pdq_glossifier/js/ckeditor5_plugins/glossifier/src/glossifyactioncommand.js
@@ -5,6 +5,7 @@
 
 import { Command } from 'ckeditor5/src/core';
 import {
+  convertTagStringToHTML,
   prepareEditorBodyForGlossificationRequest,
   getContentLanguage
 } from './utils/content-preparation';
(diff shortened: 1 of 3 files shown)


alinai:
@lburack , For a term that is italicized in the WYSIWYG, should that term appear in italics in the glossification modal? 
Note, that currently, on dev-ac and ODE, no matter a term is bold or not in the WYSIWYG, it appears as bold in the  glossification modal.

cc: @KateMashkinaNIH , @laurelthrash 

image
image

andyvanavery31:
@alinai with confirmation from @adrianacastaneda , the preferred behavior is that the formatting does not appear in the modal popup (i.e., no italics or strikethroughs), and the terms that can be glossified maintain the display of the term being bold with the checkbox to the right of the term.

andyvanavery31 (approved):
This PR will be approved since it fixes the main problem, and backlog ticket #4716 was made to fix the sporadic showing/not showing of italics styles on the glossification modal popup.

bryanpizzillo (approved):

PR #4729, opened 2025-05-21
(#4143) Run PHPCS against additional extensions.

### LINKED ISSUE #143: Update Drupal Core
We are using a 8.7-x dev from early November, it needs to be updated.

### LINKED ISSUE #4143: PHPCS not scanning PHP files with other extensions
### Issue description
PHPCS  is currently only scanning a subset of PHP files, with the `.php` extension. Other files in Drupal also have PHP and should be scanned.

### Issue solution
This line from the stock `phpcs.xml` file should be added to ours:

````
<!-- Set extensions to scan (taken from Coder 8.3.6). -->
  <!-- @see https://git.drupalcode.org/project/coder/blob/8.3.6/coder_sniffer/Drupal/ruleset.xml#L8 -->
  <arg name="extensions" value="php,module,inc,install,test,profile,theme,css,info,txt,md,yml"/>
````

That will make `.module` and `.install` files get picked up, among other things. File types to be tweaked as needed.

Doing this will also involve having to fix all the PHPCS errors in those files.

### See also
#2920

### PR DESCRIPTION
**Pull request by blairlearn**
Fixes simple items only:
- Whitespace fixes.
- Order of using statements.
- Remove unused variables and usings.
- Comment formatting for line length (some wording changes for space).
- Remove old update hooks
- Remove phpcs exclusion of files that no longer exist/no longer have problems.

Rule violations not fixed in this PR have been excluded in phpcs.xml and added to #4730 

Closes #4143

### CODE REVIEW

**Review by jfrank-nih — APPROVED** [review]

File: phpcs.xml [hunk 1]
```diff
@@ -65,4 +69,23 @@
   <exclude-pattern>docroot/profiles/custom/cgov_site/modules/custom/pdq_core/src/Commands/RevisionPrunerCommands.php</exclude-pattern>
   <exclude-pattern>docroot/profiles/custom/cgov_site/modules/custom/pdq_glossifier/js/plugins/cgov_glossifier/dialogs/glossify.js</exclude-pattern>
   <exclude-pattern>docroot/profiles/custom/cgov_site/modules/custom/pdq_glossifier/js/plugins/cgov_glossifier/dialogs/glossify.js</exclude-pattern>
+  <exclude-pattern>docroot/modules/custom/app_module/app_module.module</exclude-pattern>
```
bryanpizzillo, line 72: I take it these were all a bridge too far? Can you comment on the PR the issue number where these are captured to fix?

bryanpizzillo (changes requested):
So whether it is in a separate commit or not, we should get rid of the old update hooks that we are reformatting in this PR given we can't actually test them. The key thing is to either comment what the last hook function was or keep the last update hook funtion and empty out the old update hook body. We do need to know this to make new update hooks.

This clean up will probably also go a long way toward PHPStan issues in any update hooks. I do think it might raise a question about pdq update hooks and extra functions that only those hooks are calling.

phpcs.xml
bryanpizzillo: One question, why are there 2 .install files here? Are just the update hooks the offenders?

blairlearn:
It looks like app_module.module is the only place I missed that.  😅

blairlearn:
Replaced by #4750

PR #4739, opened 2025-05-29
(#4738) Write JS Unit Tests for Glossifier Plugin

### LINKED ISSUE #4738: Enabler: Create JS Unit Tests for Glossifier
In working through #3006 it became very evident that we need unit tests for as much of the CKEditor plugin as we can get. (Writing CKEditor specific test can be a future task, but there are many things like fetching glossaries that can be tested)

This ticket is to put Jest in place and get "some" tests in. (8 suites with 76 tests according to the current count. 😸 ) Some refactoring will probably be done to separate the selection dialog from the manipulation function.

bryanpizzillo:
- Added Jest (and friends) to pdq_glossifier package.json
- Refactored utils and elements of glossifyactioncommand into utils for better testing
- Added tests for all items under utils
- Update blt front-end tests to test glossifier

Closes #4738

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/pdq_glossifier/.gitignore
+++ b/docroot/profiles/custom/cgov_site/modules/custom/pdq_glossifier/.gitignore
@@ -1,1 +1,2 @@
 js/build
+coverage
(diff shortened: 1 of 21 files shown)


bryanpizzillo:
@KateMashkinaNIH  this is not exactly for merging, but it would be great to get some light testing of the glossification feature. I did some refactoring and a bunch of tests.

PR #4818, opened 2025-07-15
Ticket/4762 phpstan pdq

blairlearn:


docroot/profiles/custom/cgov_site/modules/custom/pdq_drug_information_summary/src/Plugin/rest/resource/PDQResource.php
```diff
@@ -189,6 +189,7 @@ public function post(array $data) {
       /** @var \Drupal\Core\Entity\ContentEntityStorageBase $storage */
```
bryanpizzillo, line 189: See if you set this type to `\Drupal\node\blah\NodeStorage` if it does not require you to add:
```
/** @var \Drupal\node\Entity\Node */
      $node = $storage->loadRevision($vid);
    }
```

bryanpizzillo: You should see if you even need to do the cast for `/** @var \Drupal\Core\Entity\ContentEntityStorageBase $storage */`

bryanpizzillo: This probably won't work because phpstan-drupal does not define what the loadRevision* functions can return. But try it anyway -- however, you should really fix the storage type since we know what NodeStorage is.

bryanpizzillo (changes requested):
This is fine, but I would like you to look into the type guessing as mentioned below.

PR #4828, opened 2025-07-21
(#4535) Update Node 22

### LINKED ISSUE #4535: Upgrade front-end build tools to Node 20.x
We are using Node 16 which is old. I am sure that Node 20 will cause its own set of heartache. (Jest is probably going to be an issue. Maybe Typescript.)

While working on the Glossifier tests I was unable to do more modern testing because of our Node version. (Could not get MSW (for network mocking) working because I could use neither the Jest test environment workarounds for MSW, nor Vitest) (And on not having a test framework with a real browser I had a number of annoying headaches getting browser API things running just right.)

bryanpizzillo:
Local cypress did not run without needing effort. So trying CI/CD.
Should have made this draft because I want to do some npm audit fixes.

Closes #4535


bryanpizzillo:
Gonna do a big PR with #4830

PR #4875, opened 2025-08-04
(#4861) Removes cgov common entrypoints and references

### LINKED ISSUE #4861: Remove cgov_common entrypoints
We need the script and styles within cgov/cgov_common under themes because it is referenced in ncids_trans. We don't need the cgov_common entrypoints anymore though. These were the JS/CSS that were built by webpack and referenced in the cgov_common.libraries.yml file. We ran into an issue when removing the old glossification in that we had to edit the common entrypoint so that we could remove the references to the files we were removing. We should test these changes, but nothing uses them, so it would just be best to remove them altogether and not have it be an issue anymore.

So this is to remove `docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/src/entrypoints` and everything under it. We should also remove `docroot/profiles/custom/cgov_site/themes/custom/cgov/cgov_common/cgov_common.libraries.yml` while we are at it. No sense in referencing things that we are no longer building.

dlescarbeau:
Closes #4861


bryanpizzillo (changes requested):
Please revert the delete of docroot/translations/drupal-10.4.6.es.po, otherwise everything else looks good. (well, looks good if it passes testing :D)

alinai:
Regression tests (FE and CMS) and Visual regression passed. Glossification looks good on ODE988.

andyvanavery31 (approved):

PR #4882, opened 2025-08-05
(#4879) Update Event Yaml with different dates and time

### LINKED ISSUE #4879: Update YAML Events to Have Different Dates & Time
### Issue Description
Automated testing is not acting as expected because more than one event has the same date and time, and the displayed order of those events in a Dynamic List of Events varies due to the upload order of the events when creating a new test environment. This CR is to update the latest events so this no longer happens.

### Technical Notes
- This is a hot fix to take care of the overarching issue which will be handled by #4880 

- Test Content Events will have different dates and times

- The last few updates to the platform introduced YAML content for Events that occur at the exact same date and time. Since the display order of Events on a Dynamic List have the fallback rule of upload time, the order of the events on said list vary and break automated testing.

dev-rana-publicis:
Closes #4879

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_yaml_content/content/952_embed_infographic_test_pages.content.yml
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_yaml_content/content/952_embed_infographic_test_pages.content.yml
@@ -1172,9 +1172,9 @@
   field_page_description:
     value: 'Event Page - Page Description'
   field_event_start_date:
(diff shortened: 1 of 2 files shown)


andyvanavery31:
Passes my review. Will add the product checkmark once it passes QA.

bryanpizzillo (commented):
This looks good to me. I will click the approve button after testing.

KateMashkinaNIH:
LGTM - @bryanpizzillo hit that approve button

andyvanavery31 (approved):

PR #4947, opened 2025-09-09
(#4946) Update Deployability Test to Use Base

### LINKED ISSUE #4946: Spike: Revisit CI/CD test-deployability workflow
We have run into an interesting scenario. We are trying to remove a list option from a dropdown in Drupal. Drupal will not let it be removed unless not fields are using it. There is no content in production that is using this option. HOWEVER, the YAML in the master branch DOES use it. So we have no way of resolving this issue unless we push the YAML to master, which, since we do multiple releases could be adding features used by that same YAML.

So the question here is, can we install the PR target branch and then ensure the update hooks in the PR work. This way we could remove the offending items from the YAML in a separate commit merged to the target (most likely develop) and then ensure the PR can be applied. 

Any sort of weird "because another update hook ran in the commits before this in a release branch that now conflict with an update hook in the PR, which would not have been noticed on its own" will just need to be sorted out in the release branch testing.

There are still some quirks with this process in `automation0refreshed` in ACSF, but we can always stage to ACSF first, then clean up offending content, then deploy the commit.

P.S. Any changes to the target branch since the last run should force a re-run. I.E. if the target is develop and you rebase the PR because develop was updated, then the rebase should rerun the test-deployability workflow.

bryanpizzillo:
Closes #4946

Workflow ( this PR ) run that used develop as the base: https://github.com/NCIOCPL/cgov-digital-platform/actions/runs/17590750317/job/49970674538#step:3:472
Workflow ( [prototype/test-4946-check-master](https://github.com/NCIOCPL/cgov-digital-platform/tree/refs/heads/prototype/test-4946-check-master) ) run that used master as the base: https://github.com/NCIOCPL/cgov-digital-platform/actions/runs/17590285123/job/49969154535#step:3:472

### DIFF
--- a/.github/workflows/test-deployability.yml
+++ b/.github/workflows/test-deployability.yml
@@ -3,15 +3,11 @@ name: Test Ability to Deploy
 on:
   push:
     branches:
-      - master
-      - main
       - develop
       - 'hotfix/**'
       - 'release/**'
       - 'feature/**'
       - 'prototype/**'
-    tags:
-      - '*'
     paths:
       # IMPORTANT: Changes to these file specs must, Must, MUST  be reflected in the file specs
       # for pull_request and the check_for_past_run job.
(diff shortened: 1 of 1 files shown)


jfrank-nih (changes requested):
Unfinished comment.

PR #4969, opened 2025-09-19
(#4966, #4967) Fix error on CI environments from purge queue emptying

### LINKED ISSUE #4966: Test Ability to Deploy should fire on BLT changes
If the BLT yaml is changed then we could be adding/removing modules on a given environment, including CI. Those changes could affect the ability of the workflow to complete successfuly. So we need to be testing the ability to deploy on changes to BLT configuration as well.

### LINKED ISSUE #4967: Install dummy purger in CI
### Issue description
The CI workflow for Test Ability to Deploy now needs the dummy purger installed after #4758 was completed. That's because #3331 now runs a `p-queue-empty` and gets this error:

```
  [Symfony\Component\Console\Exception\CommandNotFoundException]  
  Command "p-queue-empty" is not defined.      
```

> **ESTIMATE** TBD

Run the Test Ability to Deploy workflow.

- Success

### What's the actual result?
- Failure

jfrank-nih:
Closes #4966
Closes #4967

### DIFF
--- a/.github/workflows/test-deployability.yml
+++ b/.github/workflows/test-deployability.yml
@@ -14,6 +14,7 @@ on:
       - '**/config/install/*.yml'
       - '**/modules/custom/*/*.install'
       - '.github/workflows/test-deployability.yml'
+      - 'blt/*'
   pull_request:
     types: [opened, reopened, synchronize, ready_for_review]
     paths:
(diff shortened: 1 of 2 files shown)


bryanpizzillo:
So this is going to be sketchy, but, Develop is in a broken state which is why Test Ability to Deploy is failing. Technically we don't know if CI will not install until we try after this is merged.

PR #4978, opened 2025-09-23
(#4848) Disallowed Attributes Migration Script

### LINKED ISSUE #4848: Create migration script for articles with embedded media

dev-rana-publicis:
Closes #4848

### DIFF
--- a/docroot/modules/custom/ncids_html_transformer/tests/src/Unit/BasicMigrationTransformerTest.php
+++ b/docroot/modules/custom/ncids_html_transformer/tests/src/Unit/BasicMigrationTransformerTest.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace Drupal\Tests\ncids_html_transformer\tests\src\Unit;
+namespace Drupal\Tests\ncids_html_transformer\Unit;
 
 use Drupal\ncids_html_transformer\Services\NcidsDisallowedClassesTransformer;
 use Drupal\ncids_html_transformer\Services\NcidsDisallowedStylesTransformer;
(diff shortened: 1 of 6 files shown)


bryanpizzillo:
@dev-rana-publicis you need to account for the Drupal anchor element attributes for managed links. (`'data-entity-uuid', 'data-entity-substitution'`) There may be others, check the linkit module / look at some links in (real) WYSIWYG content.

(Obviously with your recent changes, you won't be touching `<drupal-embed>` elements.)

laurelthrash:
Skipped over ODE testing. Our ODE pages do not have bad attributes since we made the YAML for these pages ourselves. Thus, this script is not applicable. Will only test this script on ACSF. Cc: @andyvanavery31

docroot/modules/custom/ncids_html_transformer/tests/src/Unit/BasicMigrationTransformerTest.php
```diff
@@ -2,6 +2,7 @@
 
 namespace Drupal\Tests\ncids_html_transformer\tests\src\Unit;
```
bryanpizzillo: This path is actually wrong. It should be `Drupal\Tests\ncids_html_transformer\Unit`. (I learned this looking at something else, and that else failing.)

But also, you should have tests for the actual transformers you are making -- you should not keep editing this file or it will become a bottle necks.

andyvanavery31:
Passed on Final ODE testing.

KateMashkinaNIH (approved):

bryanpizzillo (approved):

PR #4979, opened 2025-09-23
(#4975) Add acquia/acsf-tools

### LINKED ISSUE #4975: Aquia MEO: Install acsf-tools
Acquia PS team requires that their [acquia/acsf-tools](https://github.com/acquia/acsf-tools) utility be available in our codebase in order to upgrade to MEO.

It looks like the acsf-tools version needs to 11.x: "If you upgraded Drush to 11.x then use acsf-tools 11.x". They don't appear to provide a 12.x version. 

We do need to specifically install the `*.2.x` version as that version has been modified by Acquia to support both ACSF and MEO.

blairlearn:
Closes #4975

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,9 +21,11 @@ docroot/modules/contrib
 docroot/themes/contrib
 docroot/profiles/contrib
 docroot/libraries
+
+# Keep *our* drush commands, but ignore those installed by composer.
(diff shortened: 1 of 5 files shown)


jfrank-nih (approved):
Just need to run ACSF regressions

PR #4988, opened 2025-10-07
Ticket/4591 inner page body section

dev-rana-publicis:


bryanpizzillo (changes requested):
A few items:
1. The options for the WYSIWYG should not be copied from the NCIDS Streamlined. 
    * They should be:
       * Replaces global and entity tokens with their values - this allows us to reference drupal tokens from within content.
       * Correct faulty and chopped off HTML - this is important because you do not want to limit the HTML, just clean up bad HTML
       * Linkit URL converter - this is for managed links
    * The current full additionally has the following, and the reasoning about why we might not need them.
       * Align images - I don't think this is needed because we don't allow non-media images to be added
       * Display embedded entities - this should not be there yet since we have no embeds configured.
       * Track images uploaded via a Text Editor - again I think this is for non-media images being added via the WYSIWYG
2. Make sure the filter order is similar.
3. You also need to check the "Clean Styles" checkbox in the "NCI CKEditor5 Cleanup" options at the bottom of the format/editor config page. This is what cleans up the styles, if it is not checked, the wild west will start again.

I am still going through the front-end bits, but wanted to add these.

docroot/profiles/custom/cgov_site/themes/custom/ncids_trans/templates/content/article/paragraph--body-section.html.twig
```diff
@@ -9,7 +9,8 @@
  */
 #}
 {# Setup attribs if schema org is enabled #}
-{% set section_attribs = create_attribute() %}
+{% set section_attribs = create_attribute().addClass('usa-prose') %}
+
```
bryanpizzillo, line 13: `usa-prose` should not wrap the h2. You should just be SMARTLY wrapping:
```
  {% if schema_org_type %}
    <div itemscope itemprop="acceptedAnswer" itemtype="https://schema.org/Answer">
      <div itemprop="text">
  {% endif %}
  {{ content.field_body_section_content }}
  {% if schema_org_type %}
      </div>
    </div>
  {% endif %}
```

Smartly means that if there is `<div itemprop="text">` then just add a class to that div. If it is not then you need to add a div when it is not schema org. 

THEN you need to add `nci-heading-h2` to the h2. Or, make it `cgdp-article-body__heading` and extend `nci-heading-h2`.  Then you can also control the `<p>` tag spacing in `cgdp-article-body__heading` rather than do it in a hacky.

docroot/profiles/custom/cgov_site/themes/custom/ncids_trans/templates/content/article/node--cgov-article--full.html.twig
```diff
@@ -1,4 +1,5 @@
 {{ attach_library('ncids_trans/article') }}
+{{ attach_library('ncids_trans/ncids-minimal') }}
```
bryanpizzillo, line 2: This should not be added here... that was a bit of misspeak on my part and confusion on yours. The `ncids-minimal` and `ncids-trans-common` libraries are referenced from **page** templates, not the **node** templates. So when I said you need to change it, I meant you need to figure out how to deal with the page template. Because right now you still are getting all 3...
Image

We can go a couple of ways, but maybe the following is the best:
1. Copy `docroot/profiles/custom/cgov_site/themes/custom/ncids_trans/templates/layout/page--ncids-inner-page.html.twig` to `docroot/profiles/custom/cgov_site/themes/custom/ncids_trans/templates/layout/page--ncids-inner-page--migrated.html.twig`
2. Change `page--ncids-inner-page--migrated.html.twig` to reference `{{ attach_library('ncids_trans/ncids-minimal') }}` 
3. Then add a theme suggestion for `--migrated` if the bundle is in a list of "converted" types. For now this list only should have `cgov_article`

bryanpizzillo (changes requested):
Ok, here is some stuff in the front-end part to work on. Oh, with the page template, I don't **think** there are any changes to the HTML because it should be pure NCIDS.

bryanpizzillo (changes requested):
From this mornings conversation:
1. Delete `05500_coping.content.yml`
2. Fix the buttons in fact book content to match prod. (Although tables and charts will not be right)
3. We need to look into styling the table in the WYSIYWG. Although, I can definitely live with pulling this out into its own ticket to get this in.

PR #5011, opened 2025-10-28
(#4911) WYSIWYG Link Analytics

### LINKED ISSUE #4911: Analytics | Add Analytics Click Event to Body Link Clicks on NCIDS WYSIWYG
### Request Description
We'd like to add an analytics click event when users click on a body link within the Intro Text and/or Body Sections on inner pages using NCIDS WYSIWYG's (Full and Streamlined without Headings as applicable). 

### Technical Notes
- There are currently click events for components on MLPs that use the streamlined NCIDS WYSIWYG, but on Article Pages no click event fires on body links. 
- Glossified links have their own new analytics for when the pop-up loads, but we'd like this click event to start firing to show when a user has clicked on the link and track in the same way as other links. 
- Embedded items within these pages also fire click events and this new click event should not apply to them.

### Analytics Requirements:
On click of a body link on inner pages using an NCIDS WYSIWYG, an EDDL event shall be raised with the linkName and event of: Inner:WYSIWYG:LinkClick

This click event shall send the following information as data elements to EDDL:
- location: This should be set to 'Body'.
- linkType: This should return “Internal”, “External”, "Media", "Glossified", "Email", or "Other"
  - Follow the same rules as defined on #4949 for the linkType values 
- linkText: This should pass the text of the link clicked. This value should be truncated after 50 characters.

Which page types should this be on?
- Article

This shall **not** include links for:
- Embedded Content
- Feature cards
- Content Blocks
- Infographics
- Videos
- Cases where the WYSIWYG is used by a component that has its own tracking

---------------------------------

This functionality will will be eventually available on the following content types, once they have been migrated to NCIDS and use the NCIDS WYSIWYG's (Streamlined without Headings as applicable and Full).
- Press Releases
- Biographies
- Events
- Cancer Centers
- Blog Posts
  - For Blog Posts - there is an existing click event in the legacy analytics BlogBodyLinkClick that should be removed when this ticket is completed.

karlikpj:
Closes #4911

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_yaml_content/content/13910_wysiwyg_analytics_test.content.yml
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_yaml_content/content/13910_wysiwyg_analytics_test.content.yml
@@ -1,0 +1,190 @@
+
+## Create Test Content to be Used in Lists
+- entity: "node"
+  type: "cgov_article"
+  langcode: en
+  status: 1
+  moderation_state:
+    value: 'published'
+  title: "Test WYSIWYG Anayltics Links"
+  title__ES:
+      value: "Test WYSIWYG Anayltics Links - ES"
+  field_page_description:
+    value: "Test page for WYSIWYG Anayltics Links"
+  field_page_description__ES:
+    value: "Test page for WYSIWYG Anayltics Links - ES"
+  field_browser_title:
+    value: "Test WYSIWYG Anayltics Links"
+  field_browser_title_ES:
+    value: "Test WYSIWYG Anayltics Links - ES"
+  field_site_section:
+    - '#process':
+        callback: 'reference'
+        args:
+          - 'taxonomy_term'
+          - vid: 'cgov_site_sections'
+            computed_path: '/test'
+  field_pretty_url:
+    value: "test-wysiwyg-analytics"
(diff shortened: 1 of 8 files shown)


bryanpizzillo (changes requested):
The JS for the analytics was added to the editor's css/js. Only article uses this WYSIWYG. Also that editor is not used for the intro text on article, and no ncids editor is used for intro text for blogs. So at the very least this analytics.ts should move to wysiwyg/common since it is for more than editor config. Or, it should move somewhere else altogether. 

@andyvanavery31 - what is supposed to be tracking analytics? Legacy content that we agreed to not touch, or just article?

dlescarbeau:
> The JS for the analytics was added to the editor's css/js. Only article uses this WYSIWYG. Also that editor is not used for the intro text on article, and no ncids editor is used for intro text for blogs. So at the very least this analytics.ts should move to wysiwyg/common since it is for more than editor config. Or, it should move somewhere else altogether.

Yeeaahhhh I wasn't sure where this should go, honestly, and figured the WYSIWYG folder might be the best bet (I was the one that suggested its current location). Are you thinking it should just be its own component and included where appropriate?

davisaw:
Analytics are looking good on the test page https://ncigovcdode1037.prod.acquia-sites.com/test/test-wysiwyg-analytics

Issue #4954, opened 2025-09-11
Remove nano symlink from docroot


jfrank-nih:
With the retirement of the Nano site, we need to remove the `docroot/nano` symlink so that it no longer exists as a path that can be accessed on the lower tiers.

PR #5027, opened 2025-11-05
Applied nci_meo_code_changes_nov04.patch for MEO switch

bryanpizzillo:
TO-DO: Get rid of nano! (https://github.com/NCIOCPL/cgov-digital-platform/issues/4954)


docroot/profiles/custom/cgov_site/modules/custom/cgov_core/src/CgovCoreTools.php
```diff
@@ -568,7 +567,7 @@ public function getCloudEnvironment() {
 
     // Check if site_env matches a prod environment name...
     if (in_array($site_env, self::PROD_AH_SITE_ENVS) ||
-      preg_match('/^.*(www-prod-acsf|www-cms).*$/', $site_env)) {
```
bryanpizzillo: Hey @jfrank-nih -- I don't think this is something that should have been changed... But also, why is only this here?

docroot/sites/default/settings.php
```diff
@@ -1,19 +1,22 @@
...
+    DRUPAL_ROOT . '/sites/settings/post_new_relic.php'
+  ];
+  foreach ($settingsFiles as $settingsFile) {
+    if (file_exists($settingsFile)) {
+      // phpcs:ignore
+      require $settingsFile;
+    }
+  }
   return;
 }
```
bryanpizzillo: 2 things:
1. The folder location means they are getting loaded in [global.settings.php](https://github.com/NCIOCPL/cgov-digital-platform/blob/1c7a3e3b34c284e5746f43081de5a3f03df7c5df/docroot/sites/settings/global.settings.php)
   * These probably come first
   * We need to plot the order again of all the includes. 
2. Return? Really? So are we skipping our config settings?

docroot/sites/default/settings.php
```diff
@@ -1,19 +1,22 @@
...
-// ===== Added by acsf-init, please do not delete. Section end. =====
-}
+
+// @codingStandardsIgnoreFile
```
bryanpizzillo: This is also a little silly.

docroot/sites/settings/cgov_caching.settings.php
```diff
@@ -23,22 +23,31 @@
...
-    case '01test':
-    case 'test':
-    case '01live':
-    case 'ode':
-      // Cache settings.
-      $config['system.performance']['cache']['page']['max_age'] = 16588800;
-      break;
+  // MEO environments.
+  if ($_ENV['AH_ENVIRONMENT_TYPE'] === 'meo') {
+    switch ($env) {
```
bryanpizzillo: This is messy. We should re-work this so it is just 1 switch. Maybe just fake the meo environment names...

But also, what in here is NOT going to get the max_age? Maybe we negate this whole thing. It looks like this is just for the everything but the local env.

(NOW, it USED to have different settings, but I think changes there make this no longer valid.)

bryanpizzillo:
Based on discussions with Acquia we no longer need:

## Settings
* post_memcache.php - only needed by non cloud-next environments and was a port of the ACSF only settings.
* post_new_relic.php - We should not be using it anymore.
* pre_includes.php - DRS is called in the settings.php, so no need to call it again.

We also cleaned up the settings.php file to make it simpler

## Hooks / Factory Hooks
* 000_meo_db_scrub - another port from acsf that was acsf specific.
*

PR #5034, opened 2025-11-11
(#5032) Enable ClamAV for MEO

### LINKED ISSUE #5032: Enabler: Enable virus scanning for uploads (MEO style)
**Context:** We tried doing this once in #3398 and it wasn't feasibly in ACSF. We undid it and virus scanned using automations. With the move to MEO we would have to rework those automations. Since Acquia Cloud Next (in theory) supports ClamAV in daemon mode we should try this again.

We need to enable virus scanning for uploads using the ClamAV module. Acquia provides ClamAV on the Acquia infrastructure. 

We may also want to investigate a cron job to scan files on the server at regular intervals. Or this may only be needed as a one-time task when first enabling ClamAV.

### Tasks

**Development**

- [x] Update dockerfile for local development to [install ClamAV](https://www.drupal.org/project/clamav)
- [x] Install and enable the [ClamAV Drupal module](https://duckduckgo.com/?q=clamav+drupal+module&t=brave&ia=web)
- [x] [Configure ClamAV](https://docs.acquia.com/cloud-platform/manage/antivirus/) module to point to executable

**Ops**

- [ ] Identify whether `clamscan` is present on CI/CD pipelines and if an automated test to scan a file can be created
    - If so, do we add the test file to our code repo? (GitHub might not appreciate it.) 
    - Do we download the test file from a known location?
    - Forego the test entirely and just do the manual test?
    - It looks like the eicar test file is just a text string. We could include this in a test but broken up into pieces, which would avoid making any scanners cranky. It could then be reassembled into a file by the test for scanning.
- [ ] Work with QA team to get a manual test with the [EICAR test virus](https://en.wikipedia.org/wiki/EICAR_test_file) added to the regression tests ([download links](https://www.eicar.org/download-anti-malware-testfile/))
- [ ] Initiate a one-time scan on the PROD servers of all previously uploaded material using `clamscan -ri`

> **ESTIMATE** TBD

### Resources:
* Instructions: https://docs.acquia.com/cloud-platform/manage/antivirus/
* Background: https://docs.acquia.com/cloud-platform/arch/security/#cloud-antivirus-uploads
* Module: https://www.drupal.org/project/clamav

### Prerequisites
- None

### Sub-Tasks
- Test using test virus

### Testing Steps
- Log in to CMS and attempt to upload a file (image or otherwise) into the system.
    - Expected:  Upload should complete and you will be able to save the file.
- Extra test: Log in to CMS and attempt to upload the EICAR test virus.
    - Expected: Upload should get rejected.

jfrank-nih:
Enables the ClamAV module (again) but limits it to only installing on MEO sites.

Closes #5032


bryanpizzillo (changes requested):
The best way to handle supporting ClamAV on some environments and not others is to make a new module.
1. Create a `cgov_antivirus` module.
2. Give the new module a dependency on clamav
3. Stick the `clamav.settings.yml` in that module's `config/install`
4. Update blt.yml and add the module to the `modules.<env>.enable`  lists for the meo environments.
5. Update blt.yml and add the module & the clamav module to the `modules.<env>.uninstall`  lists for the non-meo environments.

There is no need for anything in cgov_core.

ChadChandlerPS:
@bryanpizzillo

Agreed, that's a cleaner approach. Just to confirm ...the plan is to:

1. Create a new `cgov_antivirus` module with `clamav`as a dependency in `info.yml `and move `clamav.settings.yml `into its config/install/
3. Update `blt.yml `to add `cgov_antivirus `to the enable lists for **meodev**, **meostage**, and **meoprod,** and add `cgov_antivirus `and `clamav `to the uninstall lists for all non-MEO envs.
5. Finally, remove everything ClamAV-related from cgov_core ....the dependency, the config file, and the update hook

ChadChandlerPS:
We can close this PR as this other PR [Enable ClamAV for MEO ](https://github.com/NCIOCPL/cgov-digital-platform/pull/5196) is exactly what we need and will be deployed to ODE and ACSF for testing.

jfrank-nih:
Replaced by #5196

PR #5083, opened 2025-12-16
(#5080) Nested External Link Bug

### LINKED ISSUE #5080: Bug: NCIDS External Link Icons Do Not Reach Multiple Children Levels in HTML
### Issue description
With the update of Article Pages to NCIDS, we are updating all tables to have the scrollable feature. As they get this wrapper, the content within the tables is now more than one child away from the rules created in NCIDS to apply the External Link Icon to any links that do not follow our external link rules. The digital platform, via the NCIDS Full WYSIWYG, needs to be updated with a new version of NCIDS that fixes this so the external link icon appears on content that exists within multiple levels of the hierarchy, such as in the new scrollable tables.

1.  Go to a page with a new scrollable table made with the NCIDS Full WYSIWYG
2. Find an external link within the table
3. See that there is no external link icon for the external link

- Links must follow our rules for external link icons appearing

### What's the actual result?
- If the link is nested within multiple children, such as a link in a scrollable table added through the NCIDS Full WYSIWYG, and the link meets the rule for displaying the external link icon, the external link icon does not display

karlikpj:
Closes #5080

For testing: Please create an Article page with a table that contains external links.

### DIFF
--- a/docroot/profiles/custom/cgov_site/themes/custom/ncids_trans/front-end/src/lib/components/wysiwyg/editors/ncids-full-html/index.scss
+++ b/docroot/profiles/custom/cgov_site/themes/custom/ncids_trans/front-end/src/lib/components/wysiwyg/editors/ncids-full-html/index.scss
@@ -5,7 +5,7 @@
 @forward '../../common/cgdp-embed-summary-box';
 @forward '../../common/cgdp-embed-video';
 @forward '../../common/cgdp-wysiwyg-table';
-@forward "../../common/cgdp-embed-pullquote";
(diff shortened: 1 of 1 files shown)


bryanpizzillo (changes requested):
Also @karlikpj  can you add this to the ncids_full_wysiwyg class, and constrained to usa-prose in that class...?

blilianyu (approved):
Created an article page and see the external link icon in tables on all breakpoints. 

Passes for design. cc: @andyvanavery31 @monika-jaeger

alinai (approved):

PR #5152, opened 2026-03-04
(#4835) Remove axios and replace with fetch

### LINKED ISSUE #4835: Bump Axios to at least 1.10.0
> [!IMPORTANT]
> We should really look into swapping out to use the Fetch API instead of Axios as most browsers now support AbortSignal for handling timeouts. (Which is basically the main reason for using Axios.) This will allow us to avoid chasing the always updating Axios AND we don't have to upgrade to Jest until we want to.

### Issue description

This is a hold over from #4829 - as it was not a super quick fix.

There are a few audit issues that we need to bump axios to fix. One of which needs us to get >=1.10.0. I tried it, and this is not going to be a quick thing.

This is the result of `npm test` after bumping the module. This, IIRC, is because we need to setup babel stuff to get it to understand how to do `import`. I forget if it was just a matter of getting a bable preset, or what. That being said, we also need to bump just for another issue and maybe jest-ts got better in Jest 30?
```
    Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

    Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.

    By default "node_modules" folder is ignored by transformers.

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
     • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/configuration
    For information about custom transformations, see:
    https://jestjs.io/docs/code-transformation

    Details:

    /var/www/docroot/profiles/custom/cgov_site/themes/custom/ncids_trans/front-end/node_modules/axios/index.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,jest){import axios from './lib/axios.js';
                                                                                      ^^^^^^

    SyntaxError: Cannot use import statement outside a module
```

dlescarbeau:
- Removes axios
- Updates tests with mock fetch requests

Closes #4835


laurelthrash (approved):
Passes my review but adding a note about my findings. I reviewed the front end Fact Book high charts. There were very minor discrepancies between the charts on the ODE compared to production, however, these did not impact the readability of the charts. Additionally, the discrepancies were not easily reproduceable.

PR #5176, opened 2026-03-18
(#5174) Update autologout to v1.7.

blairlearn:
- Unregisters js_cookie in cgov_saml_auth_config because autologout forgot to.

This update impacts the automatic logout after two hours.  Note that this limit is only in effect for SSO logins.

Closes #5174

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,19 +17,6 @@
         "drupal": {
             "type": "composer",
             "url": "https://packages.drupal.org/8"
-        },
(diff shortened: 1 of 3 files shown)


bryanpizzillo (changes requested):
3 things... well 2 things and a question.

KateMashkinaNIH (approved):

bryanpizzillo (approved):

PR #5194, opened 2026-04-01
(#4557) Add publishing perms to site_admin role

### LINKED ISSUE #4557: Add publishing permissions to site_admin role
In the event we need to emergency turn off publishing in Drupal, one has to remove publishing rights from content_editor and advanced_editor, and grant those permissions to the site_admin role. It would make more sense that in this case, if the site_admin already had the permissions, then we only need to toggle the permissions on the two lower rolls. This would decrease the needed commands to turn off & turn on.

ChadChandlerPS:
Closes #4557


KateMashkinaNIH (approved):

adrianacastaneda (approved):
Looks good!

dev-rana-publicis (approved):

PR #5196, opened 2026-04-06
Enable ClamAV for MEO

### LINKED ISSUE #5032: Enabler: Enable virus scanning for uploads (MEO style)

ChadChandlerPS:
Enables ClamAV for MEO Sites only

Closes #5032 

Note: Can likely close [older PR](https://github.com/NCIOCPL/cgov-digital-platform/pull/5034)

### DIFF
--- a/blt/blt.yml
+++ b/blt/blt.yml
@@ -48,6 +48,8 @@ modules:
         cgov_caching_cdn,
         acquia_purge,
         akamai,
+        cgov_antivirus,
+        clamav,
       ]
   ci:
     enable:
@@ -59,7 +61,9 @@ modules:
         acquia_connector,
         shield,
         cgov_https_config,
-        akamai
+        akamai,
+        cgov_antivirus,
+        clamav,
       ]
   ode:
     enable:
@@ -75,7 +79,9 @@ modules:
         cgov_syslog,
         syslog,
         cgov_caching_cdn,
-        akamai
+        akamai,
+        cgov_antivirus,
+        clamav,
       ]
   dev:
     enable:
@@ -91,7 +97,9 @@ modules:
         cgov_syslog,
         syslog,
         cgov_caching_cdn,
-        akamai
+        akamai,
+        cgov_antivirus,
+        clamav,
       ]
   int:
     enable:
@@ -106,7 +114,9 @@ modules:
       [
         cgov_syslog,
         syslog,
-        cgov_caching_nocdn
+        cgov_caching_nocdn,
+        cgov_antivirus,
+        clamav,
       ]
   test:
     enable:
@@ -123,7 +133,9 @@ modules:
         syslog,
         devel,
         cgov_caching_cdn,
-        akamai
+        akamai,
+        cgov_antivirus,
+        clamav,
       ]
   ## We have no prod in ACE.
   meodev:
@@ -136,6 +148,7 @@ modules:
         cgov_https_config,
         cgov_caching_cdn,
         akamai,
+        cgov_antivirus,
       ]
     uninstall:
       [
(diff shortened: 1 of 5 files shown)


docker/web/Dockerfile
```diff
@@ -68,6 +68,7 @@ RUN DEBIAN_FRONTEND=noninteractive \
   php8.3-zip \
   mysql-client-8.0 \
   git \
+  clamav \
```
bryanpizzillo: Please revert this. We should not be using this anymore, and we also are not running locally.

ChadChandlerPS: With `clamav `no longer installed in the image, `freshclam` will fail on Docker build since there's nothing to update. The two lines go hand in hand so removing `clamav` without removing `freshclam` would break the build. 

Need to also remove"
`Line 83 — RUN freshclam`

Right? @bryanpizzillo

ChadChandlerPS: Removed line 71 `clamav` 
Removed line 83 - `RUN fresh clam`
from `Dockerfile`

Removing `clamav `from the package list and removing `RUN freshclam` is precisely what a full file revert would have done. We just got there by removing the lines directly rather than using git revert, but the end result is the same.

ChadChandlerPS:
Looks like BLT is doing it's job - module **not** enabled on ODE but **_is_** enabled on `meostage`

bennettcc (approved):

bennettcc:
Approving on behalf of QA and product - passed testing on ACSF

bryanpizzillo (commented):
This looks good to me from a Drupal standpoint. @jfrank-nih (or @blairl-nih) you can approve an merge when it gets to the time you are ready.

bryanpizzillo (approved):

PR #5197, opened 2026-04-07
(#5160) Remove Related Resources from Blog Series

### LINKED ISSUE #5160: [R4.3] CR: Remove Related Resources and article-footer from Blog Series
### Request Description
Currently, Blog Series is trying to render Related Resources on its twig, but it is not a field that can be filled out on the CMS. Therefore, we need to remove the Related Resources from the Blog Series twig for its footer, and the result will be that there will be no stand-alone "article-footer" on Blog Series.

- The Related Resources will be removed from the Blog Series twig. There will be no front-end nor CMS user experience changes

- Related Resources is a field that gets rendered on Blog Series, but it is not a field that can be added through the CMS.

### Acceptance Criteria
- There will be no changes to the front end of Blog Series
- There will be no changes to the CMS UI/UX of Blog Series
- The Related Resources field will be removed from the Blog Series Twig

dev-rana-publicis:
Closes #5160


ChadChandlerPS (approved):
Looks good!

PR #5202, opened 2026-04-15
(#3462) Fix Spanish event date localization

### LINKED ISSUE #3462: Date displays in English on Spanish events
### Issue description
Wrong language for the date when an event is translated to Spanish

> **ESTIMATE** TBD

1. Create a new event in English.
2. Save and publish the event.
3. Create a Spanish translation of the event
4. Publish the translated event

- The Spanish version of the event displays the date displays in Spanish.

### What's the actual result?
- The Spanish version of the event displays the date displays in English.

- image

ChadChandlerPS:
Twig's `|date()` filter always outputs month names in English regardless of the node's language. Pre-format event dates in preprocess using `DateFormatter::format() `with the node's langcode so month names are correctly localized.

Because the fix here is in preprocess_node which runs for all view modes of event content type, once the localized variables became available it wouldn't have been right to not fix the other templates that were using date() so this PR fixes the date issue everywhere event dates are rendered:

- **Full node view:** .node—cgov-event—full.html.twig
- **Event list view mode:** node—cgov-event—event-list-item.html.twig
- **NCIDS condensed collection:** node—cgov-event—event-ncids-collection-condensed-html.twig via the macros

**SCREENSHOT OF SPANISH EVENT DATE**
image

Closes #3462

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_event/cgov_event.install
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_event/cgov_event.install
@@ -141,3 +141,25 @@ function cgov_event_update_10001() {
   }
 
 }
+
+/**
+ * Add cgov_event date formats for localized date display.
+ */
+function cgov_event_update_10002() {
+  $formats = [
+    'cgov_event_date_long' => ['label' => 'Cgov Event Date Long', 'pattern' => 'F j, Y'],
+    'cgov_event_date_short' => ['label' => 'Cgov Event Date Short', 'pattern' => 'F j'],
+    'cgov_event_time' => ['label' => 'Cgov Event Time', 'pattern' => 'g:i A'],
+  ];
+
+  $storage = \Drupal::entityTypeManager()->getStorage('date_format');
+  foreach ($formats as $id => $data) {
+    if (!$storage->load($id)) {
+      $storage->create([
+        'id' => $id,
+        'label' => $data['label'],
+        'pattern' => $data['pattern'],
+      ])->save();
+    }
+  }
+}
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_event/config/install/core.date_format.cgov_event_date_long.yml
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_event/config/install/core.date_format.cgov_event_date_long.yml
@@ -1,0 +1,7 @@
+langcode: en
+status: true
+dependencies: {  }
+id: cgov_event_date_long
+label: 'Cgov Event Date Long'
+locked: false
+pattern: 'F j, Y'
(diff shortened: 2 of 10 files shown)


erikscottmeyer (approved):

bryanpizzillo (commented):
@ChadChandlerPS - I don't know if you need all that preprocessor work. Did you try `format_date` which calls the Drupal [date formatter](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Datetime%21DateFormatter.php/function/DateFormatter%3A%3Aformat/11.x)?

You can then translate the format date strings like https://github.com/NCIOCPL/cgov-digital-platform/blob/4fb375b873373068b1adb15b355da32cc920941b/docroot/profiles/custom/cgov_site/translations/cgov_site.es.po#L13-L15
.
At the very least you should not have to load the node and then get its translation. Actually, in thinking a little more, this is probably wrong too... If an English node is showing up in a spanish listing somehow, then it should still show the Spanish format. I note this because in the event listing, IIRC, we use a view mode to render the event, which would be a view mode preprocessed by your function.

ChadChandlerPS:
> @ChadChandlerPS - I don't know if you need all that preprocessor work. Did you try `format_date` which calls the Drupal [date formatter](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Datetime%21DateFormatter.php/function/DateFormatter%3A%3Aformat/11.x)?
> 
> You can then translate the format date strings like
> 
> https://github.com/NCIOCPL/cgov-digital-platform/blob/4fb375b873373068b1adb15b355da32cc920941b/docroot/profiles/custom/cgov_site/translations/cgov_site.es.po#L13-L15
> 
> 
> .
> At the very least you should not have to load the node and then get its translation. Actually, in thinking a little more, this is probably wrong too... If an English node is showing up in a spanish listing somehow, then it should still show the Spanish format. I note this because in the event listing, IIRC, we use a view mode to render the event, which would be a view mode preprocessed by your function.

@bryanpizzillo Ahh these `PO` files are handy.

Also, the .po file only has a translation for F j, Y. The templates also use F j (start date on multi-day same-year events) and g:i A (times). Those would need additional .po entries to be fully localized too.

Happy to give it a shot if you'd like.

andyvanavery31 (approved):
This passes product review. 

Other currently known bugs that can be observed in this testing, but will be fixed in other work:
- Creating a Spanish Event (not translating an English one) throws an error for an unknown node, but successfully creates the Spanish Event
- All Spanish Events (both when created directly in Spanish and in being translated from English) do not display the time of the event.

ChadChandlerPS:
> This passes product review.
> 
> Other currently known bugs that can be observed in this testing, but will be fixed in other work:
> 
> * Creating a Spanish Event (not translating an English one) throws an error for an unknown node, but successfully creates the Spanish Event
> * All Spanish Events (both when created directly in Spanish and in being translated from English) do not display the time of the event.

I'm going to go ahead and fix this right now - should be working by morning.

andyvanavery31:
Update to add times to Spanish events works correctly, and remains passing for Product.

erikscottmeyer (approved):

andyvanavery31 (approved):

alinai (approved):
Passed Regression Tests

PR #5219, opened 2026-04-27
(#4880) Update order rules for events

### LINKED ISSUE #4880: CR: Add Order Rules for Events Dynamic List for Events at the Same Date & Time
### Request Description
When Events have the same date and time and are displayed on a Dynamic List, then the order that they appear is governed by the order of which they were uploaded. This request is to add rules to the Dynamic List to handle when Events have the same date and time.

- The new ordering rules for displaying events on a Dynamic List are:

1. Date
2. Time
3. Alphabetical
4. Oldest Upload Time

- The ordering rules for displaying events on a Dynamic List are:

1. Date
2. Time
3. Oldest Upload Time

### Acceptance Criteria

1. When a list of Events are displayed on an NCIDS Dynamic List and two or more events have the same date and time, then they will be ordered alphabetically.
2. When a list of Events are displayed on an NCIDS Dynamic List and two or more events have the same date, time, and event name, then they will be ordered in the order that they were uploaded to the system.

### Related Tickets
* YAML Update Ticket #4879

dlescarbeau:
- adds rule to sort alphabetically if dates and times are the same
- adds rule to sort by creation time if dates, times, and titles are all the same
- updates granularity of date / time sort to be to the minute rather than the second for admin user clarity

Closes #4880

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_event/config/install/views.view.cgov_events.yml
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_event/config/install/views.view.cgov_events.yml
@@ -627,7 +627,53 @@ display:
             label: ''
             field_identifier: field_event_start_date_value
           exposed: false
+          granularity: minute
+        title:
+          id: title
+          table: node_field_data
+          field: title
+          relationship: none
+          group_type: group
+          admin_label: ''
+          entity_type: node
+          entity_field: title
+          plugin_id: standard
+          order: ASC
+          expose:
+            label: ''
+            field_identifier: ''
+          exposed: false
+        created:
+          id: created
+          table: node_field_data
+          field: created
+          relationship: none
+          group_type: group
+          admin_label: ''
+          entity_type: node
+          entity_field: created
+          plugin_id: date
+          order: ASC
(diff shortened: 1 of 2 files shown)


lburack (approved):
Had two events with the same date and time and the event that was created first displayed first. This didn't follow alphabetical rules properly.
- Test Future Event 2 was created before Lindsay's Event Test
- but Lindsay's Event Test would come first alphabetically

https://ncigovcdode1111.prod.acquia-sites.com/about-nci/organization/cgh/events

CC @dlescarbeau  @andyvanavery31

erikscottmeyer (commented):
I don't think its a bad idea to address the copilot suggestion to have final tie-breakers on sort after created by nid.  Why don't we do that?

erikscottmeyer (approved):
I think the grnaularity issue raised by Otto isn't real here (start and end dates only have a granularity of minute anyway, and we want the granularity to be second for the other sort criteria.)

KateMashkinaNIH (approved):

lburack (approved):
This now passes my review. Rules are being followed as expected when events have the same date and time (not accounting down to the second anymore either!)

PR #5215, opened 2026-04-22
(#5075) Upgrade to Drupal 10.6.7

ChadChandlerPS:
- Upgrades Drupal core from 10.5.6 to 10.6.7
- Upgrades ImageMagick 3.4.0 to 4.0.2
- Upgrades File Metadata Manager 2.6.0 to 3.1.0
- Updates CKEditor plugin dev dependencies to match core (47.6.0)
- Adds `cgov_core_update_10017()` enable mysql database driver module
- Updates Spanish translation files (`.po`) for all upgraded packages

Closes #5075

PR #5221, opened 2026-04-28
(#5208) Remove drupal/acsf module and references

### LINKED ISSUE #5208: Remove drupal/acsf module
The ACSF connector module (drupal/acsf) was required by ACSF. As we are in MEO this module is doing nothing for us anymore. The only reason it was not deployed during the MEO deployment was that we needed it to still be there so we could uninstall it from the sites.

Task:
- [ ] Remove drupal/acsf composer package
- [ ] Remove references to acsf in `blt/blt.yml`

ChadChandlerPS:
Removes `drupal/acsf` from `composer.json` and `composer.lock`. Removes **acsf** from module toggle uninstall lists in `blt.blt.yml `for follow environments.

- local
- meodev 
- meostage 
- meoprod 

**MERGE AFTER:** This PR should be merged after the [Drupal 10.6.7 upgrade
PR](https://github.com/NCIOCPL/cgov-digital-platform/pull/5215) due to security advisory blocks on the current develop branch.
image

Closes #5208


ChadChandlerPS:
Removing `drupal/acsf` from `composer.json` and `composer.lock `causes `ddev blt setup `to fail with a PreExistingConfigException on cgov_saml_auth_config - specifically `autologout.settings` and `samlauth.authentication` already exist in active configuration. 
image

A clean branch off develop installs fine. Is there something about `drupal/acsf` that was involved in the install process that maybe prevented this config conflict?

**EDIT to include**
Getting a consistent CI failure with Class `'Drupal\cgov_core\CgovPurgeConfigInstaller' `not found in `cgov_dummy_purger_install()`

ChadChandlerPS:
I asked codeex the exact question I put above:

> Removing `drupal/acsf` exposes a latent config ownership bug in `cgov_saml_auth_config` — it ships `config/install/autologout.settings.yml` and `config/install/samlauth.authentication.yml` but those configs are owned by **autologout** and **samlauth**. 
> 
> Drupal installs those modules first, creates their configs, then `cgov_saml_auth_config` tries to install its own copies and hits `PreExistingConfigException`. 
> 
> `ACSF `was masking this via `AcsfConfigStorage` which overrides Drupal's config storage layer and silently handled the conflict. Without `drupal/acsf` in vendor, raw Drupal exposes it. The fix is moving those config values into an install hook in `cgov_saml_auth_config`similar to what `_update_10002()` already does

bryanpizzillo:
> AcsfConfigStorage

There is no such thing as AcsfConfigStorage. Maybe ask codex again and to provide references?

ChadChandlerPS:
> > AcsfConfigStorage
> 
> There is no such thing as AcsfConfigStorage. Maybe ask codex again and to provide references?

@bryanpizzillo 
**Output (edits from me at the end):** I owe you a correction. The ACSF module code was never committed to this repo (it was gitignored in vendor and has since been removed). I was speculating about what ACSF might contain and presented it as fact. That was wrong of me.

**What I'm confident in:**

- The PreExistingConfigException root cause is solid — the config ownership conflict in cgov_saml_auth_config is real and the fix is correct
- Removing drupal/acsf from composer is what surfaced it

**What I don't actually know:**

- The exact mechanism by which ACSF's presence was masking the conflict, but after reading the link you gave it looks like something in its dependency tree was causing Features to start earlier in install queue

PR #5254, opened 2026-05-29
(#3220, #5253) Fix issue with dates in ical

### LINKED ISSUE #3220: Issue with dates appearing in the outlook calendar
### Issue description
When we change the date of an event, publish the event and then download the event from the “Add to Outlook Calendar”, the old date appears in the downloaded Outlook Calendar, instead of the changed/updated date.  

> **ESTIMATE** TBD

1.  Go to an existing event in the cms of Cgov.
2. Change the date. 
2. Save and Publish.
3. Verify the changed date at the front end.
4. Download the Calendar by clicking the "Add to Outlook Calendar"
5. Open the Calendar in Outlook and see that the old date is appearing on the Outlook Calendar.

- The changed date should be reflected on the Outlook Calendar and both the front end of the content and the Outlook Calendar should be in sync.

### What's the actual result?
-  The old date appears in the downloaded Outlook Calendar, instead of the changed/updated date.  

image

### Related Tickets
* Issue #3903 
* Issue #4392 
* Issue #3363 
* Issue #3904

### LINKED ISSUE #5253: Refactor iCal Controller
There are multiple tickets related to the iCal controller having to do with unnecessary file generation, useless caching, and some other items. this needs to be cleaned up. 

This covers the following issues, which are all around problems with the controller and our approach.
* Issue #3903 
* Issue #3363 
* Issue #3904 

@andyvanavery31  can add any additional information, I am just creating the ticket so it can be closed.

erikscottmeyer:
- Also refactored the ics controller to be cleaner.

Closes #3220
Closes #5253

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_event/src/Controller/ICalendarController.php
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_event/src/Controller/ICalendarController.php
@@ -2,18 +2,15 @@
 
 namespace Drupal\cgov_event\Controller;
 
+use Drupal\Core\Cache\CacheableResponse;
 use Drupal\Core\Controller\ControllerBase;
+use Drupal\Core\Datetime\DrupalDateTime;
 use Drupal\Core\Entity\EntityStorageInterface;
-use Drupal\Core\File\FileExists;
-use Drupal\Core\File\FileSystem;
 use Drupal\Core\Session\AccountProxy;
-use Drupal\file\FileRepositoryInterface;
 use Eluceo\iCal\Component\Calendar;
 use Eluceo\iCal\Component\Event;
 use Symfony\Component\DependencyInjection\ContainerInterface;
-use Symfony\Component\HttpFoundation\BinaryFileResponse;
 use Symfony\Component\HttpFoundation\RequestStack;
-use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 
 /**
  * The iCalendar controller.
@@ -32,17 +29,6 @@ class ICalendarController extends ControllerBase {
    * {@inheritdoc}
    */
   public $request;
-  /**
-   * {@inheritdoc}
-   */
-  protected $file;
-
-  /**
-   * The file repository service.
-   *
-   * @var \Drupal\file\FileRepositoryInterface
-   */
-  protected $fileRepository;
 
   /**
    * Constructs an ICalendar Controller object.
@@ -53,17 +39,11 @@ class ICalendarController extends ControllerBase {
    *   The current user.
    * @param \Symfony\Component\HttpFoundation\RequestStack $request
    *   The request stack.
-   * @param \Drupal\Core\File\FileSystem $fileStorage
-   *   The file system.
-   * @param \Drupal\file\FileRepositoryInterface $file_repository
-   *   The file repository service.
    */
-  public function __construct(EntityStorageInterface $entityStorage, AccountProxy $currentUser, RequestStack $request, FileSystem $fileStorage, FileRepositoryInterface $file_repository) {
+  public function __construct(EntityStorageInterface $entityStorage, AccountProxy $currentUser, RequestStack $request) {
     $this->entity = $entityStorage;
     $this->currentUser = $currentUser;
     $this->request = $request;
-    $this->file = $fileStorage;
-    $this->fileRepository = $file_repository;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


erikscottmeyer:
I've confirmed that everything works as expected on the ODE: https://ncigovcdode1131.prod.acquia-sites.com/about-cancer/coping/feelings/relaxation/2019-investigators-site I added this to calendar (worked) then changed date and time and published and added again (worked with updated date).  Also tested /node/[NID]/icalendar for non-Event node (blog) https://ncigovcdode1131.prod.acquia-sites.com/node/1351/icalendar and for https://ncigovcdode1131.prod.acquia-sites.com/node/1351999999/icalendar; both gave, as expected iCalendar Error
iCalendar Error, please contact the System Administrator . Also logged two different errors, the first (non-event node): https://ncigovcdode1131.prod.acquia-sites.com/admin/reports/dblog/event/3071 iCalendar download failed: requested Node ID 1351 is not a cgov_event bundle. the second, node doesn't exist: https://ncigovcdode1131.prod.acquia-sites.com/admin/reports/dblog/event/3076 iCalendar download failed: requested Node ID 1351999999 does not exist.

alinai:
@erikscottmeyer , If we have "00" in the seconds field (e.g. 3:25:00) on the "Event Start Date & Time" field, both Event page and Outlook Calendar shows same time but if we have any number in the seconds field (e.g. 3:25:32) on the "Event Start Date & Time" field, then Outlook time is increased by 1 minute.

image

image


cc: @andyvanavery31 , @KateMashkinaNIH

andyvanavery31:
> @erikscottmeyer , If we have "00" in the seconds field (e.g. 3:25:00) on the "Event Start Date & Time" field, both Event page and Outlook Calendar shows same time but if we have any number in the seconds field (e.g. 3:25:32) on the "Event Start Date & Time" field, then Outlook time is increased by 1 minute.
> 
> image image
> cc: @andyvanavery31 , @KateMashkinaNIH

@alinai I can't reproduce this issue even if I set both the start and end times to have ":59" at the end. I tested both downloading to Apple Calendar and to Outlook and both worked correctly. Could we get someone to confirm this? We might need to see if it's a Mac vs PC issue

ChadChandlerPS (approved):

KateMashkinaNIH (approved):

andyvanavery31 (approved):

docroot/profiles/custom/cgov_site/modules/custom/cgov_event/src/Controller/ICalendarController.php
```diff
@@ -123,54 +107,43 @@ public function download($nid) {
...
+      'Content-Length' => strlen($content),
+      'Cache-Control' => 'no-cache, private',
     ];
 
-    // \Drupal\Core\EventSubscriber\FinishResponseSubscriber::onRespond()
-    // sets response as not cacheable if the Cache-Control header is not
-    // already modified. We pass in FALSE for non-private schemes for the
-    // $public parameter to make sure we don't change the headers.
-    return new BinaryFileResponse($uri, 200, $headers, $scheme !== 'private');
+    return new Response($content, 200, $headers);
```
bryanpizzillo: This should return a Drupal\Core\Cache\CacheableResponse and on the response you should add:
`$response->addCacheableDependency($node);`

This will bubble up all the cache tags for the node. So if the event is updated the cache clears, but otherwise Varnish and Akamai will keep it in the cache until the next cache clear. 

You would then remove the Cache-Control header since the CacheableResponse would return it AND, well, we want it cacheable. 

There *is* technically a world in which Akamai might be best to have .ics files set to never cache. But that tells clients' browsers not to cache, but Akamai and Varnish will.

docroot/profiles/custom/cgov_site/modules/custom/cgov_event/src/Controller/ICalendarController.php
```diff
@@ -123,54 +107,43 @@ public function download($nid) {
...
-      'Content-Type' => $mimetype,
+      'Content-Type' => 'text/calendar; charset=utf-8',
       'Content-Disposition' => 'attachment; filename="' . $filename . '"',
-      'Content-Length' => $file->getSize(),
-      'Content-Transfer-Encoding' => 'binary',
-      'Pragma' => 'no-cache',
-      'Cache-Control' => 'must-revalidate, post-check=0, pre-check=0',
-      'Expires' => '0',
-      'Accept-Ranges' => 'bytes',
+      'Content-Length' => strlen($content),
```
bryanpizzillo: There are some notes that I found that Content-Length should not be set when dealing with non-streamed responses (e.g., `Transfer-Encoding: chunked` HTTP header) because as of Drupal 10.2 Drupal does it for you. See https://www.drupal.org/node/3298551.

bryanpizzillo (changes requested):
This is preventing caching and should be fixed.

docroot/profiles/custom/cgov_site/modules/custom/cgov_event/src/Controller/ICalendarController.php
```diff
@@ -74,44 +54,48 @@ public static function create(ContainerInterface $container) {
...
+        $message = 'iCalendar download failed: requested Node ID @nid does not exist.';
+      }
+      elseif ($node->bundle() !== 'cgov_event') {
+        $message = 'iCalendar download failed: requested Node ID @nid is not a cgov_event bundle.';
+      }
+
+      $this->getLogger('cgov_event')->warning($message, ['@nid' => $nid]);
+
+      $error_html = '<h1>iCalendar Error</h1><p>iCalendar Error, please contact the System Administrator</p>';
+      return new CacheableResponse($error_html, 400, ['Content-Type' => 'text/html; charset=utf-8']);
```
bryanpizzillo, line 81: This can stay a response because a 400 is an error and we don't want things caching this url forever with an error.

andyvanavery31:
Approved by Product following the latest push.

bryanpizzillo (approved):

PR #5267, opened 2026-06-09
(#5209) Fix admin UI language negotiation to always display in English

### LINKED ISSUE #5209: Fix Admin UI Language Negotiation to Always Display in English
### Issue description
The site's language negotiation is currently configured to use user preferences for interface translation. When editors switch to Spanish for content translation, the entire Drupal admin UI ( including the admin menu ) also switches to Spanish. 

The desired behavior is for the admin interface to remain in English at all times, regardless of the content language being edited.

> **ESTIMATE** TBD

### When viewing Spanish content, the admin menu/interface should not be translated. It should be set to English, the default language.
Screenshot of desired behavior

Image

-

The current functionality is that the entire admin menu is translated.
-

- The Drupal admin menu and interface always display in English regardless of the editor's selected content language
- Editors can switch content to Spanish for translation purposes without the admin UI language changing
- User language preferences do not affect the admin interface language
- Spanish content translation functionality continues to work as expected
- Verified in both English and Spanish content editing workflows

ChadChandlerPS:
Adds a custom `LanguageNegotiationAdminEnglish` plugin that returns English for authenticated users, keeping admin menu in English regardless of content language. Anonymous visitors on `/espanol/ `pages still receive URL-based interface detection so `.po` translations (e.g. por/by) load correctly. Removes language-user-admin from the enabled interface negotiation methods.

Closes #5209

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.module
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_core/cgov_core.module
@@ -18,6 +18,7 @@ use Drupal\Core\Entity\EntityTypeInterface;
 use Drupal\Core\Field\FieldDefinitionInterface;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Language\Language;
+use Drupal\Core\Language\LanguageInterface;
 use Drupal\Core\Render\Element;
 use Drupal\Core\Session\AccountInterface;
 use Drupal\Core\Site\Settings;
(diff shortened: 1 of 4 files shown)


ChadChandlerPS:
Note: This was more overkill than I originally had hoped. I was able to get it working via core config changes but strings/dates do not get translated to Spanish via core config changes. We wanted to time box this effort so this PR is really just to see what happens when Kate runs the tests against it. Not suggesting we go with this exact approach necessarily. If I spent more than half a day on it perhaps I could find a way to do this without a plugin.

tldr; let's just see what QA tests come back with in re: to fail/pass

KateMashkinaNIH:
regression tests run https://github.com/NCIOCPL/cgov-digital-platform-acceptance-tests/actions/runs/27973160368

ChadChandlerPS:
If this PR were to merged as it stands - strings (i.e., Spanish dates, words)  that originate from .po files will not load for logged in users, but WILL continue to operate as normal for external not logged-in users.

Text that is added manually  on the actual Spanish node form WILL show to all users regardless of their logged in status. Example below:
image

**TLDR;** The only Spanish text logged in users see is Spanish text that is manually added on the Spanish version of the node as seen in picture above.  The benefit here is when switching to Spanish the entire admin menu doesn't also become unnavigable. This has _zero impact_ on the external site visitor.

ChadChandlerPS:
This can be closed as now that QA's tests now work for English -- we will just set preferred language to English.

PR #5292, opened 2026-06-30
(#3184) Bog Post cannot add Blog topic before Related Resources inter…

### LINKED ISSUE #3184: Blog Post: Cannot add category before internal links
### Issue description
On Production, if I select a blog category, I run into an issue with adding an internal link, where an "Illegal choice has been made"  I should be able to add a blog category and internal links, and they should not be inter-related.

image

NOTE: there is a workaround, users can add internal links FIRST before selecting one or more categories.

> **ESTIMATE** TBD

1.  Create or edit a blog post
2. Select a blog series, and a blog category
3. Once you select a blog category, try to add an internal link to related resources.  You will get the error.
4. Similarly, try to add an internal featured post. You will also get the error.

We should be able to add internal links and cards, regardless if there is a blog category selected.

### What's the actual result?
Right now, once the blog category is selected, we cannot add internal links.  

Looks like this might have been introduced in #2382 

### Related Tickets
* Issue #9999
* Issue #9999

erikscottmeyer:
…nal link

Closes #3184

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_blog/cgov_blog.module
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_blog/cgov_blog.module
@@ -48,17 +48,44 @@ function cgov_blog_form_alter(&$form, FormStateInterface $form_state, $form_id) 
         'event' => 'change',
         'method' => 'replace',
       ];
+      // Get the current entity from the form to
+      // figure out the active form language code.
+      $form_object = $form_state->getFormObject();
+      $entity = method_exists($form_object, 'getEntity') ? $form_object->getEntity() : NULL;
+      $langcode = $entity ? $entity->language()->getId() : \Drupal::languageManager()->getCurrentLanguage()->getId();
 
-      $series_id = $form_state->getValue("field_blog_series");
+      $user_input = $form_state->getUserInput();
       $blog_manager = \Drupal::service('cgov_blog.blog_manager');
-      // When creating a translation form state does not have the series
-      // information yet.
-      if ($series_id == NULL) {
-        $series_id = $form['field_blog_series']['widget']['#default_value'];
-        $series_nid = $series_id[0];
+
+      // 1. Check raw active user input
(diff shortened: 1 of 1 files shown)


ChadChandlerPS (approved):
Looks good!

alinai:
@erikscottmeyer , Spanish Translation is not showing the list of Blog Topics with check boxes on the CMS. We can see only the heading "Blog Topics".
image

image

cc: @KateMashkinaNIH , @andyvanavery31

ChadChandlerPS (approved):
One thing I noticed before your latest commit in the previous version, branch 2 was using `$series_id['target_id']` instead of `$series_id[0]['target_id']`. 
 
I had a correction ready to post but then looked at `getTopicsBySeriesId` in `BlogManager.php` and saw that branch was essentially unreachable in practice, so I let it go. Glad to see the latest commit cleaned that up anyway!

alinai:
Noticed that on ODE,  if a Spanish Blog Series doesn't have any Blog Topics that are translated into Spanish, then the English versions of those Topics (checkboxes) does not appear on the Spanish Blog Post but the "Blog Topics" heading appears. Confirmed with @andyvanavery31 that this is the desired behavior and he will reach out to IA's to see what they think is best for the "Blog Topics" heading appearing without checkboxes in the Spanish version
cc: @andyvanavery31 , @KateMashkinaNIH 

image

lburack (approved):
This is working as expected now in the CMS on the ODE. Passing for IA

KateMashkinaNIH (approved):

andyvanavery31:
This PR also fixes a bug where, if there is an English Series with English Blog Topics and a linked English Blog Post, and then the Series is translated, but the Blog Topics are not, then the translation of the Blog Post to Spanish would allow the selection of English Blog Topics. 

Now, when the above scenario happens, and there are no Spanish Blog Topics for the translated Blog Series, then the CMS will display the field title "Blog Topics" but will have no checkbox options below it. This is an accepted behavior from Product.

PR #5319, opened 2026-07-23
(#5318) Fix enlarge button text on Spanish pages

### LINKED ISSUE #5318: Fix enlarge button text on Spanish pages
After the Metatag 2.x upgrade, the content-language meta tag is no longer output. `enlarge.js` relied on this tag to render "Ampliar" on Spanish pages. With the tag gone, Spanish pages always show "Enlarge" instead.

1. Navigate to any Spanish-language page that contains a data table 
2. Ensure viewport width is > 1024px
3. If the table is wider than its container, an "Enlarge" button will appear above it
4. Observe: button reads "Enlarge" instead of "Ampliar"'

Note: When the table is enlarged in either language, an icon of an "X" appears to close it.

### Root Cause

`enlarge.js` reads `meta[name="content-language"]` to determine page language for button text. This meta tag was removed as part of the Metatag 2.x upgrade. With the tag absent, the language check returns undefined, which never equals 'es', so Spanish pages always render English button text.

ChadChandlerPS:
After the Metatag 2.x upgrade, the content-language meta tag is no longer output. `enlarge.js` relied on this tag to render "Ampliar"/"Cerrar" on Spanish pages. With the tag gone, Spanish pages always show "Enlarge"/"Close" instead

`domManipulation.js` had the same content-language dependency with an unnecessary fallback - simplified to use `document.documentElement.lang` directly.

Closes #5318

### DIFF
--- a/docroot/profiles/custom/cgov_site/themes/custom/cgov/src/utilities/domManipulation.js
+++ b/docroot/profiles/custom/cgov_site/themes/custom/cgov/src/utilities/domManipulation.js
@@ -91,21 +91,11 @@ export const getMetaData = (metaTags, document) => {
 
 /**
  * TODO: Extend with extra checks, this is very specific to CGOV.
- * @param {HTMLElement} [document=window.document]
+ * @param {Document} [document=window.document]
  * @return {string}
  */
 export const getDocumentLanguage = (document = window.document) => {
-  // MIGRATION NOTE:
-  // The fallback values were added during migration because a content-language attribute was not available
-  // at the time this file was ported.
-  const language = document.querySelector('meta[name="content-language"]')
-    ? document
-        .querySelector('meta[name="content-language"]')
-        .getAttribute("content")
-    : document.documentElement.lang
-    ? document.documentElement.lang
-    : "en";
-  return language;
+  return document.documentElement.lang || "en";
 };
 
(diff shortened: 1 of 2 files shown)


erikscottmeyer (approved):

andyvanavery31 (approved):

alinai (approved):

docroot/profiles/custom/cgov_site/themes/custom/cgov/src/utilities/domManipulation.js
```diff
@@ -95,17 +95,7 @@ export const getMetaData = (metaTags, document) => {
  * @return {string}
  */
 export const getDocumentLanguage = (document = window.document) => {
```
erikscottmeyer, line 97: I agree; I would change line 94 to @param {Document}

ChadChandlerPS: Done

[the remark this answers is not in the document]
ChadChandlerPS: done

andyvanavery31 (approved):

PR #5360, opened 2026-08-26
(#3988) Remove ncids promo block dark theme hack

### LINKED ISSUE #3988: Remove nci-promo-block--dark condition once prod has been updated
### Issue description
**!!!UPDATE!!!** We actually had to keep the list item in the Field Storage for field_ncids_theme in order to do the update as Drupal does not let you remove options on a field that is in use. This needs to be removed as part of this and **the update must be tested.** I would like for this to be actually tested on ACSF since that environment will also have revision data populated.

[PR 3982](https://github.com/NCIOCPL/cgov-digital-platform/pull/3982) applies the dark theme to the promo block if the value from the database is `dark` or `nci-promo-block--dark`. After updating prod with [PR 3982](https://github.com/NCIOCPL/cgov-digital-platform/pull/3982), we will have to resave the landing pages that use `NCIDS Promo Block` and then we can remove our hack as our database will have the correct values.

### Related Tickets
* Issue #3981

olitharp-nci:
Closes #3988

### DIFF
--- a/docroot/profiles/custom/cgov_site/modules/custom/cgov_home_landing/config/install/field.storage.paragraph.field_ncids_theme.yml
+++ b/docroot/profiles/custom/cgov_site/modules/custom/cgov_home_landing/config/install/field.storage.paragraph.field_ncids_theme.yml
@@ -16,9 +16,6 @@ settings:
     -
       value: light
       label: Light
-    -
-      value: nci-promo-block--dark
-      label: "DON'T USE"
   allowed_values_function: ''
 module: options
 locked: false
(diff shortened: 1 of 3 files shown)


bryanpizzillo:
@olitharp-nci based on changes we did in https://github.com/NCIOCPL/cgov-digital-platform/pull/4891, there probably needs to be an update hook to remove the option. IIRC Features would balk on the import of the config because it thought it sketch.  (I could be wrong.)

Locally an install of master followed by checking out your branch and doing a `blt drupal:update` will tell you if it will without the update hook. (Technically the deployment check workflow should test this, but it can miss things)

This is something that should be tested on ACSF though to make sure it installs.

